PR #5, opened 2016-10-29
Fix asset path.

Jeroen-G:
Perhaps you can test if it works on your install too.
I had the issue that all of Voyager's assets were not loaded, because the base url was not localhost but localhost/myapp/public.
Using the asset() helper, the files are loaded properly from the assets folder.

### DIFF
--- a/config/voyager.php
+++ b/config/voyager.php
@@ -27,7 +27,7 @@ return [
     |
     */
 
-    'assets_path' => '/vendor/tcg/voyager/assets',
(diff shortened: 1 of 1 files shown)


Jeroen-G:
Okay, the same thing might be the case for the storage path.

tnylea:
Thanks man! I'm pushing a new commit to version 0.9.14 that will fix the storage issue that you may have been having. If it was in the TinyMCE editor where it would not link correctly, that has been resolved.

Thanks again!

PR #13, opened 2016-10-31
Update browse.blade.php

putera:
Show the display name instead of field name on browse.


tnylea:
Thanks for catching that man 👍

PR #16, opened 2016-10-31
Seeder Data Row Patch

brandonferens:
This updates the seeder for the data rows with a proper select tag for page status

### DIFF
--- a/src/Models/Page.php
+++ b/src/Models/Page.php
@@ -3,6 +3,7 @@
 namespace TCG\Voyager\Models;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Auth;
 
 class Page extends Model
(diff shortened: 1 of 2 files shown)


src/Models/Page.php
```diff
@@ -3,6 +3,7 @@
 namespace TCG\Voyager\Models;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Auth;
```
fletch3555, line 6: This addition doesn't provide any benefit to the Pull Request.  Is it needed?

brandonferens: I believe we can remove this PR. I had created this on Master, issued the PR. Then I created a separate branch for a different PR and it had this change on it. So it's already up.

Issue #18, opened 2016-10-31
Artisan stops working after install


afsardo:
After installing voyager.

`php artisan voyager:install`

If you try to run any command (`php artisan make:model TwitchStream`) it throws an error:

```
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 2 passed to Illuminate\Routing\UrlGenerator::__construct() must be an instance of Illuminate\Http\Request, null given, called in /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/RoutingServiceProvider.php on line 60 in /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/UrlGenerator.php:103
Stack trace:
#0 /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/RoutingServiceProvider.php(60): Illuminate\Routing\UrlGenerator->__construct(Object(Illuminate\Routing\RouteCollection), NULL)
#1 /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Container/Container.php(290): Illuminate\Routing\RoutingServiceProvider->Illuminate\Routing\{closure}(Object(Illuminate\Foundation\Application))
#2 /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Container/Container.php(746): Illuminate\Container\Container->Illuminate\Container\{closure}(Ob in /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/UrlGenerator.php on line 103

Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 2 passed to Illuminate\Routing\UrlGenerator::__construct() must be an instance of Illuminate\Http\Request, null given, called in /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/RoutingServiceProvider.php on line 60 in /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/UrlGenerator.php:103
Stack trace:
#0 /home/vagrant/Code/Voyager/vendor/laravel/framework/src/Illuminate/Routing/RoutingServiceProvider.php(60): Illuminate\Routing\UrlGenerator->__construct(Object(Illuminate\Routing\RouteCollection), NULL)
[...]

PR #19, opened 2016-10-31
Fix Issue #18

afsardo:
Fix Issue #18 

Removed unnecessary asset() in the config file which conflicts with non HTTP app access.

Example: Running a `php artisan` command.


tnylea:
Thanks man.

PR #26, opened 2016-10-31
Only publish the assets needed by Voyager

adriaanzon:
Instead of publishing the assets for every package.


tnylea:
Thanks man!

ghost:
This means auth scaffold will not generate?

adriaanzon:
No, that is still being done [here](https://github.com/the-control-group/voyager/blob/672a1c2b0b02b62b002d7df486daeb62b32270f4/src/VoyagerCommand.php#L57).

This will only prevent publishing assets from other packages,  see https://laravel.com/docs/5.2/packages#publishing-file-groups

PR #29, opened 2016-10-31
Use ::class property in config file

adriaanzon:
Looks a bit cleaner than the escaped backslashes

I also think 'model' would be a more suitable name than 'namespace' but changing that would be a breaking change

### DIFF
--- a/config/voyager.php
+++ b/config/voyager.php
@@ -2,7 +2,7 @@
 
 return [
 
-    
+
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


tnylea:
Just pulled this in with the latest dev-master. Should be up in the latest release later today. Thanks for the contribution. I did a few file changes so there were a few conflicts with the merge, but moving forward I'll merge the requests in instead of manually adding them myself.

Appreciate the contribution :)

Issue #30, opened 2016-10-31
SQLite not supported


afsardo:
When using the Database feature, since it's hard coded the query to select all tables and other specific MySQL queries it crashes when using sqlite driver.

`SQLSTATE[HY000]: General error: 1 near "SHOW": syntax error (SQL: SHOW TABLES) (View: /Users/afsardo/Code/.../vendor/tcg/voyager/src/views/tools/database/index.blade.php)`

So since i am really in love with this package (seriously it's awesome <3) i will try and add compatibility to all major database drivers.

adriaanzon:
For PostgreSQL (#10) it would be something like 

``` sql
SELECT table_name FROM information_schema.tables WHERE table_type = 'BASE TABLE' AND table_schema = 'public' 
```

ptcarino:
that'll force me to edit vendor files. i'm in a situtation where i can't edit vendor files when i deploy. to be precise, i'm deploying my app using bluemix and everytime i use "cf push" to deploy my app, it does a composer install so editing vendor files is not an option.

i was thinking, maybe we could specify in the config/voyager.php what db we're using and that'll dynamically change the DB::select in the src/views/tools/database/index.blade.php?

afsardo:
That's the PR i sent, #31.

It detects the database driver you are using and resolves the "query" on it's own.

All the view needs is `DBSchema::tables()` (to show all tables).

Still in the process of implementing it for all the other needed methods.

marktopper:
Is this resolved now, since the merge of the PR #31?

PR #31, opened 2016-10-31
All DB Drivers Proposal: Add SQLite driver compatibility. (Not PR ready)

afsardo:
First pull request for Proposal #30 

Conditionally get all tables from DB according to connection driver.

### DIFF
--- a/src/Facades/DBSchema.php
+++ b/src/Facades/DBSchema.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace TCG\Voyager\Facades;
+
+use DB;
+use PDO;
+
+class DBSchema 
+{
+
+    const SQLITE = 'sqlite';
+
+    public static function tables() 
+    {
+        if (static::isSQLITE()) {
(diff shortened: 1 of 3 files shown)


jaysudodeveloper:
is there a status on this, i only have the option to use sqlite and would like to use voyager as other alternatives are not up to date.

afsardo:
Some functionality isn't supported for other DB's but all the pending commits were merged.

I'll be working in making it compatible with the other 2 db systems.

jaysudodeveloper:
@afsardo is it an easy change? reading through the issues related around this it seems that voyager was hard coded for MySQL? Are there tests i can run and try to make a work around myself until a compatible release comes.

afsardo:
@jaytheking It is not hard, but not a fast "change".

For the "Get all tables" i extracted it myself and added a static class that resolves the db driver you are using and fetch them accordingly.

https://github.com/the-control-group/voyager/blob/master/src/Facades/DBSchema.php

This is the Facade you should try and implement and change where the queries for db schema logic is hardcoded.

PR #40, opened 2016-11-01
Show successful flash instead of error when saving BREAD entries.

fletch3555:
When updating a BREAD entry, if no fields were changed, the response would be null, and would trigger an "error" flash message.  It's much better for UX if a successful message is shown instead.


tnylea:
Thanks man!

PR #43, opened 2016-11-01
Add input:file uploading script

webwizo:
I have just updated the code If BREAD input type is "file". We couldn't have uploaded file.


src/Controllers/VoyagerBreadController.php
```diff
@@ -246,6 +246,17 @@ public function getContentBasedOnType($request, $slug, $row){
...
     } else if($row->type == 'file'){
 
+        $file = $request->file($row->field);
+        $filename = str_random(20);
+
+        $path =  $slug . '/' . date('F') . date('Y') . '/';
+
+        $full_path = $path . $filename . '.' . $file->getClientOriginalExtension();
+
+        Storage::put(config('voyager.storage.subfolder') . $path . $filename . '.' . $file->getClientOriginalExtension(), (string)$file, 'public');
```
fletch3555: Why not just use `$full_path` here instead of duplicating `$path . $filename . '.' . $file->getClientOriginalExtension()`?

webwizo: I just place the upload code from "image uploading". I just follow the current code and variable styles.

When I select InputType=file, there is a error and I just follow and tried with this code and file uploaded, as per media library folder and file pattern.

webwizo: Well, I have got your point. I just changed the variable :) Thanks

tnylea:
Thanks for this commit. I actually added this in myself based on your code. Here on out any PR's will be run through the Travis tests and I can merge that way.

Thanks for the contribution. This has been added to version 0.9.21

Appreciate it!

Issue #46, opened 2016-11-01
problem when view a post


rockxsj:
I login as admin,when I want to view a post, eg:http://localhost:8000/post/latest-post. 
I will get the error message like this:
`Sorry, the page you are looking for could not be found.

1/1
NotFoundHttpException in RouteCollection.php line 161:`

afsardo:
Yes, i found that problem aswell it's a minor thing.

I will send a PR as a proposal, since i do not know what behaviour the maintainers want.

guahsu:
same problem here :D..

allenjd3:
Same here- looks like it can't find the route. I could make a route but it's not obvious in the docs where it should point.

Omazon:
same problem too

fletch3555:
#49 was merged in 25 days ago and should resolve this issue.

Closing this for now.  If this is still an issue in the current release, please let me know and I'll reopen it.

PR #49, opened 2016-11-01
Fix Issue #46

afsardo:
Proposal to fix Issue #46 

Point post to the bread read action.


tnylea:
Thanks, this has been added in 👍

Issue #57, opened 2016-11-02
Is there a way to install voyager without migrating tables?


Mugluck:
Yo, your admin looks awesome! One problem, I've already got a users table and what have you, so how do I integrate this with an existing structure. Starting from blank is not an option.

afsardo:
I think what you should do is `vendor:publish` and remove the voyager `create_users_table` migration.

After that you can `artisan migrate` and `artisan db:seed`, if i'm not mistaken you are good to go ;)

Mugluck:
Yeah, I that's a viable workaround. :)

marktopper:
I'm wondering, why do you need to do this?

fletch3555:
@marktopper. Not sure if you commented in the wrong issue or what, but I'll answer anyway.  For the reasons discussed in #58

marktopper:
@fletch3555, was just wondering if @Mugluck had other reasons which we could think about when finding the _"perfect"_ solution 😄

fletch3555:
Closing as #58 was merged in.

marktopper:
Released in `v0.9.32`.

PR #58, opened 2016-11-02
Creating ability to skip migration and/or seeding on voyager:install

fletch3555:
I've seen a few comments (e.g. #57) about not wanting to run the migration and/or DB seeding as part of the install process.  Here's my proposed stop-gap.


afsardo:
If you do not run migrations and/or db seed what do you need the install for?

Also voyager only works if the db is migrated + seeded.

Sorry if i'm missing something.

Mugluck:
We may already have a viable users table, and so migration and seeding is unnecessary. Install also builds out the views and pages for Voyager, not just the database side.

fletch3555:
It allows you to only run one or the other.  Perhaps you've already run voyager:install, then modified something and wanted to re-seed the voyager tables, you could run it without re-running the migrations.  Perhaps you're installing it on top of an existing application and don't want to run the migrations.

afsardo:
@Mugluck it's not unnecessary, it needs to migrate all the other tables. I've posted a "work around" (not tested) to that problem at issue #57.

afsardo:
@fletch3555 In all of those cases it's a matter of running those commands individually, basicly what `artisan voyager:install` does is pretty much:
- `artisan vendor:publish`
- `artisan migrate`
- `artisan db:seed`

fletch3555:
@afsardo, while I agree that these command options aren't a necessity, what you posted in #57 is essentially the same as what I posted, but in more steps.  You suggest essentially running the voyager:install steps manually, which is not ideal.  Why should this command exist at all if the only way for it to be flexible is to ignore it and run the steps manually?

afsardo:
@fletch3555 That command exists so you install all the things voyager needs to work, if you make those things "flexible" i think it's much more dangerous. Dunno that's my opinion i understand your point as well ;) So all it comes down is for the Maintainers to decide.

Mugluck:
Might be nice to have an option to add something like -nomigrate to the install commands, just to simplify things, or a few commands for specifying that there is an existing users table, which means the installer only adds the necessary collumns to an existing table. But that's more of an ease of life thing.

tnylea:
@fletch3555 @Mugluck I get what you guys are saying here. There are 2 tables that will need to be migrated though, which are the data_types and data_rows. I can try and add the functionality for this next week, so I'll keep this open until then.

Thanks for the suggestion. I agree it will be needed for users of existing projects.

marktopper:
I think the big problem for users here is not that they do not want to run the migrations, its the seeding they do not like.
I personally do not like seeding for existing projects, however I have no problem seeding data for new projects.
However in this case where the user **needs** to seed the `data_types` and `data_rows` into the database in order to work, I would however force it by using migrations for that instead.
I know that this is against the principles of migrations and seeds, but rules are for being broken, at least a few times.
I however would really appreciate peoples mind on the matter.

fletch3555:
How about we add a command flag such as "essential-only" that only runs migrations/seeds for the essential tables, and not users/posts/etc.?

marktopper:
Wonderful idea @fletch3555!

We can still do it using seeders, however just without publishing them and calling them this way.

```
php artisan db:seed --class="\My\Package\Seeder\MyCustomSeeder"
```

marktopper:
@fletch3555 How about calling the flag `without-dummy` since the posts, categories and users seeders are just for dummy data.

Also, are you gonna do commits to this pull request?

fletch3555:
@marktopper,@tnylea

Sure, I'll update this PR with this change instead of the other one.

`--without-dummy` seems oddly phrased to me.  Can we do `--no-dummy-data` or something instead? Or `--no-seed-sample`? I'm open to other suggestions as well if these spark any ideas.

marktopper:
They sounds good to me 🎉

fletch3555:
Perfect. I'll pick one and let you know when done.

fletch3555:
@marktopper / @tnylea, How's this for an attempt?  I can update the documentation to go with it if you guys approve of this.

You can choose 1 of these options:

1. migrate and seed ALL tables (normal)
2. migrate ALL and only seed essential (`data_types` and `data_rows`) (`--no-dummy-seed`)
3. ONLY migrate and seed essential (`--no-dummy`)

marktopper:
There is no migration of dummy data.
Also the only reason I can see in order that someone might not wish to migrate is because they use the database table already for something else (which just will move the issue to somewhere else) or if they do not wish to use certain parts of Voyager like the menu builder (then the `menus` database table could be left out).
However I personally think that everything should be migrated to ensure that the database matches the requirements, however I wish to separate items like the menu builder from this migration and into another installation. Just like what Laravel does for using the database for storing sessions, in order to do that you must run `php artisan session:table` which will create the table.
What do you think @fletch3555, what's the reason you would not migrate it all?

fletch3555:
I personally was not using the Posts/Pages/Categories features, so I was building it toward importing the minimum necessary.  This is a step toward #147.  Importing only the minimum will prove it can be done, and allow each "Rig" to get built on top of that.

marktopper:
@fletch3555, okay then lets focus on having the install script migrating everything, and seeding dummy data when the flag `--no-dummy` is left out.
Then we can focus on removing certain migrations from the installation and move them into separate commands.

fletch3555:
@marktopper, how's this? Also, I'm not sure why Travis is failing... It doesn't appear to be related to anything I changed.

marktopper:
I will look at whats up with Travis 💪

marktopper:
I merged to a new branch in order to fix the issue with Travis. Will keep you posted.

marktopper:
Merged with TravisCI support at #221

Issue #23, opened 2016-10-31
Publish views, controllers and models


willmkt:
Is there a way to publish the views?

It would be better to have full control over views, controllers and models instead of keeping it inside the package folder.

Thanks

Melcus:
Yes please

marktopper:
This have been fixed in the latest version.

Now you can run `php artisan vendor:publish`.

PR #59, opened 2016-11-02
Publishing views

fletch3555:
In response to Issue #23.  Controllers and Models can be subclassed, so they don't need to be published.

### DIFF
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -24,7 +24,7 @@ class VoyagerServiceProvider extends ServiceProvider
             });
         }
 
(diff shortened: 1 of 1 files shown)


tnylea:
This actually should be loading from this location `$this->loadViewsFrom(__DIR__.'/views', 'voyager');` right?

Unless I'm missing something, but this is loading views from the /vendor/tcg/voyager/src/views folder so that way they can be added to the view namespace.

Let me know if you think otherwise 👍  Thanks again for the contribution.

fletch3555:
@tnylea This PR was to publish the views to the resources/views/vendor/voyager path so they can be modified if needed.  The loadViewsFrom() call needs to be updated to point at the new path.

Apologies if I missed your point/question entirely.

PR #60, opened 2016-11-02
Removal of eval

maiorano84:
Eval is widely considered to be bad practice for a multitude of reasons, many of which can be found on various forums and other relevant areas.

It is highly recommended that this code be changed to remove any and all instances of eval where its use is avoidable.


maiorano84:
Closed due to inactivity by the author

PR #67, opened 2016-11-02
Update VoyagerServiceProvider.php

MamunHoque:
We need to customize view files.  So I think its must need to publish view files also so we can modify or overrides any BRED files.


afsardo:
This Pull request is a duplicate of PR #59, nice catch anyway ;)

MamunHoque:
:)

tnylea:
Thanks man. I've got this merged in via a different request that publishes the views from the admin/posts folders.

Appreciate it 👍

Issue #9, opened 2016-10-30
Adding new table


tnylea:
Adding new table in voyager needs to make sure that the Primary ID has an auto_increment value otherwise you may receive an error when adding a new data type and not specifying an ID

edmeehan-tcg:
+1 was getting an error when trying to add new entries into a table.

filipsobol:
+1, I'm not able to find any option to add "auto_increment" value from admin panel.

Edit:
I noticed that ID in new tables has auto_increment value, but it's removed when table is updated.

Laran01:
I noticed that ID in new tables has auto_increment value, but it's removed when table is updated.

marktopper:
Seems to be fixed

Ping me if I'm wrong.

PR #76, opened 2016-11-03
Fix issue for update column to user table or any existing table

MamunHoque:
I think we don’t need to edit 'ID' column because if its have any relationship with other table. so if we update id table its not work.
screenshot from 2016-11-03 16 57 20

So I skip edit command where column is  'ID' .

You can also check to add or update your user table its show an error like below: 
QueryException in Connection.php line 769: SQLSTATE[HY000]: General error: 1833 Cannot change column 'id': used in a foreign key constraint 'user_roles_user_id_foreign' of table 'test-app.user_roles' (SQL: ALTER TABLE users MODIFY COLUMN id int(10) unsigned NOT NULL FIRST)

### DIFF
--- a/src/Controllers/VoyagerDatabaseController.php
+++ b/src/Controllers/VoyagerDatabaseController.php
@@ -142,9 +142,13 @@ class VoyagerDatabaseController extends Controller
 				//Reorder the rows
 				$field_name = (string)$request->field[$index];
 				if($index > 0){
-					\DB::statement("ALTER TABLE " . (string)$table_name . " MODIFY COLUMN " . $field_name . " " . $table_type_array[$field_name] . $table_default_array[$field_name] . $table_null_array[$field_name] . " AFTER " . $request->field[$index-1]);
+					if($field_name!='id') {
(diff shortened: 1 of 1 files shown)


filipsobol:
This may also fix issue #9, because right now 'id' field loses auto_increment attribute on table update.

I think 'ID' field shouldn't be displayed in panel at all. This can be done by adding condition to foreach loop in src > views > tools > database > edit-add.blade.php

``` php
    @foreach($rows as $row)
        @if($row->Field !== 'id')
            newRow('', '{{ $row->Field }}', '{{ $row->Type }}', '{{ $row->Null }}', '{{ $row->Key }}', '{{ $row->Default }}');
        @endif
    @endforeach
```

This alone fixed issue #9 for me.

tnylea:
Thanks man! This is great. Going to run a few tests on this 👍

sarath-c-sandwich:
This works really well, and it would be great if we can refactor to controller. Anyway, for now it's good enough. Thanks man.

PR #86, opened 2016-11-04
Add ability to select user roles on users data type

handiwijoyo:
Add ability to select user roles when create or edit user.

image

### DIFF
--- a/src/Controllers/VoyagerUserController.php
+++ b/src/Controllers/VoyagerUserController.php
@@ -1,0 +1,343 @@
+<?php
+
+namespace TCG\Voyager\Controllers;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use App\Http\Requests;
+use App\Http\Controllers\Controller;
+use TCG\Voyager\Models\User as User;
+use TCG\Voyager\Models\DataType as DataType;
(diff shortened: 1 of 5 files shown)


fletch3555:
This is a good start, but seems too specific.  I'd rather see a solution where you can add fields like this to handle relations elsewhere.  So if I manually edit a model (generated or otherwise), and add a many-to-many or one-to-many relation to it, I'd like that field to be editable like you have here.

fishmad:
Why did this get deleted its what we need?

tomicakorac:
@fishmad because the solution was too specific. It would be more practical to have a universal solution for any kind of model relations, not just the user-role relation.

By the way, I was wondering, has anyone tried using Voyager along with other packages for user role management,such as Zizaco's entrust, or some other? How would that work out, if at all?

joshuatf:
Added a many to many relation option that works for this on issue #188 

@tomicakorac Middleware seems to be hardcoded in, but I would also like to see the ability to use Entrust or other packages here.  Frozennode admin package has a permission option in their configuration that works perfectly for this and allows you to use any method or package you'd like.  

That works for general access, but menu item visibility and permission for specific CRUD actions on models also needs to be addressed.  Willing to take the ball on this, but I don't think the JSON options is a good place to stick this.

PR #92, opened 2016-11-04
Register admin account based on user input during installation and add some placeholders

filipsobol:
During installation, user is asked to provide e-mail address and password for admin account. Having pre-defined credentials might be a security threat.


tnylea:
@Laran01 This is awesome! Thanks. I'll have to re-add this to the test functionality so it passes :) Thanks for the commit. I dig it 👍

filipsobol:
Sorry for adding two more commits to this pull request - I was going to do it separately (still learning how to do pull requests 😟 ).

src/VoyagerCommand.php
```diff
@@ -62,18 +65,31 @@ public function fire()
         Artisan::call('vendor:publish', ['--provider' => 'Intervention\Image\ImageServiceProviderLaravel5']);
 
         $this->info("Migrating the database tables into your application");
-        Artisan::call('migrate');
+        Artisan::call('migrate:refresh');
```
marktopper, line 68: `migrate:refresh` does not sound like that good of an idea.
How about people installing Voyager for existing projects?

marktopper:
Closing due to inactivity. Ping me to reopen.
Anyone else are welcome to make a pull request for this.

PR #56, opened 2016-11-01
Fix Issue #55

afsardo:
Fixing Issue #55 

Font awesome assets added.

PR #98, opened 2016-11-05
Fix font-awesome missing. Use publishable folder instead old asset fo…

handiwijoyo:
…lder, see PR #56


fletch3555:
I personally think we should just let @afsardo update the paths in his PR and treat this as a duplicate.

afsardo:
This is done and working better than my PR ^^ you can pull this and close mine :)

tnylea:
Thanks guys, actually using a custom font library for voyager. You can see the response here: https://github.com/the-control-group/voyager/pull/56

I've made sure to remove the occurrences of FontAwesome and replace them with the custom Voyager font class 👍

PR #106, opened 2016-11-06
Fixes artisan optimize with no db connection

### LINKED ISSUE #105: Including routes.php messes with artisan optimize on build servers
In `VoyagerServiceProvider`, routes.php get's included on application boot.
This runs the Schema::hasTable which needs an connection to a database. This does not work on build's or packages in a CI system where there is no database server:
```
  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[HY000] [1045] Access denied for user 'forge'@'localhost' (using password: NO)
```

We should check if env('DB_CONNECTION') !== null before including those routes.

WesleyE:
- Fixes artisan optimize breaking on CI servers with no db connection. 
Fixes #105

### DIFF
--- a/src/routes.php
+++ b/src/routes.php
@@ -14,7 +14,7 @@ Route::group(['middleware' => ['web', 'admin.user']], function () {
 		return view('voyager::profile');
 	});
 
-	if(Schema::hasTable('data_types')):
+	if(env('DB_CONNECTION') !== null && Schema::hasTable('data_types')):
 		foreach(TCG\Voyager\Models\DataType::all() as $dataTypes):
 			Route::resource('admin/' . $dataTypes->slug, 'TCG\Voyager\Controllers\VoyagerBreadController');
 		endforeach;


src/routes.php
```diff
@@ -14,7 +14,7 @@
 		return view('voyager::profile');
 	});
 
-	if(Schema::hasTable('data_types')):
+	if(env('DB_CONNECTION') !== null && Schema::hasTable('data_types')):
```
marktopper, line 17: I would rather like using the `config` method over the `env` method.

WesleyE: Yeah, this is a bit of a choice. It defaults to `'default' => env('DB_CONNECTION', 'mysql'),` in the default Laravel database.php config. So it might confuse people when it will still error out when you've got no .env file.

marktopper: I see your point there

tnylea:
Awesome. Thanks guys!

PR #120, opened 2016-11-07
Route prefixing

maiorano84:
As a feature, it would probably benefit users to be able to define the base route on which Voyager exists. As it stands, all routes are prefixed with "admin".

While this works, some users may not want something quite so obvious, or they may require a means by which they can change this prefix to something else in order to prevent conflicts on existing installations.

This branch seeks to correct that by adding a configuration value that users may change. All routes declared in each view make use of this by leveraging route aliasing. Tests have been modified to take all of this into account.

The Media Uploader Javascript has also been modified to use a class name which can take in an optional object that defines the base url on which the AJAX routes exist.

Testing in a local Valet installation yields satisfactory results, but may require further investigation before merging.

**EDIT:** All tests passing, Valet installation yields satisfactory results under `/voyager` route prefix.


maiorano84:
Looks like the latest commits addressing the icon fixes are causing conflicts. Will resolve manually later tonight.

tnylea:
Thanks man!

fletch3555:
For flexibility, can we change the config to open it up for any route configuration?

So instead of:

```
$router->group(['prefix' => config('voyager.routes.prefix', 'admin'), 'namespace'=>$this->routeNamespace], function(){
    include __DIR__.'/routes.php';
});
```

you'll end up with something like:

```
routeConfig = config('voyager.routes.config', array());
// set sensible defaults
if (!array_key_exists('prefix', routeConfig)) {
    routeConfig['prefix'] = 'admin';
}
if (!array_key_exists('namespace', routeConfig)) {
    routeConfig['namespace'] = $this->routeNamespace;
}
$router->group(routeConfig, function(){
    include __DIR__.'/routes.php';
});
```

**_Note: This is untested code... please test before committing it!_**

---

Edit: I'll look into this and submit my own PR.  Just realized this one was already merged ;)

maiorano84:
I'll also look into making this an interactive part of the installation process.

As it stands, if you run the installer and then change your route prefix, you'll have to also change the paths in the menu builder. In order for the paths to be correct, the route prefix needs to be set _before_ running the installer.

PR #131, opened 2016-11-09
Assets

Thiktak:
Change all asset ressources by the asset() version.
This will allow to the system to retrieve all ressources whatever the directory installation.

What have been changed ?
- css links
- images (icons, avatar, ...)
- js files

### DIFF
--- a/src/views/bread/edit-add.blade.php
+++ b/src/views/bread/edit-add.blade.php
@@ -135,6 +135,6 @@
       $('.toggleswitch').bootstrapToggle();
     });
   </script>
-  <script src="{{ config('voyager.assets_path') }}/lib/js/tinymce/tinymce.min.js"></script>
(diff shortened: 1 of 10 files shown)


tnylea:
So actually, if the URL was an https:// URL, this would still serve up the asset via http:// Let me know if there's another reason why this may be better and I can definitely merge it in.

I would think it would be best to keep the URL's to the assets relative. Let me know, maybe I'm missing something :)

Thanks 👍

Thiktak:
OK,

For now, the relative path is not working for me (as the routes) because I'm using subdirectories. And yes, I had not thought about the https.
I will think for a better solution :)

Edit: Are you sure it's not working with https ? According the laravel docuentation: asset() function _Generate a URL for an asset using the current scheme of the request (HTTP or HTTPS)_

Issue #121, opened 2016-11-07
[Improvement] Consider removing "eval" calls on VoyagerBreadController


agustindidiego:
Please consider removing "eval" calls like:
```php
eval('$dataTypeContent = ' . $dataType->model_name . '::find(' . $id . ');');
```
with the use of "class_exists" and "call_user_func", ex:
```php
$dataTypeContent = "";
if(class_exists($dataType->model_name)) {
        $dataTypeContent = call_user_func(array($dataType->model_name, 'find'), $id);
}
```
or creating a new instance of the model:
```php
$model = new $dataType->model_name;
$dataTypeContent = $model::find($id);
```

maiorano84:
I had PR #60 sitting for six days that addressed this issue without any indication of the change ever being implemented. By that point, the branch had gone stale and I closed the issue. I can reopen, or redo the branch (there weren't many changes).

Would any maintainer/author comment on this? I cannot think of any compelling reason why Eval should be used anywhere in this codebase. Its use is entirely avoidable, and unless there's something I'm missing, we should consider its use to be a security hole.

fletch3555:
@maiorano84 I recommend reopening it.  I know @tnylea has been pretty busy.

I'd say this issue can be closed as a dupe of PR #60

maiorano84:
I'm in the process of creating a new branch. PR #60 was pretty deeply flawed, in that the database evaluations being performed were being chained together. While calling single methods from a variable or `call_user_func` is easy enough, the closed PR did not take into account the chained calls and was erroring out in the database side. I'll follow up once I have this addressed.

agustindidiego:
Great! I wasn't aware of PR#60
Closing as a dupe of #60

PR #132, opened 2016-11-09
No eval

maiorano84:
Originally opened as PR #60 
Also raised in Issue #121 

Eval is widely considered to be bad practice for a multitude of reasons, many of which can be found on various forums and other relevant areas.

It is highly recommended that this code be changed to remove any and all instances of eval where its use is avoidable.

Changes needed to be made in the Database Controller in order to facilitate this endeavor, as chained methods could not be supported through simple variable functions. For now, the query itself is wrapped in an anonymous function and all database calls are made explicitly.

**This does not address preexisting database creation/update issues that exist on the master version**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+.idea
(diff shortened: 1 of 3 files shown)


tnylea:
Thanks man. Pulling in and testing and I'll add this to the newest version today 👍

PR #133, opened 2016-11-10
Cleanup

maiorano84:
**These are purely non-functional changes.**

The purpose of this branch is to bring the project in line with PSR-1/2 standards, as well as correct any minor instances of code smell and inconsistencies.


tahq69:
definitely should be merged; clean code - better readability!

tnylea:
Thanks for resubmitting 👍

PR #146, opened 2016-11-11
Register dependency providers in VoyagerServiceProvider

jorenvh:
Register the intervention/image service provider from in the VoyagerServiceProvider. 
Users of the package don't have to register that provider manually anymore.

I also removed the provider in the installation instructions in the readme.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ Add the Voyager service provider as well as the Image Intervention service provi
 
 ```
 TCG\Voyager\VoyagerServiceProvider::class,
-Intervention\Image\ImageServiceProvider::class,
 ```
 
 Lastly, we can install voyager by running
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -15,6 +15,13 @@ class VoyagerServiceProvider extends ServiceProvider
     private $routeNamespace = 'TCG\\Voyager\\Controllers';
 
     /**
+     * Dependency providers to regiser.
+     */
+    private $providers = [
+        \Intervention\Image\ImageServiceProvider::class,
+    ];
+
+    /**
      * Bootstrap the application services.
      *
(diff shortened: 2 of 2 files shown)


tnylea:
Thanks man. I'll merge this in. Quick question though, wouldn't we want to add the Voyager Service provider to be automatically included as well? Let me know :)

Thanks.

jorenvh:
@tnylea that's not possible the voyager provider is the main provider for your package, you can only do that for your dependencies. Otherwise your provider will never be loaded.

maiorano84:
@jorenvh That's correct, Service Providers cannot be self-loading. I like the fact that the Intervention dependency is included with the primary Voyager provider.

I'm wondering, since I haven't tested this myself:

If the Intervention library has already been loaded by a previously-declared library, would that necessitate a check to see if the library has already been loaded? Or would Laravel skip over it, recognizing that it's already been registered?

jorenvh:
@maiorano84 good question! I haven't taught about that, we should test that indeed. However I used this code before in my own packages or services within Laravel and there was no issue reported about this.

But we should definitely test that use case.

arcanedev-maroc:
@tnylea This is a duplicated PR: #142

jorenvh:
@arcanedev-maroc I didn't check for an existing PR, sorry for that. 

Still think my solution is more future proof, when there are more dependencies you only have to add them to the providers array, so less duplicate code.

arcanedev-maroc:
@jorenvh Yeah i agree but for the time being it has one dependency that you need to register its service provider.

And my solution has less code (1 line) then your solution. 😉

jorenvh:
@arcanedev-maroc fair point ;)

jorenvh:
Since this is a duplicate for PR: #142 and their is only one dependency at the moment you can better use the solution provided by @arcanedev-maroc.

PR #150, opened 2016-11-12
Fix #148

### LINKED ISSUE #148: [Fix] Sidebar remember expand/collapse state
Would be great to remember if sidebar is expanded/collapsed.

marktopper:
Saves wether or not the menu is expanded into a cookie.


marktopper:
Fix #148

akazorg (commented):
I was thinking about this issue, why don't we use HTML Storage?
Something like this example: https://gist.github.com/simonbingham/3994802

marktopper:
@akaZorg, that is a pretty good idea.

I will update my pull request to support using locale storage.
And then it will be optional from the configuration.

Will keep you updated.

akazorg:
Nice that you like it.
About the optional configuration, I guess we don't need it since this behavior should persists across the session. Please correct me if I'm wrong, maybe I didn't get the full picture.

Keep up good work :+1:

marktopper:
@akaZorg: If you say so ;)
It's easier for me then.

Will be back with this feature later today.

akazorg:
Hello @marktopper, any advance on this?

akazorg (approved):
Tested, is working.

akazorg:
Your idea is working and it's much better.
While using LocalStorage is impossible to get the stored value before page load, creating a unpleasant behaviour xD
Good job @marktopper , please merge this!!

marktopper:
I had some emergency calls, so didn't had time for trying out LocalStorage.

However now you are for using cookies or what?

tnylea:
Kick ass man! I dig this. Merging now 👍

akazorg:
@marktopper cookies are good!! never had problem with them.
Maybe we could save user preferences on a db session, it can be handy for other stuff to.

marktopper:
@akaZorg That will require a api call to the endpoint. Still a good idea for people which always browse the site from different devices. However I personally prefer that it does not need to make an API call to the endpoint, so I would however stick to this, but I could make it configureable.

marktopper:
However I really like the idea for making user preferences for other things, I will go ahead and make some system for this in a pull request later.

Issue #108, opened 2016-11-06
[Feature Request] Use views for menu builder


marktopper:
Would it be possible in future to use views for the menu builder.

That makes it way more customizable.

__EDIT__: If approved, then I do not mind making it into a pull request.

fletch3555:
What do you mean by "use views"?

marktopper:
@fletch3555: https://laravel.com/docs/5.3/views

tomicakorac:
@marktopper I think what @fletch3555 asked means what did you mean by "use views for the menu builder". How do you want to use views in the menu builder? Can you please explain your question to us?

marktopper:
I will try to make a pull request this weekend to show what I mean.

It's like Laravel's Pagination template moved to using views.
[Read how it works now](https://laravel.com/docs/5.3/pagination#customizing-the-pagination-view).

marktopper:
I also just noticed another small issue with the current way that the menus are builded.

There are no eager loading, so it could take a while to load for sites having many menus with many sub menus.

marktopper:
First implement have been done at #154.
Currently only for the menu of `admin_menu`-type.
What do you think?

Issue #107, opened 2016-11-06
[Bug] Admin Sub Menu Items Not Workign


marktopper:
When using the menu builder on the admin menu to create a sub menu.

Then it can not open the sub menu on the admin panel.

handiwijoyo:
PR #74 fixed this issue, but only for 2 level menu

akazorg:
I created PR #74 working with 2nd levels only because I needed that fix right away.

If we take a look at [src/Models/Menu.php](src/Models/Menu.php), we realize that views are hard coded on the model, so I guess most of the code will change. I'm sure @tnylea have plans for this?

I also noticed that some of the menu items (Tools and Settings) are coded at [src/views/master.blade.php](src/views/master.blade.php), if we moved this to the database, we would be able to manage this items using the Menu Builder.

marktopper:
@akaZorg: We could indeed move the menu items from the view to the database. It just requires that 2nd levels works first.

Also I think it would come in handy if the menu where build using views instead, that makes it way more customizable. I do not know if @tnylea have any plans so far.

akazorg:
Hello, (sry about delay).
2nd levels are working at the moment, just from the 3rd + is not. I would advance more work on that, but I need confirmation from project maintainer - hey @tnylea, are you alive? :dancing_men: any plans for the menu?

I noticed you post issue #108 regarding views, we should do it indeed!

marktopper:
This works now, but only for 2nd level menus.

Let's keep this open until more levels are supported.

I might try to solve this somehow in the near future.

PR #154, opened 2016-11-12
Use views and eager-load menu items

marktopper:
**DO NOT MERGE THIS YET - STILL WORKING ON IT**

- [x] Eager load menu items
- [x] Make view for `admin_menu` type.
- [ ] Make view for `admin` type.
- [ ] Make view for `bootstrap` type.
- [ ] Make view for `default` type.

Currently there is a fallback, so if you enter a `type` which does not have a view, then it will fallback to the original code. However this will be removed once done, since I'm gonna make a view for all of those cases.

It will be backwards compatible, however you might need to update your `voyager` config file to begin using views.


marktopper:
Fixes #108 

Related to #107 

> Now the Javascript only needs to support unlimited childs, since the view does support it. I might work on this later on.

marktopper:
/ping @tnylea

Would you approve the idea? Would like to know your comments on it before continuing?

akazorg:
Hey @marktopper, you can remove the item "Eager load menu items", it was updated already.
Cheers

marktopper:
@akaZorg, I think you did put in the wrong reference id.

akazorg:
can't find the PR xD it's ok, not important

marktopper:
@akaZorg I will remove it then ;)

brandonferens:
Well, snap. @marktopper this is a much better approach than what I did #196. The menu builder code is pretty confusing and didn't allow for custom views.

We also refactored a few things the same...

marktopper:
@brandonferens. The only problem with this is that my code is breaking changes.
It requires that the user add the new configuration to the configuration in order to work.
I will continue on this and hopefully have something ready to be tested soon, however I am open for feedback and ideas on things to changes, which you might have, since you created the same thing ;)

brandonferens:
Ah, yes. Check out https://laravel.com/docs/5.3/packages#resources and scroll down to `Default Package Configuration`. It allows you to use the default config form within the package, and if the user decides to override any of the values, they can. I've done this a few times. Let me know if you would like some help with it!

marktopper:
Thanks @brandonferens. I will make it work.

Also if you want to help you can just do a pull request to my fork of this (branch `issue-108`).

marktopper:
I'm going to close this and merge #196 instead.
However I will make a new pull request moving the current menu builder code into separate views, just like I started on here for this one.

PR #155, opened 2016-11-12
[Bug] fix menu builder url;

tahq69:


tahq69:
Error in admin panel appeared because of outdated config file;

marktopper:
Could you separate those commits into multiple pull requests instead?

PR #167, opened 2016-11-14
Modified BREAD (If doest have Model)

putera:
Small modification on BreadController if the database table doesnt have a Model


tnylea:
Good stuff man :) Thanks for the contribution.

PR #169, opened 2016-11-14
Fixed updating tables

brandonferens:
Updating tables was failing due to some timestamp issues, incorrect routes, and such.

I made some small tweaks to formatting so I could read it all a bit better. I would really love to see this whole thing broken out into collections and separate method calls. It's quite a bear to work on ;)


tnylea:
Thanks man. Actually added that route fix in another push, could you resolve the issue and then I'll merge in this request. Appreciate it 👍

PR #170, opened 2016-11-14
Fixed delete actions

fergthh:
This pull request solve a bug in delete actions of the forms on the lists.

Steps to reproduce the issue:

* Try deleting an item, then **cancel**.

* Now, try deleting the same or another item again.

fix_delete_actions

The form action url looks something like this.

In the first click

captura1

And when we click again

captura2

I've only changed the way it identifies and analyzes the url of the form action.


tnylea:
Man! You rock!

Thanks for the awesome GIF representation of this bug.

PR #179, opened 2016-11-18
Database controller patch

brandonferens:
Here is the updated pull request for the database controller. Conflicts have been fixed.


src/views/tools/database/edit-add.blade.php
```diff
@@ -91,7 +91,7 @@
     <div class="page-content container-fluid">
         <div class="row">
             <div class="col-md-12">
-                <form action="@if(isset($table)){{ route('voyager.database.edit_table', $table) }}@else{{ route('voyager.database.create_table') }}@endif"
+                <form action="@if(isset($table)){{ route('voyager.database.update_table') }}@else{{ route('voyager.database.store_table') }}@endif"
```
marktopper: Should it be `route('voyager.database.update_table', $table)`

marktopper:
👍 However, I have to look into the failing test.

marktopper:
Somehow the route `GET /admin/database/edit-categories-table` returned a `500 Internal Server Error`.

brandonferens:
Yeah, I got that error as well. Been off for Thanksgiving so haven't had a chance to look it over. I'll check it out now.

marktopper:
Let me know once you have looked at it.

brandonferens:
I'm confused with what's going on on my system. I need to just kill this PR. It looks like there have been enough changes within the package to warrant that. What I would really like to see, and perhaps I will work on it, is a much better system for doing anything database. All of that code should not be within the controller. It should all be extracted out and placed into jobs or other classes. All the update method should do is tell the system to update, then the jobs take over and do what is necessary for each row in the db.

marktopper:
Indeed, we really need to work on the controllers.

brandonferens:
I started working on the database controller. I'm curious how you feel about extracting those different pieces of logic into Traits? Then just include those at the top then have access to something like `$this->renameTable($originalName, $tableName)`, for example. I don't really see a reason to need an entire new class to instantiate.

The other option would be to setup a job for each piece needed. What are your thoughts?

marktopper:
👍 Nice, looking forward to merge.

PR #185, opened 2016-11-20
FIX - env function call in routes.php

YannikFirre:
Hello,

I propose not to use env() in routes but config().

### DIFF
--- a/src/routes.php
+++ b/src/routes.php
@@ -17,7 +17,7 @@ Route::group(['middleware' => ['web', 'admin.user']], function () {
         }
     ]);
 
-    if (env('DB_CONNECTION') !== null && Schema::hasTable('data_types')):
+    if (config('database.default') !== null && Schema::hasTable('data_types')):
         foreach (TCG\Voyager\Models\DataType::all() as $dataTypes):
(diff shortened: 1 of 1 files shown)


marktopper:
There is a reason for the madness.

I reviewed it here as well #106

YannikFirre:
@marktopper Yeah, i understand in local but if you don't use the .env file in production, all is broken :'(

My use case : i put .env in production first time, i execute artisan config:cache then i delete the .env file.

marktopper:
I agree that the current solution does not work perfectly, but neither does this one.

I however can not figure out what a perfect solution would look like.

marktopper:
I myself would prefer a try-catch instead of this.

YannikFirre:
Sorry, but i don't understand in what situation it doesn't work perfectly ? :P 

`config('database.default') = env('DB_CONNECTION', 'localhost')`

marktopper:
The thing that #106 tries to do, is to only run the database call in routes if the system have been installed. However this is the case when `env('DB_CONNECTION')` is not `null`. However doing the using `config('database.default')` will never return `null`, not even when Voyager have not been installed.

marktopper:
Closing, however still looking for another to handle this.

PR #188, opened 2016-11-22
Add multiselect field type and extra fields (relationships) to bread

joshuatf:
Many to many relationships are an issue right now.  This allows extra fields to easily be added to bread that aren't initially included from the columns and define relationships.

Since relationships can't easily be parsed from a Laravel model, defining them in a static method is probably best.    With this pull request it can be done by defining an array of items with the relationship method name as the field.  Multiple selects are updated in the database using Laravel's sync method.

```php
public static function adminFields()
    {
        return [
            [
                'Field' => 'roles',
                'Type' => 'multiple_select',
                'Null' => 'YES',
                'Key' => '',
                'Default' => null,
                'Extra' => '',
            ]
        ];
    }
```

Relationship options for multiple select are defined in the BREAD settings by key and label used for the select option.  These values are attributes of the class defined by the field above.

```json
{
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
```

This could pretty easily be extended similarly for one to many relationships to avoid defining relationship options directly in custom views (as is done now for Category in the Post demo example).


joshuatf:
Sorry, I should have made a separate branch for the middleware.  

The model middleware much like the admin middleware can be customized from settings.  This will allow a default package configuration if you choose to keep roles in, but it will also allow third party role/permission packages.  Permission on the index view also toggles menu visibility.

Example usage:

```php
'model_permission'` => function($action)
    {
        $user = Auth::user();
        if ($user->hasRole('super_admin')) {
            return true;
        } elseif ($user->can($action)) {
            return true;
        }
        return false;
    },
```

The $action variable is 'admin.{model_name}.{route_action}' so posts index would would be:

 `admin.posts.index`

Personally I think the roles should be left to a separate package so that Voyager isn't trying to be too many things, but for the sake of demonstration purposes and easy setup, it makes sense to leave in as an option.  With that said, this method would also allow you to run a loop on the available fields to create a GUI for setting model permissions by role.

marktopper:
Great job 👍 
However, are you able to great some test cases for this?

marktopper:
We also have to discuss wether or not this should be backwards compatible.
Since now current users of Voyager must add the two new configurations fields in order to make this work, else their application would fail.
I will try to get in touch with @tnylea for this.

joshuatf:
@marktopper Glad to write up some tests when I have time.  

After working actively with this the past week, there are no issues with the basic n-n relationships, but conditional n-n (e.g. using a where clause in the relationship) doesn't work since the select options are generated by running an ::all() on the related model.  I'll try to revise this as well in the next push.

As for the middleware, I would think breaking changes are okay since we're still not at 1.0 yet.  The alternatives are:

1. Return true if config not set (potentially dangerous if dev is unaware of changes)
2. Return false if config not set (probably more confusing than error currently thrown)
3. Use the previous middleware conditions if config not set (also throws an error if the user hasn't migrated the demo roles)

Documentation should mostly solve this issue and I don't think any of the above are great solutions, but if you have a better idea, let me know.

marktopper:
I have been in touch with @tnylea and we come up with the solution that this will make it to a minor version change. So once this is ready we will bring up the release of v0.10.0.

So don't mind the backwards compatible.

marktopper:
Whenever your ready I can merge this to a v0.10 branch.

marktopper:
Could you fix the merge conflicts or should I go ahead and do this and then merge into `master`?

joshuatf:
Sorry, pretty busy today, but might be able to get to it tomorrow.

In regards to custom querying I mentioned before, this will have to be added to options later.  Does not appear there is a way to determine these directly from the Laravel model.

marktopper:
We will figure out a way.

joshuatf:
Fixed conflicts.  Will need to update in accordance with new PR styling.

As for the query conditions, Laravel does not provide a way to determine these.  From what I understand the possible solutions are:

1.  DocBlock param - not terrible, but not ideal
2.  Regex - too messy
3.  Manual config - I think placing in the options will work best.

```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "query": "WHERE role='admin'"
    }
}
```

brandonferens:
Regarding the style, and not sure where to put this comment, but I personally prefer this:
```
$foo           = 'bar';
$reallyLongFoo = 'bar';

$foobar = 'Foo ' . $bar;
```
Instead of:
```
$foo = 'bar';
$reallyLongFoo = 'bar';

$foobar = 'Foo '.$bar;
```
The first one just breathes a bit more and is easier to read. But that is my preference. Opinions?

akazorg:
Bread Relationships are really needed, great job @joshuatf, looking forward to use it.

marktopper:
The style fixes will automatically be applied after a merge.

tnylea:
@joshuatf Testing this out in the latest version :) and I'm going to add some of this to the documentation. The only thing is I'm having a problem when trying to add a relationship.

Let's say I want the `author_id` to map to the `users` table. How might I write the relationship in the JSON file?

```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "query": ""
    }
}
```

If you could let me know that would be super helpful. I'm trying to go through the code so maybe I'll figure it out 👍 

trying to figure out how this one works:

```
<?php $relationshipClass = get_class(app($dataType->model_name)->{$row->field}()->getRelated()); ?>
```

Is there more content that needs to be added to the model? Keep me posted. 

Thanks!

joshuatf:
@tnylea You'll just need to add the user relationship to your model.  That line just grabs the relationship class based on the field name provided.  So if your field was author, in your model you would just define your relationship like normal:

```
    public function author() 
    {
    	return $this->belongsTo('App\User');
    }
```

The key and label are the attributes on the foreign table.

However, this PR only works for many to many relationships.  I did not want to add 1-n relationships until getting #200 updated as it's going to produce a lot of unnecessary code that we'll have to rewrite soon after.  I will most likely try and handle that issue this week if no progress has been made so far.

joshuatf:
@marktopper Why where the permissions and roles configuration options removed?  Seems like a really odd choice to not go agnostic with the user/roles system.  This will prevent a lot of users from using their own role systems or third party packages.  Is there a benefit to this?

Issue #190, opened 2016-11-22
[FatalError] DatabaseController Handling of Timestamps


joshuaajones:
FYI, this is a **fantastic start to a great package** and I hope to help contribute soon.  Still getting my feet wet with it.  This is pretty much the foundation of the Database / Model interaction, and thus critical to the success of the package.

That said, I had to do a completely fresh install to confirm this issue out of the box.

### Reproduction steps:
1. Fresh Laravel 5.3
2. Fresh Voyager
3. Login > Tools > Database > 
4. Edit any table with timestamps 
    * Simply updating (failed)
    * Adding a column (failed)
    * Deleting one or both timestamps (failed)
5. Create table with manually entered nullable timestamp column(s) ie: 'created_at' (failed)
6. Create table with available UI button for timestamps (success)
    * Repeat step 4 with this new table (failed)
   
It seems like all the failures resulted in an issue with handling timestamps, 
and the most commonn exception was:

```
FatalErrorException in VoyagerDatabaseController.php line 210:
Call to a member function nullable() on null

in VoyagerDatabaseController.php line 210
at FatalErrorException->__construct() in HandleExceptions.php line 133
at HandleExceptions->fatalExceptionFromError() in HandleExceptions.php line 118
at HandleExceptions->handleShutdown() in HandleExceptions.php line 0
at VoyagerDatabaseController->TCG\Voyager\Controllers\{closure}() in VoyagerDatabaseController.php line 116
at VoyagerDatabaseController->TCG\Voyager\Controllers\{closure}() in Blueprint.php line 69
```

**Looking forward to the solution!**

marktopper:
This also happens when creating a new table with a nullable timestamp.

joshuaajones:
Thanks Mark,  I didn't realize my typo there - updated the step 5 description.

marktopper:
Just created a pull request to fix this issue, however there is another [issue](https://github.com/laravel/framework/issues/16526) with the Laravel Framework which will have to be resolved first.

joshuaajones:
Nice catch, that was going to be my next issue!  DBAL not registering the timestamp Type.  Odd that's gone on so long unnoticed!   Thanks for the PR.  Implemented on my end, could not reproduce the same error as before.

marktopper:
Glad to hear.

marktopper:
The other issue that we have come by is because Laravel uses Doctrine/DBAL.

However they do not want to support MySQL-specific column types like `timestamp`.
[...]

PR #194, opened 2016-11-23
Handle timestamps correctly

marktopper:


marktopper:
This fixes #190.

However, after this there is another issue related to Laravel.
[Read more.](https://github.com/laravel/framework/issues/16526)

marktopper:
The other issue that we have come by is because Laravel uses Doctrine/DBAL.

However they do not want to support MySQL-specific column types like `timestamp`.
Therefor I had to build a [Laravel Package](https://github.com/marktopper/doctrine-dbal-timestamp-type) for it, but this solves the other issue.

PR #196, opened 2016-11-23
Build menu based on custom view

brandonferens:
With this PR, a user can build their own custom menu. Create a custom menu view and specify that as the `$type` parameter!

For example, a user can place a custom menu under `views/menus/custom-menu` and reference it with: 

```
{!! Menu::display('my-menu', 'menus.custom-menu') !!}
```

If the view doesn't actually exist, it just returns the `buildOutput()` method instead.

I also cleaned up a some of the code in the `display()` method. It's been refactored so there aren't any nested if statements and also removed any else statements as well. It's a little less complicated to read.

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -23,43 +23,47 @@ class Menu extends Model
         $menu = $instance->where('name', '=', $menu_name)->first();
 
         $menu_items = [];
+
         if (isset($menu->id)) {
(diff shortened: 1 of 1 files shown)


marktopper:
I have closed my own pull requests of the same.
However I will make a new pull request moving the current menu builder code into separate views.

PR #207, opened 2016-11-26
Disable datatables default order (sort)

handiwijoyo:
Datatables by default will sort the first table column which will cause Eloquent orderBy ignored.


marktopper:
Nice catch!

I will just go ahead and test your solution before merging.

marktopper:
Tested and works.
Great job 👍

Issue #35, opened 2016-11-01
"Publish" Routes


fletch3555:
Similar to Issue #23, would it be possible to restructure how the routes are included so we can have control over where they're included?  I'm working on a multi-tenant setup and only want this admin page available for one of them.

Great project by the way! I'd love to help contribute in some way.

marktopper:
You can now change the route prefix in the configuration for all routes generated by Voyager.

Please ping me if there was anything else you wish to do.

fletch3555:
@marktopper along with the prefix, is there a way to change other route properties, such as subdomain? Simply changing the prefix doesn't really solve my multitenant problem.

Thanks!

marktopper:
Reopened

fletch3555:
Taken care of a different way as part of v0.10

PR #219, opened 2016-11-28
Refactor routes

fletch3555:
This is a fairly substantial rewrite of how routes are handled.  I've done some testing locally and all seems to be functioning identical to before, so I don't believe I've introduced any bugs with this.

I'll provide a docs update to go along with this if you're good with it @marktopper and @tnylea


fletch3555:
This would take care of my needs as part of #35 by allowing developers to import Voyager routes under whatever route configuration they want (subdomain, prefix, etc.)

fletch3555:
@marktopper, Could you take a look at why TravisCI is failing here? As far as I can tell from my testing, this is building the correct route ("voyager.dashboard" = "http://localhost/admin"), but it seems to be seeing the default laravel homepage.

publishable/routes/voyager.php
```diff
@@ -0,0 +1,86 @@
...
+|
+| This file is where you may override any of the routes that are included
+| with Voyager.
+|
+*/
+
+Route::group([
+    'namespace'  => '\\TCG\\Voyager\\Http\\Controllers',
+    'middleware' => 'web',
+    'prefix'     => 'admin',
```
marktopper, line 16: Prefix is hardcoded now.

publishable/database/seeds/MenuItemsTableSeeder.php
```diff
@@ -46,7 +44,7 @@ public function run()
                 'id'         => 3,
                 'menu_id'    => 2,
                 'title'      => 'Posts',
-                'url'        => "/{$prefix}/posts",
+                'url'        => route('posts.index'),
```
marktopper: Route names in Voyager should be prefixed with Voyager. This seems to be missing in many lines, please go though.

marktopper:
However, I just pulled down your branch, and I myself see the frontpage instead of the dashboard.
I'm simply being redirected.

I recommend that you go though the routes and make sure all of them uses the `voyager` prefix. Then if you still have problems with this, I will look into it.

fletch3555:
I followed the TravisCI script step by step and was still getting the dashboard when I tested. Guess I'll have to retest tomorrow

marktopper:
It seems that Travis does not get the the database seeded correctly.
It leaves out the following tables:
- menu_items
- roles
- settings
- user_roles

Since the seeders are ran in this order:
```
Seeded: CategoriesTableSeeder
Seeded: DataTypesTableSeeder
Seeded: DataRowsTableSeeder
Seeded: UsersTableSeeder
Seeded: PostsTableSeeder
Seeded: PagesTableSeeder
Seeded: MenusTableSeeder
Seeded: MenuItemsTableSeeder
Seeded: SettingsTableSeeder
Seeded: RolesTableSeeder
Seeded: UserRolesTableSeeder
```

I think the problem is in either the `MenusTableSeeder` or `MenuItemsTableSeeder` seeder class.

marktopper:
The issue is in the `MenuItemsTableSeeder`, and are cause that it uses `route(something)`, however the routes for Voyager are not loaded when trying to seed the table, which throws an exception.

marktopper:
I just got it to work by requiring the routes in the installation command.
However this is a bad idea to use `route`, since that will generate a full url like `http://localhost/admin` from `route('voyager.dashboard')` and input it in the database as `slug`.
However then it's not really a slug anymore. What are peoples thoughts on this?

fletch3555:
@marktopper Finally got Travis working!  Not 100% sure I like the solution, but it works.

marktopper:
Nice job. Can you fix the conflicts and let me know, then I will merge into `v0.10`.

fletch3555:
Working on the conflicts now.  This will need a docs update before we tag/release it

fletch3555:
conflicts resolved. waiting on Travis to make sure I didn't break anything

PR #224, opened 2016-11-29
Use dynamic namespace for "App"

### LINKED ISSUE #73: Problem with rename the App name
I rename the App via 
`php artisan app:name Pollr`

When I install via voyager:install something wrong with service call. I checked in directory
Vendor\tcg\voyager\src\controllers
most of them still called App as the prefix, why not adapt to the new App Name?
`use {App}\Http\Requests;`
`use {App}\Http\Controllers\Controller;`

fletch3555:
Looks like Voyager is already fairly well de-coupled from "App".  This pulls in the "App" namespace dynamically when prepopulating the BREAD form.

Fixes #73


marktopper:
Great job 👍

PR #227, opened 2016-11-29
Add validation error messages

handiwijoyo:
Adding validation error message, and change the JSON to use validation group.

Old validation:
```
{
    "rule": "required|max:12"
}
```

New validation with error messages:
```
{
    "validation": {
        "rule": "required|max:12",
        "messages": {
            "required": "This :attribute field is a must.",
            "max": "This :attribute field maximum :max."
        }
    }
}
```


marktopper:
This have breaking changes. However this have been added to `master` which now is the future for `v0.10` release.

Issue #232, opened 2016-11-29
Error while creating table with Voyager


sdevid:
- Laravel Version: 5.3.*
- Voyager Version: 0.9.33
- PHP Version: 5.6.16
- Database Driver & Version: MySQL (10.1.9)

### Description:
 created a new laravel project and installed 0.9.33 and when i trying to create table with voyager, i got the message:
 _Exception: Missing argument 2 for TCGVoyagerHttpControllersVoyagerDatabaseController::buildQuery(), called in C:laragonwwwoyager2endor cgoyagersrcHttpControllersVoyagerDatabaseController.php on line 38 and defined_ 
And the table not created.

### Steps To Reproduce:
http://www.awesomescreenshot.com/01566ydk0d

brandonferens:
@sdevid I'm working on it and will have a PR is just a moment.

tnylea:
Just got it merged in. I am going to do a few other merges and update a few tests and I will tag a new version. Thanks @brandonferens 👍

PR #233, opened 2016-11-29
Fix for bug #232

brandonferens:
This was mentioned in #232. Added parameter default to ensure table creation.


tnylea:
Thanks @brandonferens 👍

PR #241, opened 2016-11-30
Set NULL for empty foreign keys in Bread Controller

wxs77577:
Not only Seeders but also BreadController.Be sure the `0` or EMPTY STRING of foreign key (such as `parent_id`) can be changed to `NULL`


marktopper:
Not all `*_id`s must be a foreign key.

PR #244, opened 2016-11-30
Roles update

akazorg:
Hey Voyagers!
Like some of you, I needed a user role without the permission for breaking the system, (not able to manage Roles, Menu Builder, Database and Settings).

This PR will sort out this issue, and fixes a few more things.

**1. New Role**
Was created a Role named "root" that holds full permissions, while the role "admin" have restricted access. A user account "root" was also added, we can test login with: "root@admin.com" and "admin@admin.com". If you don't like the names, we could use "admin" for full rights, and have another role, like "manager" for the normal access...

**2. Role Assigned on BREAD**
In table data_types was added a field named "role", that is used at the Middleware Validation.
For changing a role, goto Database / Edit BREAD, scroll down and you'll find a role input.

**3. Role Validation**
Validation is using the existing VoyagerAdminMiddleware, was added an extra parameter (role) used for the verification. The role defined in BREAD, will be used for controlling this access.

**4. User Role Manager**
Added a view for Users, with available roles inputs, however this isn't saving yet, we need to handle relationships. Until we don't find a way for managing User Roles, maybe we can delete this file?


fletch3555:
I'd prefer to manage access through permissions, instead of directly checking for hard-coded role names. Permissions can be given to roles, and Users can be given a role.  #245 allows for the creation and management of these permissions.

akazorg:
@fletch3555, this update adds some fixes, but we still need to be able to save Users Roles: 
screenshot_20161130_160934

Roles are hard-coded in routes web.php **only**, if you look at Middleware it was removed and it's now being processed dynamically, for a given role parameter.
When you manage the BREAD, you can define the role:
voyager-bread

I personally don't see the point of having a killer RBAC system like #245 is developing, but I understand we all have different needs.

marktopper:
The permission system in `v0.10` supports this.

akazorg:
Thank's, I was about to close this.
It would be more interesting to develop on top of the BREAD, instead of coding new sections.

jcanchor:
To others looking at how to manage full control, I normally see this role referred to as a SuperAdmin. So in my Voyager dashboard I did just that, created a role called SuperAdmin and in Roles it is a pretty simple setup. Just check "Select All" at the top and you have full permissions on every table. So I agree with Fletch and Mark.

akazorg:
Hi @JCanchor, this was long ago, but it's true you have a solution to manage it.
The only down side, if you need to handle a huge list of permissions, then you will be willing for a multi-roles system.
Anyway, when the time came, we will have multi-roles :)

Issue #240, opened 2016-11-30
[Feature]New User-Role-Permission


handiwijoyo:
I'm currently working on new role-permission based user. 

List of changes:

- [x] User can only  have a role. (Please let me know if you need multiple roles on user / use cases).
- [x] Remove role_user table.
- [x] Add admin role to default admin user.
- [x] Add admin_role to config to change default admin role.
- [x] Add permissions & permission_role table.
- [x] Add permissions generator during BREAD creation.
- [x] Manage Role's permissions
- [x] Add role selection to User interface.
- [x] Add permission checker on Bread Controller

Ref #33 #110 

**Generate Permission on BREAD editor**

![image](https://cloud.githubusercontent.com/assets/4060094/20763624/6c658d8a-b75d-11e6-8bb4-f4d69f2efbfa.png)

**Adding Permissions to Role**

![image](https://cloud.githubusercontent.com/assets/4060094/20771488/334c5ef6-b77c-11e6-9455-1065eb5ed7d7.png)

**Select User Role**

![image](https://cloud.githubusercontent.com/assets/4060094/20771633/cc924c10-b77c-11e6-8249-a31e884397a8.png)

fishmad:
Nice, I think multiple roles would be better.

handiwijoyo:
@fishmad could you please give example of use cases for multiple roles?

PR #245, opened 2016-11-30
Add user-role-permission-builder #240

handiwijoyo:
See #240 for details.


fletch3555:
Tested briefly and looks great!  I'll let @marktopper and/or @tnylea take a look as well before this gets merged in.

My only concern is how big the permissions list may get.  Is there a way we could refactor the UI for that so it's a little more compact? (Can be a new issue, doesn't have to be built into this PR)

handiwijoyo:
@fletch3555 Thanks. This PR will be only basic implementation of permissions. Once it merged, we can add more patch/feature.

For example, I want to move the permission checking to somewhere else https://github.com/the-control-group/voyager/blob/4c89f10397227c7a46af4a8d394a31222d7ed3a9/src/Http/Controllers/VoyagerBreadController.php#L421

akazorg:
Hello, nice job here.
Personally I would use roles only, like we have now, but it's just my opinion.
I think users should be able to have multiple Roles, and not to have a single role associated with.

marktopper:
Can you fix the conflicts? Then I will merge in.

handiwijoyo:
@marktopper resolved!

PR #247, opened 2016-12-01
changing route prefix

unabomber87:
hi,
When changing routes prefix on config/voyager.php the url on the admin menu keep the same route prefix (admin) which is saved on the DB. I have made some changes to make the change of the prefix change automatically the url on the admin menu and removed the prefix from the seeder class.


marktopper:
Not an ideal solution

PR #249, opened 2016-12-01
Controllers command

marktopper:
This will make it possible to create Voyager controllers easy using `php artisan voyager:controllers`.

You can customize the location by using the `location` parameter.


marktopper:
Gonna remove the location parameter and use it from the config instead.

src/Commands/ControllersCommand.php
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @return void
+     */
+    public function fire()
+    {
+        $stub = $this->getStub();
+        $files = $this->filesystem->files(base_path('vendor/tcg/voyager/src/Http/Controllers'));
+        $location = $this->option('location');
+
+        if (!$this->filesystem->isDirectory(base_path("app/{$location}"))) {
```
fletch3555: May want to use `app_path` instead as it's already namespace-aware

PR #250, opened 2016-12-01
Add config option to allow override voyager master view

handiwijoyo:
_config/voyager.php:_

image


akazorg:
This is a nice one. I think we should also be able to edit all views.

fletch3555:
Does [Overriding views](https://the-control-group.github.io/voyager/docs/#voyager-docs-customization-overriding-views) not already allow this?

akazorg:
@fletch3555 no, this is hard-coded.. all blade files include `@extends('voyager::master')`

fletch3555:
Yes, but Laravel already supports this OOTB: https://laravel.com/docs/5.3/packages#views

handiwijoyo:
@fletch3555 I never notice that! Thanks for pointing that out. I will close this.

PR #265, opened 2016-12-02
Fix default double backslashes model name 

handiwijoyo:
Stripslashes to use single backslash


marktopper:
Not the right fix.

The right fix would be to remove the backlash after `$this->getAppNamespace()`.
Since that returns with a backslash like `App\`.

handiwijoyo:
@marktopper fixed

PR #288, opened 2016-12-03
Refactor install command and seeders

marktopper:


marktopper:
Almost done refactoring.
Also changed the usage of the installation command to accept the `with-dummy` option instead of seeding dummy data by default.

PR #291, opened 2016-12-04
Update edit-add.blade.php

yhif:
~~fix Undefined offset: 3~~

Update:

In menu tools > database add new table and add timestamp fields will appear **Undefined offset: 3** exception.

Voyager version 0.9.34.

Repair:

Modify the **edit-add.blade.php** file to line 263
`$('#' + unique_id).find('.fieldType').val('timestamp').attr('readonly', 'readonly').attr('disabled', 'disabled');`

Remove the disable attribute, replace with
`$('#' + unique_id).find('.fieldType').val('timestamp').attr('readonly', 'readonly');`


marktopper:
Please describe what error it fixes and how to throw this error.

Also, what version of Voyager is this on.

marktopper:
This seems to be fixed in `v0.10` already.

yhif:
@marktopper 👍

PR #293, opened 2016-12-04
Show version on admin panel

marktopper:
**Open for discussion**

This is just one of many ways to do this, wondering if this is a solution that would be good enough.
This however have to be modified manually every time we release a new version.

skaermbillede 2016-12-04 kl 12 32 34

**EDIT:** This will be shown on the bottom of the Voyager admin panel.

### DIFF
--- a/resources/views/master.blade.php
+++ b/resources/views/master.blade.php
@@ -208,7 +208,7 @@ $menuExpanded = isset($_COOKIE['expandedMenu']) && $_COOKIE['expandedMenu'] == 1
 <footer class="app-footer">
     <div class="site-footer-right">
         Made with <i class="voyager-heart"></i> by <a href="http://thecontrolgroup.com" target="_blank">The Control
-            Group</a>
+            Group</a> - {{ Voyager::getVersion() }}
     </div>
 </footer>
 <!-- Javascript Libs -->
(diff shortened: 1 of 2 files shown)


putera:
This is really good @tnylea . Hopefully will merge to the system. Great @marktopper 👍

marktopper:
My only concern is that it have to be updated manually.
I was hoping for someway of automatication of this.

Im still open for other options, however if nothing seems to work, this might be merged.

tnylea:
@marktopper perhaps this might work inside of the `getVersion()` function:

```
public function getVersion(){
    $composer_lock = __DIR__ . '/../../../../composer.lock';
    $version = null;

    if(File::exists( $composer_lock )){

        // Get the composer.lock file
        $file = json_decode( File::get( $composer_lock ) );

        // Loop through all the packages and get the version of voyager
        foreach($file->packages as $package){
            if($package->name == "tcg/voyager"){
                $version = $package->version;
                break;
            }
        }

    }

    return $version;
}
```

Make sure to add:

```
use Illuminate\Support\Facades\File;
```

above the class. We'll want to do a check if the `getVersion()` returns a null then we just don't show anything. Otherwise this function will retrieve the composer.lock file and find the "tcg/voyager" package and then display the version 👍

marktopper:
Thanks @tnylea, will look into this and make a new commit tomorrow.

Issue #289, opened 2016-12-04
env('DB_CONNECTION')  is null when config have been cached


tanteng:
- Laravel Version: #.#.#
- Voyager Version: #.#.#
- PHP Version:
- Database Driver & Version:

### Description:
```
php artisan cache:clear
php artisan route:cache
php artisan config:cache
```
If run the command to cache config, you can see the exception 404 not found, then run php artisan route:list,you can't find the route for posts,users,categories and so on.

Exception:

```
NotFoundHttpException in compiled.php line 8912


    in compiled.php line 8912
    at RouteCollection->match(object(Request)) in compiled.php line 8236
    at Router->findRoute(object(Request)) in compiled.php line 8177
    at Router->dispatchToRoute(object(Request)) in compiled.php line 8173
    at Router->dispatch(object(Request)) in compiled.php line 2471
    at Closure$Illuminate\Foundation\Http\Kernel::dispatchToRouter(object(Request)) in Pipeline.php line 53
    at Closure$Illuminate\Routing\Pipeline::getInitialSlice(object(Request)) in compiled.php line 3226
    at CheckForMaintenanceMode->handle(object(Request), object(Pipeline::getInitialSlice;1945972947)) in compiled.php line 9900
    at Closure$Illuminate\Pipeline\Pipeline::getSlice(object(Request)) in Pipeline.php line 33
    at Closure$Illuminate\Routing\Pipeline::getSlice(object(Request)) in compiled.php line 9885
    at Pipeline->then(object(Kernel::dispatchToRouter;1270569954)) in compiled.php line 2415
    at Kernel->sendRequestThroughRouter(object(Request)) in compiled.php line 2399
    at Kernel->handle(object(Request)) in index.php line 54
```
tcg/voyager/routes/web.php line 20:
```
if (env('DB_CONNECTION') !== null && Schema::hasTable('data_types')):
```
env('DB_CONNECTION')  is null when config have been cached.

marktopper:
This is one of my biggest issues with the current code.
However I have no idea how to do this properly then.

I'm open for suggestions.

tanteng:
Illuminate/Foundation/Bootstrap/DetectEnvironment.php line 18：
```
public function bootstrap(Application $app)
{
    if (! $app->configurationIsCached()) {
        $this->checkForSpecificEnvironmentFile($app);

        try {
            (new Dotenv($app->environmentPath(), $app->environmentFile()))->load();
        } catch (InvalidPathException $e) {
[...]

Issue #284, opened 2016-12-03
Bug: Routes will invalid when config have been cached.


tanteng:
- Laravel Version: 5.3.*
- Voyager Version: 0.9.34
- PHP Version: PHP 7
- Database Driver & Version:MYSQL

### Description:

Here is not perfect!
Look this the code: 
```
env('DB_CONNECTION')
```

Do you known the value will be null when the config have been cached? I always use php artisan config:cache to cache config, so the route is invalid.
```
    if (env('DB_CONNECTION') !== null && Schema::hasTable('data_types')):
        foreach (TCG\Voyager\Models\DataType::all() as $dataTypes):
            Route::resource($dataTypes->slug, 'VoyagerBreadController');
    endforeach;
    endif;
```

file path:vendor/tcg/voyager/routes/web.php

marktopper:
#289

PR #295, opened 2016-12-04
Fixed issue related to env('DB_CONNECTION') are null when routes & configs are cached.

fergthh:
As mentioned in the documentation, on [Upgrade Guide 5.2 from 5.1](https://laravel.com/docs/5.3/upgrade#upgrade-5.2.0):

```
Caching And Env

If you are using the config:cache command during deployment, you must make sure that you are only calling the env function from within your configuration files, and not from anywhere else in your application.

If you are calling env from within your application, it is strongly recommended you add proper configuration values to your configuration files and call env from that location instead, allowing you to convert your env calls to config calls.
```
So, my proposal only change the if statement call to a try-catch block.

Ref #289 & #284


marktopper:
Thanks 👍

PR #296, opened 2016-12-04
Feature/overwrite models

marktopper:
This allows overwriting the models for Voyager.
I will leave this open for some time so we can discuss whether or not this is a good way to do it.

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -68,4 +68,27 @@ return [
         'subfolder' => 'public/', // include trailing slash, like 'my_folder/'
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Models config
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify models related to your Voyager system
+    |
+    */
+
+    'models' => [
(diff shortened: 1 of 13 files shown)


handiwijoyo:
@mark can you give some use cases for override the voyager models?

marktopper:
Of cause!

Someone created a issue once where they wanted to use soft delete for data, this can now easily be implemented by extending the models and set the `SoftDelete` trait on the model.

However I'm not sure wether or not to include this feature as it is. So still open for a discussion on wether or not this should be merged.

milkmeowo:
Awesome!
This feature awesome for someone already used `Zizaco/entrust`,`Custom Trait` or anything else
And this feature also can make voyager more flexible

fletch3555:
I'm not sure I agree with this.  What's stopping developers from just subclassing the model in the normal OOP fashion, adding what they need, then just changing the class/package path in the Voyager Database tool?  This would be my preference on how to do it.  I'd prefer to give developers the idea that they can use this without writing any code (too CMS-like, which we're trying to avoid).

marktopper:
Thanks for the discussion.
I have decided to close this.

PR #299, opened 2016-12-04
Update controllers permissions, hide non-bread-menu

handiwijoyo:


marktopper:
Just did a commit, let me know what you think.

handiwijoyo:
This is looks good to me :D Thanks @marktopper. I will let you know if there are any issue. You can merge it for now.

PR #302, opened 2016-12-05
Settings now start blank so need to check for no order value.

craigharman:
Not sure if settings are now blank on purpose, but as they are this causes an issue getting the settings order. Have put a patch in place to check for null value .


handiwijoyo:
@craigharman for `v0.10` when installing voyager need to add `--with-dummy` to seed the data.

PR #309, opened 2016-12-06
Refactor routes again

fletch3555:
- [X] Refactor routes so developers can include them with `Voyager::routes();` (like Auth does)
- [X] Change route paths to more logical pattern (still needs a little tweaking, but the heavy lifting is done)
- [X] Update all hidden _method fields to use method_field instead
- [X] Update all csrf_token fields to use csrf_field instead
- [ ] ...
- [ ] Profit...?!?


marktopper:
Tests have now been fixed.

PR #318, opened 2016-12-06
Delete-menu-patch

ezequielpintos:
Fix for issue in delete url for menu item.
   Url built with javascript is for instance /admin/menu/delete_menu_item//3
This is because id in route voyager.menu.delete_menu_item is menu/delete_menu_item/{id} and an extra "/" was added with JS.


fletch3555:
This is already fixed in v0.10, which will be released in the next couple days.

Issue #114, opened 2016-11-07
Disable Add Timestamp Field Button Event once it is click.


GNSubrahmanyam:
Hi, 

@tnylea  & @handiwijoyo Disable Add Timestamp Field Button Event once it is click. If the .delete-row button is clicked enable the  Add Timestamp Field Button Event.

Why because it is creating multiple instance of created_at and updated_at in users view. 

I am also working to solve this issue.

1. Normal Screen Shot

![image](https://cloud.githubusercontent.com/assets/18217008/20046035/f3554a3c-a4cb-11e6-93ec-477d25d7bc88.png)

2. Timestamp field's multiple instance screen shot.

![image](https://cloud.githubusercontent.com/assets/18217008/20046100/6dd2875c-a4cc-11e6-9fed-ec6c0e6e3d60.png)

My work to disable this

 ```
$('#newFieldTimestamps').click(function(){
           newRow('timestamps');
            $('#newFieldTimestamps').attr('disabled').unbind("click", timestampsHandler);/*Disable newFieldTimestamps once it is clicked.*/
        });

var timestampsHandler = function() {
   newRow('timestamps');
};
``` 
```
$('#tablebody').on('click', '.delete-row', function(){
          var clickedRow = $(this).parents('.newTableRow');
          if(clickedRow.hasClass('existing_row')){
            $(this).parents('.newTableRow').hide();
            $(this).parents('.newTableRow').find('.deleteField').val(1);
          } else {
            $(this).parents('.newTableRow').remove();
            /*If .delete-row Attr id="showNewFieldTimestamps" then delete timestamp row and enable #newFieldTimestamps click event*/
           if($(this).attr('id')){
             $(this).parents('.newTableRow').remove();
              $('#newFieldTimestamps').removeAttr("disabled").bind( "click", timestampsHandler); //Enable newFieldTimestamps once delete-row button is clicked
           }
           /*if ends*/ 
          }
        });
```
```
else if(kind == 'timestamps'){
        $('#' + unique_id).find('.fieldName').val('created_at & updated_at');
        $('#' + unique_id).find('.fieldName').attr('readonly', 'readonly');
        $('#' + unique_id).find('.fieldDefault').val('CURRENT_TIMESTAMP').attr('readonly', 'readonly');
        $('#' + unique_id).find('.fieldType').val('timestamp').attr('readonly', 'readonly').attr('disabled', 'disabled');
        $('#' + unique_id).find('.fieldNull').parent().hide();
[...]

PR #319, opened 2016-12-06
Prevent multiple add timestamp fields fixes Issue #114

allenjd3:
This should fix the problem of issue #114 

It disables the button when it is clicked once then re-enables it when the timestamp field is deleted.


marktopper:
This should properly also make it into the `v0.10` branch.

PR #320, opened 2016-12-07
Changing how Voyager views get overridden.

fletch3555:
Updating the view loading to use the built-in laravel view-overriding functionality.  Views formatted with the package name (e.g. `voyager::posts.index`) will automatically check the `/resources/views/vendor/voyager/` directory first.


marktopper:
Can you document this?

PR #323, opened 2016-12-07
[Patch] Remove unencessary role controller method

handiwijoyo:


marktopper:
It's on my todo to test this out.

PR #345, opened 2016-12-08
When user is saved in admin if the image is not set, set it to default

tnylea:
When a new user is created in the admin and they do not specify an avatar it shows as a broken image when they login. 

With this pull request now anytime the avatar is not set it will set it to the default 'users/default.png' image.


marktopper:
You merged this 2 seconds before I was gonna do it 😂

PR #346, opened 2016-12-08
[Patch] Some pages fields need to be nullable to avoid error

handiwijoyo:


marktopper:
Also just added this to the documentation for migrating.

PR #357, opened 2016-12-09
moving out the extra view check in the bread controller

tnylea:
I didn't know that Laravel automatically looked inside of the /resources/views/vendor/{package} folder to override package views.

So, I'm reverting back my old change. My bad. Moving forward, I'm going to have someone else Merge my PR's. 👍


fletch3555:
It's all good!

PR #369, opened 2016-12-11
Save default User Role onCreated

akazorg:
Default user role was not saved.


fletch3555:
Good catch, thanks!

marktopper:
Reverted this since it should be optional if the developer want to set the save the user model instantly when setting a role.

Instead this should be done `$user->setRole('admin')->save();`.
Made this possible in my latest PR.
Also made this for the event that sets the default role when creating users.

Revert PR: #374
Fix PR: #375

However, thanks a lot for pointing this out 👍

PR #370, opened 2016-12-11
Fixed Menu display apply sortBy order + check if menu exists

akazorg:
Using the menu at the Front-end, sortBy wasn't being applied.

Also added a check if menu exists.


fletch3555:
Thanks for this!

PR #374, opened 2016-12-12
Revert "Save default User Role onCreated"

marktopper:
Reverts the-control-group/voyager#369


fletch3555:
Just for reference, there is a "revert" button on a PR after it has been merged.  We can make use of that instead of these "extra" PRs

marktopper:
@fletch3555 This is the output of that button 😂

Issue #187, opened 2016-11-22
Duplicate entry 'test' for key 'menus_name_unique'


marktopper:
If you go ahead and make a menu called `test`, and then create it once again.

Then it comes up with this exception page, however shouldn't it be a friendly error message instead saying that name is taken?

### Exception
```
QueryException in Connection.php line 769:
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'test' for key 'menus_name_unique' (SQL: insert into `menus` (`name`, `updated_at`, `created_at`) values (test, 2016-11-22 08:05:04, 2016-11-22 08:05:04))
```
Made from `VoyagerBreadController.php line 254`

handiwijoyo:
Fixed in #518

PR #383, opened 2016-12-13
Error message for duplicate menu creation. Fix for issue #187

allenjd3:
This is a possible solution to #187


handiwijoyo:
I think this should handled by validation unique instead of checking it manually. Also it will easier to add another validation later.

https://laravel.com/docs/5.3/validation#rule-unique

marktopper:
I must agree with @handiwijoyo on this one.

PR #390, opened 2016-12-14
Importing class DB

putera:
Importing the class DB


src/Http/Controllers/VoyagerBreadController.php
```diff
@@ -5,6 +5,7 @@
 use Illuminate\Http\Request;
 use TCG\Voyager\Models\DataType;
 use TCG\Voyager\Voyager;
+use DB;
```
marktopper: It would be better to use the actual `DB` object instead of the alias, since if the user removes the alias, then this would break.

marktopper:
Please remove the `composer.lock` file from your commit, it have been removed from the package.

src/Http/Controllers/VoyagerBreadController.php
marktopper: If I remember right it would be `Illuminate\Support\Facades\DB`.

marktopper:
Thanks 🎉

PR #400, opened 2016-12-15
Fix bug  <field>=><display_name>

xule:
Fix bug

### DIFF
--- a/publishable/views/posts/browse.blade.php
+++ b/publishable/views/posts/browse.blade.php
@@ -23,7 +23,7 @@
                             <thead>
                                 <tr>
                                     @foreach($dataType->browseRows as $rows)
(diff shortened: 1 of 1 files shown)


publishable/views/posts/browse.blade.php
```diff
@@ -22,8 +22,8 @@
                         <table id="dataTable" class="table table-hover">
                             <thead>
                                 <tr>
-                                    @foreach($dataType->browseRows as $rows)
-                                    <th>{{ $rows->field }}</th>
+                                    @foreach($dataType->browseRows as $row)
+                                    <th>{{ $row->dispaly_name }}</th>
```
fletch3555, line 26: I'm guessing this is supposed to be "display_name"

Issue #204, opened 2016-11-26
[Feature] Generate migration file when database table created


handiwijoyo:
We need to generate migration file when database table created / edited via voyager. 

It will be easier to pass the project to another developer and to run testing during development, since all database created will be wiped out when we are refreshing the database or run migrate:refresh command.

I will not using table creation feature at moment, and will manually create migration & model file then add bread to it.

marktopper:
This would also be a good idea to have when modifying existing database tables.

akazorg:
We should work on that indeed!!

brandonferens:
+1

brandonferens:
I'm actually not sure how I feel about this. There are so many potential caveats and directions this could go. One of the benefits of having this dynamic system is that you don't really need migrations. HOWEVER, it is a pain to move from a dev server to a live server. Perhaps we utilize a package like [Migrations Generator](https://github.com/Xethron/migrations-generator) to create the necessary migrations only when the user wants them created. They could check a box that creates the migration, or the could choose from a list of db tables to have migrations created from.

craigharman:
What about also having a global setting "create migrations" that always created migrations when a new table was created or existing table updated?

handiwijoyo:
I believe using https://github.com/Xethron/migrations-generator will the best solution for now.

fletch3555:
Using the Xethron migrations generator is a good workaround, but I agree that this would be a reasonable addition to Voyager.  I'm open to further suggestions on how we can implement it (ideally without requiring an Artisan command to do so).

fletch3555:
I'm going to close this for now since it's so old.  We may revisit this after the 1.0 release.  In the meantime, if anyone wants to tackle it, we're certainly open to pull requests

belthaZornv:
Bump - this is the only feature that is missing to support the db functionality; it's a tiny bit useless to be able to create db tables without having the migrations generated alongside it. - Is it possible to re-open this; and using xethron/or similar could be an hours work or so.

What do you guys think?

KABBOUCHI:
@belthaZornv db migration is done, you can follow the progress here #2604

belthaZornv:
Thanks buddy, wasn't aware. :)

psierak:
[...]

PR #408, opened 2016-12-16
Adding functionality to generate migration when creating DB table

fletch3555:
Partial implementation of #204.  This creates the migration, but does not populate it with fields.

### DIFF
--- a/resources/views/tools/database/edit-add.blade.php
+++ b/resources/views/tools/database/edit-add.blade.php
@@ -117,9 +117,18 @@
                                             </div>
 
                                             @if(!isset($table))
-                                                <div class="col-md-6">
+                                                <div class="col-md-3 col-sm-4 col-xs-6">
                                                     <label for="create_model">Create model for this table?</label><br>
                                                     <input type="checkbox" name="create_model" data-toggle="toggle"
+                                                           data-on="Yes, Please" data-off="No Thanks">
+
+                                                </div>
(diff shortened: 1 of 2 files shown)


marktopper:
It's a good start 👍

fletch3555:
I'm thinking this should be merged as-is, and we can re-implement once we figure a way to build out the fields as well.  It's currently using built-in Laravel functionality to generate the migration.

What are your thoughts @marktopper?

marktopper:
I'm in on that @fletch3555.

We can always add more features for this later on.

tokkerbaz:
Help me with my question =) I'm developing with voyager and want to tell its awesome admin package! Ive never seen cooler adminpack than this =) My question is : 
 I am in development not in production, today i want to publish to internet, and if there is any migration how can i insert to production my custom tables which ive created by voyager? 
Is there any chance that i wont lose tables in production?
Or I should export my local database to hosting? 

Thank you for your work! =)) 
You are making development easier =))

marktopper:
@tokkerbaz, you should copy your database to production.
We will however look forward for having better support for this.

handiwijoyo:
@tokkerbaz Voyager do not have this feature yet. You may want to use this package for now https://github.com/Xethron/migrations-generator

vanloc0301:
Hi @fletch3555 Is that feature continues to develop?

PR #442, opened 2016-12-25
Update VoyagerBreadController.php

ardani:
make read template can be overridden


fletch3555:
This actually isn't needed.  The view is already able to be overridden by creating a view in `resources/views/vendor/Voyager/bread/read.php`

PR #443, opened 2016-12-26
Adding fix for the menu builder and the bootstrap menu

tnylea:
Currently when editing menus from the Menu builder it will occasionally be blank in the modal and you will have to close it and open it again. This is because if you click on the text inside the 'Edit' button the `e.target` does not register and grab the `data` attributes for that menu item. Instead `e.currentTarget` needs to be used so the action of the button stays consistent.

I've also added a fix for the current bootstrap menu. Adding a dropdown and caret icon if it is a dropdown 👍

### DIFF
--- a/resources/views/menus/builder.blade.php
+++ b/resources/views/menus/builder.blade.php
@@ -156,22 +156,23 @@
         $(document).ready(function () {
             $('.dd').nestable({/* config options */});
             $('.item_actions').on('click', '.delete', function (e) {
-                id = $(e.target).data('id');
+                id = $(e.currentTarget).data('id');
                 $('#delete_form')[0].action = $('#delete_form')[0].action.replace("__id",id);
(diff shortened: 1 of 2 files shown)


tnylea:
@marktopper @fletch3555 or @handiwijoyo Whenever you guys get a chance to merge this that would be great. I don't think we are merging our own PR's correct?

I wanted to add this fix because when I go to create a Bootstrap menu and implement it in a default laravel install, the dropdown didn't work, so I added a quick fix for that so that way I could create the Voyager Menu Screencast and show it working with the default bootstrap template.

Also, I added a fix to the builder when you go to edit a menu item :)

Thanks a ton.

fletch3555:
Correct, we shouldn't be merging our own pull requests.  I'll review in a bit after I get home, unless someone else beats me to it.

Issue #444, opened 2016-12-26
require pagination for browse data


ardani:
```
$dataTypeContent = (strlen($dataType->model_name) != 0)
            ? call_user_func([$dataType->model_name, 'all'])
            : DB::table($dataType->name)->get(); // If Model doest exist, get data from table name
```
i think it will be slow down if data more than 15K,
so better put paginate for it
thanks

tnylea:
@ardani Thanks for opening the issue. This has been resolved in the latest version.

Issue #422, opened 2016-12-20
Problem with BREAD for table with 14k+ rows.


serzhiio:
Looks like, BREAD is not adopted for big tables (14k+ rows).
Voyager loads all rows on one page, it's a disaster.

marktopper:
Your welcome to make a pull request fixing this.

tnylea:
This has been updated in the latest `0.10.6` version. Users can now check if they want to do server side pagination in the BREAD editor 👍

Issue #404, opened 2016-12-15
Datatable server side in Browse


Muhammadinaam:
In browse view, please add functionality of server side loading of datatable data to improve performance on large tables.

marktopper:
The performance should be okay for most tables.

How big are your tables?

fletch3555:
I'm not sure what you mean by "server side loading".  If you're referring to pagination of the data, there is already an open issue for this (#81)

marktopper:
Closed due to inactivity.

Issue #156, opened 2016-11-12
Table view loading 1 200 000 cities from my database


alexdanielyan:
Table view loading 1 200 000 cities from my database. Thanks.

ChadTaljaardt:
Whats the issue?

maiorano84:
@team5ru What are you even asking?

alexdanielyan:
I thought it was obvious, the browser dies.

alexdanielyan:
Here is example of using datatables in right way https://datatables.net/extensions/scroller/examples/initialisation/server-side_processing.html

alexdanielyan:
There was solution, but it's a little bit complicated #81

marktopper:
Im open for pull requests on this one.

tnylea:
This has been fixed in the latest version `0.10.6` see the following PR that has resolved this: https://github.com/the-control-group/voyager/pull/446

Thanks 👍

PR #446, opened 2016-12-26
Optionally paginate data type on server-side, avoiding dataTable 

oschettler:
As fix for

* [#444 require pagination for browse data](https://github.com/the-control-group/voyager/issues/444)
* [#422 Problem with BREAD for table with 14k+ rows](https://github.com/the-control-group/voyager/issues/422)
* [#404 Datatable server side in Browse](https://github.com/the-control-group/voyager/issues/404)
* [#156 Table view loading 1 200 000 cities from my database](https://github.com/the-control-group/voyager/issues/156)
* [#81 Add server side pagination to browse](https://github.com/the-control-group/voyager/issues/81)

... I have added a new option _server_side_ to paginate the _browse_ view, using the standard Laravel _paginate()_ method instead of the original approach based on _dataTable_. The default is still to use _dataTable_, but this can now be changed by editing a BREAD and setting the switch:

new-switch

The result is that even big tables now no longer time out in the browser:

pagination

### DIFF
--- a/publishable/database/migrations/2016_12_26_201236_data_types__add__server_side.php
+++ b/publishable/database/migrations/2016_12_26_201236_data_types__add__server_side.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class DataTypesAddServerSide extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('data_types', function (Blueprint $table) {
+            $table->tinyInteger('server_side')->default(0)->after('generate_permissions');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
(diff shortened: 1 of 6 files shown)


handiwijoyo:
I think we should use dataTables server side pagination instead of Laravel pagination.

karandatwani92:
Yes, it's very important. Thanks @oschettler 
But Instead of server side. we should ajaxify it using datatables

karandatwani92:
Instead of server side. we should ajaxify it using datatables

tnylea:
@oschettler Very cool! Appreciate the commit. I'm testing it now and I'll be merging it in very shortly.
@handiwijoyo I agree, but for now this will serve as a good solution until someone wants to build the server side pagination with dataTables :)

Muhammadinaam:
This commit is not working correctly.

I added 10,000 users with tinker and then I edited the BREAD and enabled server side option. Then I noticed two problems:

1. Now voyager shows on 15 entries. Next and Previous buttons dont show records after 15 users. I have to manually add (?page=4) at the end of url
bug1

2. When I try to edit BREAD and deactivate server side option and save. BREAD gets saved but server side option is not disabled.

Further, I am not able to show ID of users in browse page. (May be this problem is not related to above commit)

PR #451, opened 2016-12-28
Adding fixes to the server side pagination functionality and styles

tnylea:
This is a quick fix for serverside pagination and the styles.


fletch3555:
@tnylea, just an FYI, StyleCI should create a commit with code style fixes after a PR gets merged.  @mark has done quite a few of them in the past

PR #452, opened 2016-12-28
remove duplicate code

wangxiaomo:
first of all, i love voyager so much.

i'm doing some i18n job and find some duplicate code, so i just removed it.


handiwijoyo:
Thanks man! @wangxiaomo

PR #453, opened 2016-12-28
Error while user was logged out.

karandatwani92:
Instead of throwing exception, hasPermission() was being called on null user while we override route and controller for custom functioning.


marktopper:
Thanks 👍

PR #490, opened 2017-01-05
add autogenerate slug function

unabomber87:
add URL slug generate function when user enter a title for a post

### DIFF
--- a/publishable/views/posts/edit-add.blade.php
+++ b/publishable/views/posts/edit-add.blade.php
@@ -209,4 +209,13 @@
 @section('javascript')
     <script src="{{ config('voyager.assets_path') }}/lib/js/tinymce/tinymce.min.js"></script>
     <script src="{{ config('voyager.assets_path') }}/js/voyager_tinymce.js"></script>
+    <script src="{{ config('voyager.assets_path') }}/js/speakingurl.js"></script>
+    <script type="text/javascript">
+        $(document).ready(function() {
+            $('[name="title"]').on('keyup', function() {
+                //console.log( $(this).val() );
+                $('[name="slug"]').val(getSlug($(this).val())) ;
+            });
+        }); 
+    </script>
 @stop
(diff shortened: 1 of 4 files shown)


publishable/views/posts/edit-add.blade.php
```diff
@@ -209,4 +209,13 @@
 @section('javascript')
     <script src="{{ config('voyager.assets_path') }}/lib/js/tinymce/tinymce.min.js"></script>
     <script src="{{ config('voyager.assets_path') }}/js/voyager_tinymce.js"></script>
+    <script src="{{ config('voyager.assets_path') }}/js/speakingurl.js"></script>
+    <script type="text/javascript">
+        $(document).ready(function() {
+            $('[name="title"]').on('keyup', function() {
+                //console.log( $(this).val() );
```
marktopper: Please remove `//console.log( $(this).val() );`

marktopper:
Shouldn't this be for other BREAD 🍞 as well? Now it will only be for posts, but how about `categories` and when I use Voyager to create a `products` page?

unabomber87:
i change it to be a global function so
when there's an input with name 'title' and another one with name 'slug' => generate slug in input slug
or an input with name 'name' and another one with name 'slug' => generate slug in input slug

akazorg (commented):
Will it update the slug input even if it isn't empty?
It would be more flexible if you can define on Bread Option, what field is causing the auto generation.

marktopper:
@akazorg: You are right, this should become a bread option.

akazorg:
@marktopper I wrote a slugify plugin in JS for working with the BREAD, it works great.

Basically you can define:
* what field is causing the update
* if you want to force slug generation, for non empty slug fields.

By default, if slug is blank it will auto generate a new one.

I can share it if you want.

marktopper:
Please share @akazorg, might be a better approach.

akazorg:
@marktopper I've found a little nasty bug, will fix it and share with you!

marktopper:
@akazorg: On Slack or where?

akazorg:
I'll let you know @marktopper, sorry, I'm just so busy atm.
I hope to find some time today.

marktopper:
It's okay @akazorg

PR #493, opened 2017-01-06
Make nullable fields insertable

antranapp:
Since the parent_id of Category Model is nullable, we need a way to insert a null value into a field.

screen shot 2017-01-06 at 13 26 33

### DIFF
--- a/publishable/database/seeds/DataRowsTableSeeder.php
+++ b/publishable/database/seeds/DataRowsTableSeeder.php
@@ -738,7 +738,10 @@ class DataRowsTableSeeder extends Seeder
                 'edit'         => 1,
                 'add'          => 1,
                 'delete'       => 1,
-                'details'      => '',
+                'details'      => '{
+"default": 1
+}
+',
             ])->save();
         }
         $dataRow = DataRow::firstOrNew([
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -57,9 +57,10 @@
                                     <label for="name">{{ $row->display_name }}</label>
 
                                     @if($row->type == "text")
+                                        <?php $options = json_decode($row->details); ?>
                                         <input type="text" class="form-control" name="{{ $row->field }}"
                                                placeholder="{{ $row->display_name }}"
-                                               value="@if(isset($dataTypeContent->{$row->field})){{ old($row->field, $dataTypeContent->{$row->field}) }}@else{{old($row->field)}}@endif">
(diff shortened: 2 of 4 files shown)


ghost:
Why not just make it null if it's blank? :)

antranapp:
See here: http://dev.mysql.com/doc/refman/5.7/en/working-with-null.html
'' IS NULL returns 0, '' IS NOT NULL returns 1, so "empty" and "null" are totally different in the view of mysql.
But anyway, the internet has seemed to discuss a lot about this theme: https://www.google.de/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=mysql+null+vs+empty 

Since voyager is a universal admin editor for all kind of projects, I prefer to give user a chance to enter "NULL" and "empty" separately. You will never know what kind of requirements could be.

I'm totally ok if anyone can implement another solution and make inserting nullable possible. I'm currently using this solution for my needs.

marktopper:
@peacemoon: Instead of hardcoding it, how would you feel about making it into an option, like this:

```
{
    "null": "__null__"
}
```

Meaning that `__null__` will become `null`.

In this case, a empty string can become null by setting:

```
{
    "null": ""
}
```

antranapp:
@marktopper this is a great idea. It will be then flexible enough for everyone. I'll adapt my pull request according to your suggestion.

marktopper:
Creating work, merging in now. 👍

hazem1taha:
@marktopper @peacemoon 
I get this error when using this json option for the parent_id field and making it nullable:

`{
    "relationship": {
        "key": "id",
        "label": "name"
    },
    "default": "null",
    "options": {
        "null": "Main category"
    }
}`

the error:

`QueryException in Connection.php line 770:
SQLSTATE[HY000]: General error: 1366 Incorrect integer value: 'null' for column 'parent_id' at row 1 (SQL: update `categories` set `parent_id` = null, `updated_at` = 2017-01-08 11:06:11 where `id` = 4)`

marktopper:
I believe it tries to insert `"null"` as a string and not a `null` value.

In order to make it parse a string into a `null` value, define the following:

```
{
    "null": "null"
}
```

Then the string `"null"` will become a `null` value.

hazem1taha:
Here is what I'm doing:
adsl

Here is the error I get:

`QueryException in Connection.php line 770:
SQLSTATE[HY000]: General error: 1366 Incorrect integer value: 'null' for column 'parent_id' at row 1 (SQL: update `categories` set `parent_id` = null, `updated_at` = 2017-01-08 20:50:51 where `id` = 3)`

Am I missing something ? anyway I'll wait for the new version you said it will be available today.

antranapp:
Use this:
{ "relationship": { "key": "id", "label": "name" }, "default": "null", "null": "null", "options": { "null": "Main category" } }

PR #495, opened 2017-01-06
Add new helper methods for permission checking

marktopper:
Add new helper methods on `Voyager` instance.

- [X] `Voyager::canOrFail('browse_pages')`
- [X] `Voyager::canOrAbort('browse_pages', 403, 'A friendly error message')`

Add new helper methods on `User` object (requires that trait `VoyagerUser` is used)

- [X] `$user->hasPermissionOrFail('browse_pages')`
- [X] `$user->hasPermissionOrAbort('browse_pages', 403, 'A friendly error message')`

In next breaking change release, the `Voyager::can('browse_pages')` will return a `boolean` instead of failing.

### DIFF
--- a/src/Traits/VoyagerUser.php
+++ b/src/Traits/VoyagerUser.php
@@ -3,6 +3,7 @@
 namespace TCG\Voyager\Traits;
 
 use TCG\Voyager\Models\Role;
+use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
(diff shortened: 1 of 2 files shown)


fletch3555:
Honestly, I think we can just hold off and introduce all of this, including your breaking change, in the next "breaking" version.  Doesn't make much sense to me to piecemeal it.

Issue #474, opened 2016-12-31
Selecting parent category when adding a new category instead of inserting the parent's ID


hazem1taha:
Does this feature exist >> Selecting parent category when adding a new category instead of inserting the parent's ID ?

If not, I suggest it as a new feature.

Thanks,

Issue #475, opened 2017-01-01
Error when submitting Category without parent_id


wtfz:
- Laravel Version: 5.3
- Voyager Version: 0.10.7

### Description:
Error shows 'Incorrect integer value' at parent_id when editing/adding category without parent_id.

### Steps To Reproduce:
1. Go to Categories menu
2. Click on Edit to any categories or Add New
3. Click Submit without parent_id

marktopper:
In the current `master` branch you should be able to add the following to the breads *Additional Options* to be able to leave the field empty.

```
{
    "null": ""
}
```

Meaning that a empty string will become a `null` value.

Also for better support of this follow #496

PR #496, opened 2017-01-06
Category parent id select dropdown

antranapp:
screen shot 2017-01-06 at 16 14 32

Require https://github.com/the-control-group/voyager/pull/493 to be merged first

Fix https://github.com/the-control-group/voyager/issues/474 and https://github.com/the-control-group/voyager/issues/475

### DIFF
--- a/publishable/database/migrations/2016_02_15_204651_create_categories_table.php
+++ b/publishable/database/migrations/2016_02_15_204651_create_categories_table.php
@@ -22,7 +22,6 @@ class CreateCategoriesTable extends Migration
             $table->string('slug')->unique();
             $table->timestamps();
         });
-
     }
 
     /**
--- a/publishable/database/migrations/2016_02_15_204651_create_categories_table.php
+++ b/publishable/database/migrations/2016_02_15_204651_create_categories_table.php
@@ -12,15 +12,17 @@ class CreateCategoriesTable extends Migration
      */
     public function up()
     {
-        // Create table for storing roles
+        // Create table for storing categories
         Schema::create('categories', function (Blueprint $table) {
             $table->increments('id');
-            $table->integer('parent_id')->nullable();
+            $table->integer('parent_id')->unsigned()->nullable()->default(null);
+            $table->foreign('parent_id')->references('id')->on('categories')->onUpdate('cascade')->onDelete('set null');
             $table->integer('order')->default(1);
             $table->string('name');
             $table->string('slug')->unique();
             $table->timestamps();
         });
+
     }
 
     /**
--- a/publishable/database/seeds/DataRowsTableSeeder.php
+++ b/publishable/database/seeds/DataRowsTableSeeder.php
@@ -723,6 +723,7 @@ class DataRowsTableSeeder extends Seeder
                 'delete'       => 1,
                 'details'      => '{
 "default": "__null__",
+"null": "__null__",
 "options": {
 "__null__": "NULL"
 },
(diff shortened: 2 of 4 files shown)


fletch3555:
You committed `CategoryTest.php` in both #493 and here.  I recommend removing it from both since it doesn't seem directly related to either

marktopper:
@peacemoon: Can you make this work with the latest changes of #493

antranapp:
I'm working on it now.

marktopper:
I think for a foreign key like this, a empty input should become `null`, since the database expect either a foreign key or `null`.

Which means the the category table had in database relations, then a empty value would become `null`.
But in the current case it would become empty and the database could complain that a category with the id of empty string does not exists.

Thoughts?

antranapp:
@marktopper  I don't think I get your idea. Could you elaborate it a little bit?

antranapp:
I think I get your idea now. It's a good idea to have a foreign key constraint for `parent_id`. I also add the default value for `parent_id`to `null`. Hope it's ok now.

marktopper:
Just gonna test this out before I merge.

marktopper:
Works like a charm! Great work 👍

hazem1taha:
@marktopper 

Is this PR merged to the master ? as I ran composer update` but the Category Model still doesn't has the parent_id relationship method.

    <?php

    namespace TCG\Voyager\Models;

    class Category extends Model
    {
    protected $table = 'categories';

    protected $fillable = ['slug', 'name'];

    public function posts()
    {
        return $this->hasMany(Post::class)
            ->published()
            ->orderBy('created_at', 'DESC');
    }
    }

marktopper:
It's merged into `master`.

https://github.com/the-control-group/voyager/blob/master/src/Models/Category.php#L20

hazem1taha:
But why I didn't get these changes after running `composer update` ? Am I missing something ? Or do these changes need to be added to a new release to be able to get it locally using the composer command ?

marktopper:
What does you `composer.json` say? And what version does your Voyager admin panel say on the bottom right corner?

hazem1taha:
`v0.10.9`

    {
        "name": "laravel/laravel",
        "description": "The Laravel Framework.",
        "keywords": ["framework", "laravel"],
        "license": "MIT",
        "type": "project",
        "require": {
            "php": ">=5.6.4",
            "laravel/framework": "5.3.*",
            "tcg/voyager": "0.10.*"
        },
        "require-dev": {
            "fzaninotto/faker": "~1.4",
            "mockery/mockery": "0.9.*",
            "phpunit/phpunit": "~5.0",
            "symfony/css-selector": "3.1.*",
            "symfony/dom-crawler": "3.1.*"
        },
        "autoload": {
            "classmap": [
                "database"
            ],
            "psr-4": {
                "App\\": "app/"
            }
        },
        "autoload-dev": {
            "classmap": [
                "tests/TestCase.php"
            ]
        },
        "scripts": {
            "post-root-package-install": [
                "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
            ],
            "post-create-project-cmd": [
                "php artisan key:generate"
            ],
            "post-install-cmd": [
                "Illuminate\\Foundation\\ComposerScripts::postInstall",
                "php artisan optimize"
            ],
            "post-update-cmd": [
                "Illuminate\\Foundation\\ComposerScripts::postUpdate",
                "php artisan optimize"
            ]
        },
        "config": {
            "preferred-install": "dist"
        }
    }

marktopper:
Just checked, it's not in a release yet.
You will have to wait for the next release or update to `dev-master`.

PR #497, opened 2017-01-06
Migrations independent of users table

jmarcher:
In my case my not only the User model is different, also the database name is different.

That's why I made this changes in order use the table name based on the model.

That means that you if you publish you vendors and then edit the namespace the migrations will take this user to resolve the user's table name

### DIFF
--- a/publishable/database/migrations/2016_01_01_000000_add_user_avatar.php
+++ b/publishable/database/migrations/2016_01_01_000000_add_user_avatar.php
@@ -11,7 +11,7 @@ class AddUserAvatar extends Migration
      */
     public function up()
     {
-        Schema::table('users', function ($table) {
+        Schema::table({*users_table*}, function ($table) {
             $table->string('avatar')->default('users/default.png');
         });
     }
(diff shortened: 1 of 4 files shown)


fletch3555:
I'm not sure how I feel about this.  To me, the migrations get published, so you (as the developer) are free to update them as you see fit.

jmarcher:
The problem is that if you only run voyager:install and you don't have a users table this command fails.

abdgad:
There are some problems with the current installation process. I will make a PR later and we'll discuss it further = )

abdgad:
@marktopper @jmarcher @fletch3555  what about including a create_users_table migration with Voyager, and add a config setting to specify users table? And all user migrations will use the table from the config..?

marktopper:
@abdgad: How would that work for existing Laravel applications that already have a `users` table?

marktopper:
This does not seems ideal.

Maybe we could do something like [this](https://github.com/jrean/laravel-user-verification/blob/master/src/resources/migrations/2016_09_13_070520_add_verification_to_user_table.php).

But use the `voyager.user.namespace` as model if defined. That will allow users to change that in order to change the migrations?

Open for other ideas, however this does not seem ideal.

PR #502, opened 2017-01-07
Documentate the usage of null values

marktopper:


marktopper:
Ignore TravisCI
Just disabled it from running on those documentation pull requests.

Issue #492, opened 2017-01-06
Need to offer more installation options if there is already an existing installation of Voyager


abdgad:
- Laravel Version: 5.3.*
- Voyager Version: 0.10.7
- PHP Version: 7.0
- Database Driver & Version: mysql (mariadb)

### Description:
Currently, if you re-install Voyager using artisan voyager:install, it will not check if there is an existing installation already in place.
The issue is that the installation process will execute again re-doing everything. One of the things that keep happening for example is appending Voyager routes to **/routes/web.php** everytime you issue in the installation command.

### Suggestions:
Check if there is an existing installation. If yes, warn the user and offer more options.
This will be an improvement to the installer. Maybe add options to Cancel, Re-Install, Override, Upgrade, etc... What do you guys think?

Also add an option to Clean and Uninstall Voyager.

A simple method to check if there is an existing installation of Voyager:
```php
/**
 * Check if there is an existing installation of Voyager.
 *
 * @return bool
 */
protected function checkExistingInstallation() {
    // if there is a copy of the voyager.php config, it means Voyager is already installed, right?
    return file_exists(config_path('voyager.php'));
}
```

### Steps To Reproduce:
exec **php artisan voyager:install** multiple times in the same Laravel app.

marktopper:
I think this would be a quite good idea, however if something went wrong and the user might want to reinstall Voyager, a option `--force` could be an option.

abdgad:
I will try to make a PR later to improve the installation process.

antranapp:
It would be also great if we can make Pages & Post optional. Not all applications need such features.

handiwijoyo:
@peacemoon Pages & Post will be removed on coming release. And we can still install those using Voyager Hook https://github.com/larapack/voyager-hooks

handiwijoyo:
This issue should be fixed on `1.0`

PR #504, opened 2017-01-07
Improve the installation process

abdgad:
The goal of this PR is to improve the installation process and add more options etc... See #492

In order to organize Installation commands and make it easier to add new options in the future etc... I created a special namespace for Installation related commands.
Did some refactoring to **InstallCommand** to make it easier to add more installation options in the future.
Created **Settings.php** which should contain required settings for installation. Currently it also contains some helper methods. I didn't know where to put them, so I just added them there...

- [X] When installing, check if there is an existing installation. warn the user.
- [X] Added installation options for the user in case there is an existing installation.
- [X] Add command to uninstall Voyager.
- [x] Find a way to reset only Voyager migrations when uninstalling.
- [x] Copy Voyager publishable resources to a special folders to delete them easily when uninstalling.
- [ ] Decide whether to rollback Voyager migrations during uninstallation, or leave to it to the user to do it manually. Cause if the user has migrations that build on top of Voyager ones, it may cause issues if we remove them.
- [ ] Add command to backup existing installation.

Please give us your suggestions = )


marktopper:
Just noticed that you added `UserTableSeeder` even if `--with-dummy` is not defined.

The thing is we need a way to *not* create a user, since some people might want to add Voyager to their existing Laravel application.

abdgad:
Ok I just reverted that = )

marktopper:
@abdgad: However if you have another idea for a way to choose wether or not to create a user, I am open minded. :)

abdgad:
@marktopper To login to Admin Area, a user is required, right? So I think it makes sense to create a default admin account even if --with-dummy is not specified since it's a required thing. What do you think?

abdgad:
The main features of this PR are ready. Please test and give me suggestions to make more improvements = )

abdgad:
Just noticed that there is a --path option for migrate command that can be used to migrate and rollback migrations from a specific path. It can be used here. The only problem is that Voyager doesn't ship with a create_users_table migration, and migration using --path will fail because users table is missing.
I suggest shipping a create_users_table migration with voyager. see #497

marktopper:
@abdgad: I can see it could be a good idea to create a user even if `--with-dummy` is not set.
Maybe it could create the user if no users currently exists, if users do already exists, it means that it is an existing application.

Also publishing `create_users_table` would give errors for applications that already have a `users` table. I'm however open for ideas.

abdgad:
@marktopper Here are my suggestions.

1) Add a table_prefix: this is a prefix that Voyager tables will use. For example 'voyager_menus'. This will set apart Voyager specific tables from the rest.
2) When installing, prompt the user to enter his user Model. We will offer him a default one.
    Add the user Model to voyager.namespace config. This is where custom user model is stored, right?
    Please tell me which User model is used cause I am a bit confused here.
3) Include a create_users_table migration: It will use the table name from specified user model, and it only creates a users table if it doesn't exist.
4) If --with-dummy is not specified, migrate only core tables that are required for Voyager to function. Tables like categories and posts are not core tables.
5) Seeds: if --with-dummy is not specified, only seed core data. No demo stuff.
    Also seed a user account if one doesn't exist.
6) If --with-dummy is specified, migrate dummy tables and seed them.

I think these steps will deal with users table issues and also make it cleaner to install and uninstall voyager.
Please tell me what you think?

marktopper:
1. We could add a table prefix, but it should be optional, as not everyone would like this.

2. Asking for the user model could be a good option. And then fallback to `voyager.user.namespace` if not defined.

3. Yes, `voyager.user.namespace` is the ideal User model.

4. Indeed, it would be nice to leave those migrations out.

5. Seeding of user account if NONE exists could be fine with me.

6. 👍

marktopper:
About your todo:
> Decide whether to rollback Voyager migrations during uninstallation, or leave to it to the user to do it manually. Cause if the user has migrations that build on top of Voyager ones, it may cause issues if we remove them.

Maybe we should ask the user if they wish to rollback and let it be up to them on uninstall?

marktopper:
Would you be able to split up the features for the backup command?
> Add command to backup existing installation.

Since I would love to help building this, just wish to know what parts you would backup how. 😄

abdgad:
@marktopper I think we should continue working on installation after Voyager is stable. Then, we'll know which things to backup etc...

fletch3555:
I'm closing this for now.  We can revisit it after a v1.0 release

PR #518, opened 2017-01-10
Fix #478 Error updating unique key table & Add Enum support!

### LINKED ISSUE #478: Duplicate key name error when updating a table with unique columns
- Laravel Version: 5.3.28
- Voyager Version: 0.10.5
- PHP Version: 7
- Database Driver & Version: MySQL 5.7.11

### Description:
If I create a table with an `unique` column, I can't update it from voyager anymore because it doesn't check if the key is already there, and it turns into a 
```
QueryException in Connection.php line 770:
SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name 'services_name_unique' (SQL: alter table `services` add unique `services_name_unique`(`name`))
```

handiwijoyo:
This will fix issue when adding, removing, updating table Unique key.

### DIFF
--- a/src/Http/Controllers/Traits/DatabaseQueryBuilder.php
+++ b/src/Http/Controllers/Traits/DatabaseQueryBuilder.php
@@ -17,7 +17,6 @@ trait DatabaseQueryBuilder
     private $typeBlacklist = [
         'char',
         'double',
-        'enum',
         'ipAddress',
         'json',
         'jsonb',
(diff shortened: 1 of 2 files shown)


handiwijoyo:
We need better approach to `enum` type. If we drop the column, we will lost that column data.

marktopper:
Can you make a separate pull request with the fixes for unique keys?

PR #520, opened 2017-01-10
Repopulate enum data type when editing database

handiwijoyo:
This will repopulate enum data type as expected 

image


marktopper:
NOTE: Before we merge this we will ensure that we have got support for `enum`.

marktopper:
Related issue #547

marktopper:
I think we should check the database connection for the table before showing the `enum` select option. Since enum is only supported by MySQL.

marktopper:
@handiwijoyo, I have now made a check to only show the `enum` type when for `MySQL` connections.

Issue #531, opened 2017-01-11
Improve Voyager tables


abdgad:
Hello,
I have some suggestions to better organize and handle Voyager tables.
Currently, tables that are required for Voyager to function, let's call them Core Tables, are mixed with other tables. This creates two main problems:
1. The user can accidentally delete or modify one of these tables and the app will break.
2. Collision: Voyager uses common table names. This can cause collision if Voyager is installed in an existing app that relies on similar table names.

**Suggestions:**
- We need to identify the core tables that Voyager relies on, and make them immutable so the user won't accidentally make a breaking change. The current way to do this is to add them in Voyager config database.tables.hidden by default.
- To fix collision, we need a prefix to distinguish Voyager tables from the rest. Maybe something like: _voyager_menus_

I believe these are the current Core Table:
- data_rows      
- data_types     
- menu_items     
- menus                         
- permission_role
- permissions              
- roles          
- settings       
- users

akazorg:
If you decide to change this, perhaps you should consider on having these table names defined on a config file, so we could name them -.+

abdgad:
@akazorg the table names will be defined in the corresponding model. No need to define them in the config.

akazorg:
@abdgad you're right.

fletch3555:
This is being tracked elsewhere.  Closing due to age and inactivity.

PR #532, opened 2017-01-11
Improve tables

abdgad:
see #531

- [x] Hide Voyager Core tables from Admin panel by default.
- [X] Add _voyager__ prefix to Voyager tables.
- [ ] Add option to display hidden tables from the Admin panel. Do not allow Edit/Delete actions. Only BREAD actions are allowed.

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -15,7 +15,7 @@ return [
         'add_default_role_on_register' => true,
         'default_role'                 => 'user',
         'admin_permission'             => 'browse_admin',
-        'namespace'                    => App\User::class,
+        'namespace'                    => TCG\Voyager\Models\User::class,
     ],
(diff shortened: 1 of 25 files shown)


abdgad:
@handiwijoyo Hidden Tables will not show by default.
I am thinking to add an option to display them from the Admin panel, but NOT allow edit/delete actions. We just allow to add BREAD to them...
What do you think?

marktopper:
NOTE: This is breaking changes. We need to handle this correctly.

abdgad:
@marktopper we will handle it once we make all the necessary changes.

marktopper:
Is this still relevant or should it be closed?

abdgad:
This should be replaced by newer PR that addresses prefixing Voyager tables.

PR #551, opened 2017-01-14
Use facade and make method non-static

marktopper:


fletch3555:
Does this need to be in v0.11?

marktopper:
If someone uses the static methods of `TCG\Voyager\Voyager`, it will require a change.
So I think it's smarter to put in `v0.11` and tell people to use the Facade instead when upgrading.

PR #552, opened 2017-01-14
Remove function

marktopper:
Removes calls to `voyager_add_post` on existing.


fletch3555:
What on earth was that function there for in the first place....?

marktopper:
@fletch3555: No clue, it has been on my todo list for quite a while. But would wait till I started working on next minor version.

PR #557, opened 2017-01-14
Another try for Slugify

akazorg:
@marktopper take a look.

Pages are configure for using the slug generator without auto force update (please read code comments at [slugify.js](https://github.com/akazorg/voyager/blob/feature-slugify/publishable/assets/js/slugify.js).

Posts are configured with force update = true, meaning it will always update the slug input. (This can be change on "bread edit").

Bread will accept a JSON object with this structure:
```
{
    "slugify": {
        "origin": "title", 
        "forceUpdate": true    // This is optional, default is false
    }
}
```
For testing, just install with dummy data, and goto Posts/Pages and try to Add + Edit something.

Happy coding ^^


marktopper:
Thanks!

Well, might do some cleanup and also fix the issue with Travis failing.

akazorg:
hey @marktopper, you're welcome!
I don't know what happen with Travis, any idea?

About the helpers approach, do you like that way? It might get cleaner this way, so we can have multiple 
helper files per subject.

akazorg:
**Slug preview**

peek_bread_2017-01-15 02-35

fletch3555:
Travis error:

```
There was 1 failure:

1) TCG\Voyager\Tests\RouteTest::testGetRoutes
http://localhost/admin/posts/create did not return a 200
Failed asserting that 500 matches expected 200.

/home/travis/build/the-control-group/voyager/tests/RouteTest.php:64
```


The PostController::create action is throwing a 500 error.  I haven't looked into the code to figure out why yet.

akazorg:
I tried but can't find why Travis is so upset xD, Thanks @fletch3555

marktopper:
Will look into Travis now.

However a comment, I'm not sure I like the global functions that are specific for one use case.
Thoughts @fletch3555?

publishable/views/posts/edit-add.blade.php
```diff
@@ -121,7 +121,10 @@
                         <div class="panel-body">
                             <div class="form-group">
                                 <label for="name">URL slug</label>
-                                <input type="text" class="form-control" name="slug" placeholder="slug" value="@if(isset($dataTypeContent->slug)){{ $dataTypeContent->slug }}@endif">
+                                <input type="text" class="form-control" id="slug" name="slug"
+                                    placeholder="slug"
+                                    {!! isFieldSlugAutoGenerator($dataTypeContent, "slug") !!}
```
marktopper, line 132: When modifying BREAD `$dataTypeContent` is defined, but when creating bread it's not defined.
It should handle that since TravisCI fails because of this.

akazorg: I'll take care of that

akazorg:
This branch was deleted by mistake.

PR #560, opened 2017-01-14
Add configurable models namespace while creating BREAD

antranapp:
Most of us when working on a large application will not leave the models in top level App folder. This is just a convenient setting to have a default configurable namespace for the models, so that we don't have to adapt the model name in the BREAD form every time when the model classes are not in default App folder.


marktopper:
Good idea, however some people might get annoyed if it does not use the application specific namespace if they changed it from `App\`.

So how would you feel about leaving the setting empty, meaning that it would use
```
$this->getAppNamespace()
```

marktopper:
After a closer look that seems to already be the case.

Can you confirm that?

antranapp:
Yes, the configuration will fall back to getAppNamesspace() if user doesn't specify any models.namespace configuration.

marktopper:
I have made a commit to this to use the default application specific namespace by default.
If this is fine with you I will go ahead and merge this in.

antranapp:
It's ok for me

PR #566, opened 2017-01-14
Merge from master

marktopper:


fletch3555:
@marktopper, I don't think we need PRs for these.  You (or I) can do it directly through git and push the changes up to github.

marktopper:
@fletch3555, it's fine with me. I however just find it easier doing it using pull requests.

But will look into doing it the other way from now on ;)

PR #571, opened 2017-01-14
[WIP] Add validation rules

marktopper:
I'm not fully done testing that this works correctly.


akazorg:
Hello @marktopper!
This is a very interesting subject, since we can handle it in so different ways.

I was curious for your approach, I see you opted for the extra table, in fact we can gain a lot of features like user/author control, history, etc. But do we really need this?

I'd rater see the JSON approach, with the spatie/translatable. I have a working version of that, but need some time to create a fork. We can work on something together if you want.

Some screen shoots (you can also see the slugify in action):

**BREAD**
peek_bread_2017-01-15 02-35

**Menu**
peek_menu_2017-01-15 02-39

fletch3555:
@akazorg, I think you commented on the wrong PR

marktopper:
Nope, it's the right one. Have to test out the rules work as expected :)

PR #572, opened 2017-01-14
Add default value for relationship

antranapp:
Imagine you have a website where you have many moderators who can add products in different categories. Depending on the user_id of moderators, you want that they should **by default**  add products to a specific category. Normally the moderators have to choose the correct category all the time. If we can set the default value of the category automatically depending on user_id of the moderators, it would be great.

I'm thinking of different approaches for this issue.
In this pull request, I've implemented one approach that I think, it could be simple and flexible enough for a quick win.
Basically, parallel to a method for relationship in the model class, you add another method with the same name prefixed with keyword `default`, which returns the default value for that relationship.

Example for the field `user_id` of a Feature Model
````
<?php

namespace App;

use App\User;
use Illuminate\Support\Facades\Auth;

class Feature extends Model
{
    public function user_id()
    {
        return $this->belongsTo(User::class);
    }

    public static function default_user_id()
    {
        return Auth::user() ? Auth::user()->id : NULL;
    }
}
````

So depending which user is logged in, we have a different  default user_id in BREAD form
screen shot 2017-01-14 at 19 48 26

screen shot 2017-01-14 at 19 48 40

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -87,14 +87,15 @@
                                         @if(isset($options->relationship))
                                             {{-- If this is a relationship and the method does not exist, show a warning message --}}
                                             @if( !method_exists( $dataType->model_name, $row->field ) )
-                                                <p class="label label-warning"><i class="voyager-warning"></i> Make sure to setup the appropriate relationship in the {{ $row->field . '()' }} method of the {{ $dataType->model_name }} class.</p>   
+                                                <p class="label label-warning"><i class="voyager-warning"></i> Make sure to setup the appropriate relationship in the {{ $row->field . '()' }} method of the {{ $dataType->model_name }} class.</p>
                                             @endif
 
                                             @if( method_exists( $dataType->model_name, $row->field ) )
                                                 <?php $selected_value = (isset($dataTypeContent->{$row->field}) && !is_null(old($row->field, $dataTypeContent->{$row->field}))) ? old($row->field, $dataTypeContent->{$row->field}) : old($row->field); ?>
                                                 <select class="form-control select2" name="{{ $row->field }}">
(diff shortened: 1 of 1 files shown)


fletch3555:
I'm thinking this is a little too specific of a use-case to be added to Voyager core.  You're more than welcome to override Voyager and build it into your own app though.

@marktopper, thoughts?

antranapp:
Actually it's not very specific but very general. The use case I described is just for everyone to understand what I want to archive here. The general idea is that: **You can set a default value for a relationship-field.**

The default value for relationship that I implemented here is also optional. You don't have to implement it if you don't want to.

One of the most popular use case for this is something like `author_id` for models like `Page` & `Post`which are currently set to current admin implicitly by hooking to the `save` method of the corresponding models.

marktopper:
How about that a default value of the JSON parameters could be a callable function like in this case `App\Feature@default_user_id`. That will make it more extendable?

Thoughts @fletch3555 and @peacemoon?

antranapp:
I've thought about it too. There are 2 possibilities in my head for this JSON structure

**1. `default` in top level.**
````
{
    "default": "App\Feature@default_user_id",
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
````

**2. `default` inside relationship**
````
{
    "relationship": {
        "key": "id",
        "label": "name",
        "default": "App\Feature@default_user_id"
    }
}
````

what do you think which one is better @marktopper ?

marktopper:
It is already possible to set a default value for any input type.
So we should stick with that (which is option 1).

antranapp:
Now you can configure the default value via a callable function like this:
````
{
    "default": "App\\Feature@default_user_id",
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
````

PR #579, opened 2017-01-16
Set models name base on the singular name of the table

nikkuang:
Currently, when creating a new table using voyager database tools and enabling the option for creating a model. Voyager will create a class base on the table name. (families =>  App\Families).

When you create a bread. the UI default suggested value for the model name is the singular value  "App\Family" which causes a problem if the users just wanted to use the default values. 

Editing bread would thrown an error "Class App\Family does not exist"


marktopper:
Thanks

PR #582, opened 2017-01-16
Refactoring index of database tool

Frondor:


Frondor:
Seems I have to correct some checks

PR #584, opened 2017-01-16
Refactoring database tool index and saving some queries

Frondor:
There was some evil `php` logic in the view, also a query inside a loop. Moved everything to the controller and cleaned the view.


abdgad:
I think it won't hurt if we change DBSchema::tables() to return an array of just table names, because it's only used in one place, in database index view.
```php
$tables = DB::select($query);

return array_map(function($table) {
    return current($table);
}, $tables);
```

abdgad (changes requested):

abdgad:
there is a small bug: hidden tables are not handled properly.
`in_array($table, config('voyager.database.tables.hidden', []));`

Frondor:
Thanks for pointing that out, just fixed it.
The test for that route seems to be failing when accessing to `/admin/database` route, although in production I have no problems at all.

PR #587, opened 2017-01-16
Fix #581: setting field content to null value while updating

### LINKED ISSUE #581: Cannot update a category which already has a parent category to get a no-parent category (parent_id = null)
- Laravel Version: 5.3
- Voyager Version: 0.10.10
- PHP Version: 5.6

**Description**:
When updating a category to get a null parent_id instead of a value, it doesn't work and the parent_id is not updated and not set as null.

Here is the content of the options field of the parent_id column in the BREAD

`
{
    "default": "__null__",
    "null": "__null__",
    "options": {
        "__null__": "Main"
    },
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
`

and it is a dropdown input type.

**Steps To Reproduce**:

1. Add a category with a null parent_id.
2. update it to get a parent category.
3. update it to get a null parent_id. 

**Result**:  The update action has no effect  and the category is still with the parent category selected in step 2.

antranapp:


marktopper:
Not exactly sure why TravisCI fails.

Any ideas?

fletch3555:
The issue is that the password field of the User model is set to "NOT NULL", but this change is allowing null password values to be accepted, which breaks at the DB level.

antranapp:
@fletch3555 @marktopper I was too quick with the fix. It should be correct now :)

src/Http/Controllers/Controller.php
```diff
@@ -53,9 +53,11 @@ public function insertUpdateData($request, $slug, $rows, $data)
             $content = $this->getContentBasedOnType($request, $slug, $row);
 
             // Set the content back to the previous value when there is really now input for this field
-            if (is_null($content) && is_null($request->input($row->field))) {
-                if (isset($data->{$row->field})) {
-                    $content = $data->{$row->field};
+            if (is_null($content)) {
+                if (is_null($request->input($row->field))) {
+                    if (isset($data->{$row->field})) {
```
fletch3555: This(L58) can be combined into the `if` above it(L57).  I'm not a huge fan of cascading conditionals, especially when they're related like this.

antranapp: I will change the code this time since I don't want to get into a discussion about coding style. I have enough stress with my colleague in my day-job for such discussion already.

But: (no offense, just my opinion)
why not when it improves the readability.  We have a php_cs to watch for the coding style and there is no complaint about that. Furthermore I would love to see more code comments . I've had a hard time to understand some part of the code. Next time you are warmly welcome to change my pull request directly if you want to change anything in coding style. I'm not mocking you here, @marktopper did that already in one of my pull request and I'm totally fine with that :) . Cheers.

marktopper: Code comments is really needed, I will take a good look at the code before we release the stable version `v1.0` and make sure that everything is commented and that all classes methods are minimal and splitted up into severals. It's on my todo for the `v1.0` release. Which I hope to come within a month.

Issue #580, opened 2017-01-16
[BUG] Media Library Cursors Event


akazorg:
If you click to rename an image, and then use the keyboard cursors, the image will change on the background.
This JS popup really requires revision, on the menu it also fail sometimes.

![peek_bug_media 2017-01-16 06-16](https://cloud.githubusercontent.com/assets/34574/21972718/54ef21d2-dbb3-11e6-8a87-c434cd29c835.gif)

akazorg:
Sending fix for this.

marktopper:
Fixed in #588 

/closed

PR #588, opened 2017-01-17
Fixing #580 Media Library Key Events

Frondor:
I had an idea about how to properly fix this, but in my opinion, the whole component should be a Vue.js one.

##
Besides of fixing the main issue #580 this PR also cache the `#files` jQuery element (a little of performance should always be welcome), and fix the problem with scrolling up and down the page when browsing files.

### DIFF
--- a/publishable/assets/js/media/media.js
+++ b/publishable/assets/js/media/media.js
@@ -111,9 +111,9 @@ var VoyagerMedia = function(o){
 		});
 
 		$(document).keydown(function(e) {
-			var curSelected = $('#files li .selected');
 			if ( isBrowsingFiles ) {
-				var curSelectedIndex = curSelected.data('index');
+				var curSelected = $('#files li .selected'),
+				curSelectedIndex = curSelected.data('index');
 				// left key
 				if( (e.which == 37 || e.which == 38) && parseInt(curSelectedIndex)) {
 					newSelected = parseInt(curSelectedIndex)-1;
--- a/publishable/assets/js/media/media.js
+++ b/publishable/assets/js/media/media.js
@@ -87,18 +87,18 @@ var VoyagerMedia = function(o){
 		
 		//********** Add Keypress Functionality **********//
 		var isBrowsingFiles = null,
-			fileBrowserActive = function(el){
-				el = el instanceof jQuery ? el : $(el);
-				if ($.contains(files.parent()[0], el[0])) {
-					return true;
-				} else {
-					$(document).off('click');
-					return false;
-				}
-			},
-			handleFileBrowserStatus = function (target) {
-				isBrowsingFiles = fileBrowserActive(target);
-			};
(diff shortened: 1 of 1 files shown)


akazorg:
Hey @Frondor, I also sent a #589 minutes ago, almost at the same time!! very funny :)

You're right, this should be re-factored in Vue.js.

Frondor:
@akazorg What a bad team we are! :laughing: 
Passing the whole thing to Vue so we can get rid of jQuery would take a whole day or 2, But yes, in terms of performance and maintainability Vue is the way to go.

Speaking about this fix, I just tried enabling/disabling an event listener to know when the `#file` element loses its focus state so the key events don't populate and make it work not only for modals, but also other elements that may be in the same page that you can manipulate with keys.
Just like windows explorer works, you have to actually focus the window to navigate between the files using keys.
I also added `files.trigger('click')` to `getFiles()` so when it executes, you don't have to click on the file browser to start using the keys.

akazorg:
@Frondor what a team!! xD you can hang on slack.
I'm not used to Vue.js so I can't give much help with this atm.

The main problem is related to the Bootstrap Modal, when open. That's why I tracked that and added a validation for the KeyPress.

Frondor:
@akazorg yeah, that's ok if you want to solve this only for bootstrap modals, but what if there's a, let's say `select2` control in the same page? You have to stop the `$document.keypress` event in that case as well...
Obviously, talking about perfomance, because you can easily add more conditions to each `if( (e.which == ## ...` (kinda ugly and messy)
I'm not a big fan of unused active events in the document :thinking:

akazorg:
@Frondor, I see what you mean, but we don't have any select at the Media.

In my opinion, the original issue is sorted out, so I'm done with this issue.

Maybe we should open another issue for the Media, and discuss the subject.

Frondor:
@akazorg Then your PR is more like a "patch" not a fix (no offense meant ofc), because it limits the `keydown` handling by the specific case of a bootstrap modal. Have in mind the whole project is still an alpha, and many changes are about to come. Why not to think forward instead of a specific case?

We don't know if voyager shall keep making use of bootstrap modals in the future, neither if some other "key controls" are going to be added to the same page (they don't need to be in the same view to conflict).

This PR also fixes other things in a more elegant way (imho :laughing: ). I'll explain them in the main post

akazorg:
The issue I initially reported is gone, so i'm done with my lovely "patch".

Maintainers will decide the best.

Feel free to continue working on this.

Frondor:
Seriously, I didn't mean to sound offensive.

akazorg:
It's ok :) I just have other things to do, and this is working now.
That bug was really urgent because a User could delete/rename another file. Really, feel free to keep working on that, there are many things to do, but I can't give much help with the Vue atm, so "i'm done" here :)

marktopper:
Thanks 👍

PR #589, opened 2017-01-17
Fix 580 media library cursors events

akazorg:


marktopper:
Sorry for first seeing this now.
Just merged the other pull request for this.
However this also worked fine, thanks for your time, sorry for not merging.

akazorg:
It's ok @marktopper, I wont cry for that :)
I do prefer my version ^^, but what really matters is that ugly bug is fixed, and I guess @Frondor's version may be better since he had some additional checks.

PR #594, opened 2017-01-17
Fix database table delete button not working properly

abdgad:
Currently in Database index, when the table delete button is clicked, _event.target_ is used, which refers to the exact element that triggered the click event.
This results in an issue if you click on the trash icon on the delete button, _event.target_ will be set to the _\<i\>_ element, which is not what we want. We want the _\<a\>_ element which has information about the table. So we use _this_ instead of _event.target_.

Steps to reproduce the bug:
Click on the trash icon on delete button.


handiwijoyo:
Thanks man!

PR #617, opened 2017-01-22
Add new entities on dashboard with setting

bestmomo:


marktopper:
Please describe what this would do

bestmomo:
On dashbord there are 3 entities 
- users
- posts
- pages

When we add new entities they dont come on dashbord and the view is in package, I didn't try to override it.
I think it would be interesting with a simple config setting to add them on dashbord (example : http://laravel.sillo.org/wp-content/uploads/2017/01/Capture-82.png).

marktopper:
I see, good idea 👍

marktopper:
I marked this as open for discussion since this can be done in multiple other ways.

I will try to think about the best solution for this.

bestmomo:
Sure there are many ways to do that.

marktopper:
I think this is fine for now.

Before I merge this in, we must fix the TravisCI issue.

bestmomo:
Looks like it doesn't like my comments...

marktopper:
Whoops, wrote `CSS` instead of `Config` in my last commit. 😂

Well, hopefully this will fix the TravisCI issue.

marktopper:
We can ignore StyleCI, that will automatically be fixed on merge.

marktopper:
Seems to be fixed, merging in.

marktopper:
This needs to be documenated

bestmomo:
Is documentation in a repo ?

marktopper:
We have the documentation [here](https://github.com/the-control-group/voyager/blob/gh-pages/docs/v0.10/Readme.md) in branch `gh-pages`.

However, if you can not figure it out I will do it 👍

bestmomo:
Done with this commit : https://github.com/the-control-group/voyager/pull/631

Hope my english is not so bad :)

marktopper:
I will go through it later 👍 
Thanks

fletch3555:
@bestmomo, can you update this (add a new PR) to include the config for the existing three blocks and remove them from the view? I'd like to simplify it so it's either all hard-coded, or all handled by a loop.

Thanks!

bestmomo:
I was thinking about something like that but I must dig a bit because of icon and grid system. And there is a bug on route...

PR #621, opened 2017-01-23
Missing URL to Voyager Icons in Edit Menu Item modal

Aym3nTN:
When I tried to edit a menu item which I created without an icon, I figured out that the URL to Voyager Icons was missing....


marktopper:
Thanks!

PR #631, opened 2017-01-25
Update Readme.md

bestmomo:
Add admin setting


marktopper:
I have just decided that we should call it something else in the root config before I make a new release tag.

What do you think about `widgets` or `panels`?

bestmomo:
Right, or simply `dashboard`, as you want.

PR #623, opened 2017-01-24
Improve the Database Manager

abdgad:
This PR aims to improve the Database Manager.
Still a work in progress...

PR #632, opened 2017-01-25
Improve the Database Manager

abdgad:
This PR aims to improve the Database Manager.
Still a work in progress...


src/Database/Schema/Column.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public function getDoctrineColumn()
+    {
+        return $this->doctrineColumn;
+    }
+
+    public function setNull($null)
+    {
+        $this->doctrineColumn->setNotnull(!$null);
```
marktopper, line 59: What exactly does this method?

src/Database/Schema/SchemaManager.php
```diff
@@ -0,0 +1,75 @@
...
+use Doctrine\DBAL\Schema\Column as DoctrineColumn;
+use Doctrine\DBAL\Schema\SchemaException;
+use Doctrine\DBAL\Types\Type;
+use Illuminate\Support\Facades\DB;
+
+class SchemaManager
+{
+    public static function manager()
+    {
+        return DB::connection()->getDoctrineSchemaManager();
```
marktopper, line 18: Love this approach 😍

marktopper:
Really nice work 👍 
Keep it up!

_You may ignore StyleCI, it will be fixed when merging_

src/Database/Schema/Column.php
abdgad: it does the opposite of setNotnull().
so we can use $this->setNotnull or $this->setNull, it doesn't matter.

src/Database/Schema/Column.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace TCG\Voyager\Database\Schema;
+
+use Doctrine\DBAL\Schema\Column as DoctrineColumn;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Schema\Comparator;
+
+class Column
```
fletch3555, line 9: What was the reason for throwing out what you did in #623? Specifically, extending DoctrineColumn, as opposed to how you're doing it here by passing it into the constructor.

abdgad: The reason I am passing this to the constructor is that Doctrine DBAL already has methods to retrieve and setup columns from the database, and extending means that I will have to create special methods to populate the columns from the database, and that will add more complexity. So I can use Doctrine methods to retrieve the Column from the database and pass it into the constructor.
Anyways, I have actually stopped working on this. I am currently working on a much simpler approach that does what need instead of trying to do many things and end up overcomplicating the whole thing. I will close this PR and publish the new one tomorrow.

abdgad:
I will publish a new PR tomorrow cause this one is just OVERcomplicated. It ended being a mess unfortunately with no actual benefits.

abdgad:
replaced by #648 :D :D :D :D

PR #638, opened 2017-01-26
Parse BREAD relationships to show labels instead of keys

Frondor:
Basically, it turns this

into this

You can optionally add a new `page_slug` property to the `relationships` object in the BREAD details in order to display proper links to relationship records. i.e.:
```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "page_slug": "admin/users"
    }
}
```

if no `page_slug` is provided, then the real "text" (`relationship.label`) is still being "resolved", but no anchor link is created.

**EDIT:** Also changed some other things.
## This is the list of additions/improvements/fixes
 - Parse BREAD relationships
 - Added `format` option to `date` row type to display a formatted date in READ/BROWSE views using Carbon's `formatLocalized()` method
 - [Parse BREAD checkboxes](https://i.gyazo.com/8975ef2bfd53742393e5abbd13c998f1.png) to display on/off labes instead of 0/1 booleans
 - [Added *edit* button](https://i.gyazo.com/4199250949dbbb084ca33a57207a8e51.png) to READ view
 - Drank 8 cups of coffee and didn't sleep at all.


Frondor:
Found a bug when editing, let me dig into it


Any clue on why this is working `$model->with(['service', 'user'])->get()` but this isn't `$model->with(['service_id', 'user_id'])->get()` ?
Talking about **[this line](https://github.com/Frondor/voyager/blob/fcd027b8b24e01c0e9e6f70b1596d261842316fa/src/Http/Controllers/VoyagerBreadController.php#L45)** 

both methods are equally the same in my model 
```
    public function service()
    {
        return $this->belongsTo('App\Service');
    }

    public function user()
    {
        return $this->belongsTo('App\User');
    }

    // Voyager BREAD
    public function service_id()
    {
        return $this->belongsTo('App\Service');
    }

    public function user_id()
    {
        return $this->belongsTo('App\User');
    }
```

Frondor:
Bug fixed, the whole BREAD is working now but it was needed to use `camelCase` relationships (which follows the PSR-1 standard) instead of the actual `snake_case` ones.

the [documentation](https://the-control-group.github.io/voyager/docs/v0.10/#core-concepts-bread-validation) must be updated to 

> Note: the method used for this relationship, must match the camelCase version of the row's name from the pages table. Which is why we used authorId as the method name to tie the relationship.

**And all the relationships must have it's foreign key set.**

A proper example for an `user_id` (column name) relation:
### Row details:
```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "page_slug": "admin/users"
    }
}
```
`page_slug` is optional

### Model method for this relationship
```
public function userId()
{
    return $this->(User::class, 'user_id');
}
```

TIP: Tested with `select_dropdown` (one-to-one) relationships, not one-to-may. So some testing will be much appreciated

Frondor:
Travis is throwing a
```
[Symfony\Component\Debug\Exception\FatalErrorException]                      
  Trait 'TCG\Voyager\Http\Controllers\Traits\BreadRelationshipParser' not fou  
  nd 
```
but it is there... (in my fork) How to deal with that? Must it be "manually" created by some maintainer first?

marktopper:
I will investigate that.

marktopper:
Can you link me to the trait? I can't seem to find it either.

Frondor:
@marktopper nvm, I placed it in the wrong folder, my bad.

I'm fixing it and also adding support for formatting dates (dataType rows?) via row details using Carbon's `formatLocalized()`:
````
{
    "format": "%A $d $B $Y"
}
````

I know that should be another PR but I don't want to generate conflict between files, and since Voyager is still in alpha... :roll_eyes:

**Comment by Frondor** [comment]
Alright, everything seems fine.
If someone wants to test it, please, go ahead.

I've added a little changelog to the end of the main post.

File: resources/views/bread/edit-add.blade.php [hunk 1]
```diff
@@ -90,16 +90,22 @@
                                         @endif
                                         <input type="file" name="{{ $row->field }}">
                                     @elseif($row->type == "select_dropdown")
```
fletch3555, line 92: @marktopper, please note that these lines were moved to their own view partials in v0.11.  This will likely have to be updated manually before the release.

marktopper: Yes, will need to make them into their separate view before merging `master` into `v0.11`.
Anyhow, thanks for reminding me.

marktopper:
Will test it out

Frondor:
@marktopper Thank you! Is there anything I can do regarding to @fletch3555 review?

marktopper:
@Frondor, thats first when merging this into the `v0.11` branch.

marktopper:
Also it seems to work like a charm 👍

hazem1taha:
After the latest release v0.10.12, The parent category dropdown field in the edit category page is not showing the parent category, instead.. it shows the "Main Category" option (even if the category has a parent category) when using this json option

    {
    "default": "__null__",
    "null": "__null__",
    "options": {
    "__null__": "Main Category"
    },
    "relationship": {
    "key": "id",
    "label": "name"
    }
    }

I think this problem is related to this PR
@Frondor @marktopper

Frondor:
I don't know if I understood, but the "Main Category" will show as long as you set the `options` property.
Anyway, I don't know if it was what you meant but found a bug when the default selected option for this kind of case, wasn't properly calculated.

I already got the fix for this, but can't send a new PR because already have another one "on hold"... I think I have to create a new branch and reset its HEAD but to be honest, I'm not that good at git :D

PR #645, opened 2017-01-27
Widgets in settings

bestmomo:


Frondor:
I guess "widgets" should be something we must register externally using packages and some `Voyager` facade method. This way we can override that `voyager::index` blade view and place the widgets the way we want, using the new Blade Components so we keep the design line of panels/boxes modules.
And each `$component` might be something predefined like tables, static images, sliders, lists, forms, etc...
It would be a nice feature to make.

i,e.: `voyager::registerWidget($name, $component, $php_class, $roles_or_permissions, $css_classes, ...$extras)`
````
<div class="container">
{!! Voyager::widget($name) !!}
</div>
````

Anyway, I always like to override the index view and build it from the ground up to my taste :D

**Comment by bestmomo** [comment]
Needs are so varied that I wonder if a collection of widgets would be judicious. I think that this view will be often overrided.

**Comment by marktopper** [comment]
Maybe make it into a `@include('widgets')` or something.

**Comment by marktopper** [comment]
For now this is good.
I have added a todo on my side to change this approach in a further version to allow packages directly register and disable widgets along with additional handlers for the grid system.

About the StyleCI failure, ignore that. It will get fixed on merge.

About the TravisCI failure, then I will look into that.

**Comment by marktopper** [comment]
TravisCI with the following reason:

```
Failed asserting that the page contains the HTML [1 Users]. Please check the content above.
```

It looks into the dashboard and tries to find the the text `1 Users`.
But I believe it's replaced with this now:
```
You have 1 User(s) in your database.
```

Will make a commit to change the test of this.

marktopper:
It also seems that the default `widgets` is not shown and therefor TravisCI continue to fail.

marktopper:
The error was with the tests.
I have fixed the issue now.

If this looks fine with you @bestmomo, let me know and I will merge.

bestmomo:
Fine for me, thanks.

tnylea:
@bestmomo Thanks for this configuration :) This was something I needed in my app I'm trying to get out the door 👍

marktopper:
@tnylea: This might be changed on `v0.11`. #692

PR #646, opened 2017-01-28
Media Directories & Files 500 error to Media Directories & Files MisMatchToken exception

sarath-c-sandwich:
Laravel: 5.3.*
Voyager: 0.10.11  but error since version 0.10.9
PHP: 5.6.28
=======

The Directories & Files does not load. In stead it show 500 Internal Error. Does it caused from CSRFToken Middleware?

+ I tried to rebuild symlink many time but still not work.
screen shot 2017-01-28 at 1 29 07 pm


marktopper:
There are so many commits in this, can't seem to figure out what the actual change it.

PR #647, opened 2017-01-28
Handle errors and revert all new save data type when error occured

nikkuang:
Added DB transaction when saving Data type with lots of fields so it won't messed up the data when an error occurs.

Handle Errors like Saving a none existing model. Show an Alert instead of crashing the app


marktopper:
Great job 👍

PR #658, opened 2017-01-30
Merge Config on register to avoid problems for existing applications

nikkuang:
Since a new config called Widget was added, existing applications won't see the default widgets without republishing the config


marktopper:
Great thinking, thanks!

PR #672, opened 2017-02-02
Multiple additions and improvements.

Frondor:
- [x] Make use of [readmore](https://i.gyazo.com/034e420277ddb8a74b945c64c8d4f477.png) library (was unused until now) for `text`, `text_area` and `rich_text_box` input types in BROWSE views.
- [x] * Allow `select_multiple` input type to simply [work like a `select_dropdown`](https://i.gyazo.com/5ee31264726e89fd8b031bc2c8da7c5c.png) (with `options` instead of `relationships`) with `JSON` columns.
- [x] Properly display of `select_multiple` rows for both `relationships` and simple `options` in [READ and BROWSE](https://i.gyazo.com/fc816559b833e1c18191ac3ccb1ab734.png) views
- [x] * Added support for `description` property in each BREAD row's `details` section, [to display it below each input](https://i.gyazo.com/c4970c93849e4e68410e61b93418cbdd.png).
- [x] Implemented logic to make use of `checkbox`es `checked` property when creating and editing
- [x] * Modified `validation` logic for BREAD rows so it can be setup either as a string like `"rules": "required|max:80"`, or an array:
```
{
    "validation" {
        "rules": [
            "required",
            "max:80"
        ]
    }
}
```
and if its a string, convert it to array, so you don't have any problem when using complex `regex` rules.

My apologizes for the spanish demos :laughing: 

--  * These changes should be documented, along with the changes in #638 regarding to `date`'s `format` property

### DIFF
--- a/publishable/assets/js/app.js
+++ b/publishable/assets/js/app.js
@@ -2,6 +2,8 @@ $(document).ready(function(){
 
   $('#voyager-loader').fadeOut();
   $('.readmore').readmore({
+    collapsedHeight: 60,
+    embedCSS: true,
     lessLink: '<a href="#" class="readm-link">Read Less</a>',
     moreLink: '<a href="#" class="readm-link">Read More</a>',
   });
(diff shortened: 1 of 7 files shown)


marktopper:
Great work 👍

Issue #676, opened 2017-02-04
Document: /admin/upgrade no longer use.


sarath-c-sandwich:
- Laravel Version: 5.3.*
- Voyager Version: 0.10.12
- PHP Version: 5.*
- Database Driver & Version: MariaDB

### Description:
That should inform user in a very clear message that they don't need to run domain.com/admin/upgrade anymore if they are currently on Voyager v.0.10.9+
VoyagerUpgradeController no longer use if user are currently based on the 0.10.9+

<img width="937" alt="screen shot 2017-02-04 at 10 14 07 am" src="https://cloud.githubusercontent.com/assets/1330852/22615230/2a6478da-eac3-11e6-81e3-6b335ee72188.png">

marktopper:
Closing due to #679

PR #679, opened 2017-02-05
Documentate when to run the upgrade script.

marktopper:
Reference #676


fletch3555:
Is v0.10.6 correct? #676 referenced v0.10.9+.  Just checking

marktopper:
I just looked at the method for the upgrade script.
Might however look more into it.

marktopper:
[It is `v0.10.6`.](https://github.com/the-control-group/voyager/releases/tag/v0.10.6)

PR #683, opened 2017-02-05
Extend the documentation with changes from #672

marktopper:
Made in #672

### DIFF
--- a/docs/v0.10/Readme.md
+++ b/docs/v0.10/Readme.md
@@ -487,6 +487,18 @@ Additionally, you may wish to add some custom error messages which can be accomp
 }
 ```
 
+Since `v0.10.13` you can do the `required` and `max:12` rule the following way:
+```
+{
+    "validation": {
+        "rules": [
+            "required",
+            "max:12"
+        ]
+    }
+}
+```
+
 ### Generating Slugs
 
 Using the bread builder you may wish to automatically generate slugs of a cetain input. Lets say you have some posts, which have a title and a slug. If you want to automatically generate the slug from the title attribute, you may include the following *Optional Details*:
--- a/docs/v0.10/Readme.md
+++ b/docs/v0.10/Readme.md
@@ -595,6 +595,8 @@ public function categories(){
 
 Now, when you have save the results from your Many-to-Many relationship the ID's of each selected value will be synced and added to your pivot table.
 
+> Please note that this can also be configured as a normal `Select Dropdown`.
+
 ### NULL values
 
 You might want to save an input field into the database as a `null` value instead of an empty string.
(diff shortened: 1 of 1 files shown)


Frondor:
Very nice! I think we can include:

- The new `description` property for all data types
- The new way of creating relationships with camelCase names instead of `author_id()` 
- The new property for `relationship`: 
```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "page_slug": "admin/users"
    }
}
```
To display them as links.
- Point out that `select_multiple` can also be configured as a normal `select_dropdown` (without `relationship` but `options`).
- And finally, the new `format` property for date fields, which formats the input with `Carbon::formatLocalized()`

marktopper:
Thanks @Frondor. I have made changes according to your feedback.
Please let me know if you notice anything else or have additional comments on my latest commits.

However, I am not much into the BREAD relationship, so could you possibly make a documentation note about the new property for `relationship`?

docs/v0.10/Readme.md
```diff
@@ -461,6 +462,26 @@ This is only valid if you have set your image to be resized. If you specify your
...
 
+#### Date
+
+```js
+{
+    "format" : "Y-m-d"
+}
+```
+
+The date input field is where you can input a date. In the JSON above you can specify the `format` value of the output of the date.
```
Frondor: The date input field is where you can input a date. In the JSON above you can specify the `format` value of the output of the date. It allows you to display a formatted `date` in _browse_ and _read_ views, using Carbon's `formatLocalized()` method

docs/v0.10/Readme.md
```diff
@@ -529,12 +562,12 @@ The **key** above will be the row that we want to use as the value in the select
...
 
 ```
-public function author_id(){
+public function authorId(){
     return $this->belongsTo(User::class);
 }
 ```
 
-> Note: the method used for this relationship, must match the name of the row from the `pages` table. Which is why we used `author_id` as the method name to tie the relationship.
+> Note: the method used for this relationship, must match the name of the row from the `pages` table. Which is why we used `authorId` as the method name to tie the relationship.
```
Frondor: > Note: the method used for this relationship, **must match the camelCase version** of the row from the `pages` table. Which is why we used `authorId` as the method name to tie the relationship.

docs/v0.10/Readme.md
```diff
@@ -573,6 +606,8 @@ public function categories(){
 
 Now, when you have save the results from your Many-to-Many relationship the ID's of each selected value will be synced and added to your pivot table.
 
+> Please note that this can also be configured as a normal `Select Dropdown`.
```
Frondor: > Please note that this can also be configured as a normal `Select Dropdown`.

Please note that this can also be configured as a normal `Select Dropdown` in case you need a simple multiple selection with fixed `options` instead of `relationship`.

Frondor:
@marktopper very nice! I just sent a review with a few suggestions regarding to some strings.

Now, about the relationships, I've explained it in that PR, you can use this text on this PR

> You can optionally add a new `page_slug` property to the `relationship`s object in the BREAD details in order to display proper links to relationship records. i.e.:
> 
> {
>     "relationship": {
>         "key": "id",
>         "label": "name",
>         "page_slug": "admin/users"
>     }
> }
> if no `page_slug` is provided, then the real "text" (`relationship.label`) is still being "resolved", but no anchor link is created.

marktopper:
Thanks @Frondor. Updated the pull request.

PR #695, opened 2017-02-07
changed bread table id to name & some fixes for db editing

ctf0:
- changed bread table id to name to help user id which table he is editing
- fix removing a field, db gets updated but still give error
- fix removing a unique field could also give an error
- clear the changed/removed table names from both `data_types &
permissions`
- show `add/view/edit/delete` according to permissions
- add a storage disk option so user can chose any instead of `local`.
- remove `storage.subfolder` as its no longer needed and it may cuz issues with the file
system.
- fixed the post excerpt having too many spaces because of bad formatted view.
- <s>moved all view files to publishable for easier editing.</s>
- put the views back to resources, including the `posts` views
- some cleanup for the db migration & seeders


Frondor:
> fixed row html tags showing up in bread views

Is it related to any issue or something? Can you show what's the problem you're facing? I don't think unescaping everything is a solution. We can use `strip_tags` instead.

> changed bread table id to name

Why? Once again, was that a bug? Did someone report it already?

ctf0:
- for the html, u can clearly see it in the `posts`, which i think showing up html tags to the end user is not right, also its only escaped where it should be

- try to edit any database table, with the id you had to go up to check what is the table name u r editing, but now you have the name clearly in both the url & the breadcrump

Frondor:
@ctf0 

> also its only escaped where it should be

Not really, for example, there's no reason to unescape `select_dropdown` or `select_multiple` options. There  is no content with HTML tags in those.
The only `dataType` that should work with HTML content should be `rich_text_box`, maybe `text_area` as well, but nothing else. And even in those cases, we shouldn't unescape the content for obvious reasons regarding to app security.

I agree that end users shouldn't see raw HTML, but do not unescape the contents.
The solution for this should be to use `{{ strip_tags($field, '<b><i><u>') }}` on `rich_text_box` dataType (and `text_area`? I'm not sure) at `read.blade.php` and `browse.blade.php` views.

ctf0:
actually u r right, i just assumed that this part `{!! $item->{$row->field} !!}` means it only runs on the fields that have its type of `text` not a global escape of course

Frondor:
Thank you for spotting the glitch with raw HTML. I've included a fix for it on #697 

You can fix this PR to keep working on the table names and forget about the raw HTML issue, or open a new one.

ctf0:
i have made a couple of more changes, can u check em as well ?

Frondor:
@ctf0 please, focus on your suggestion about table names and revert the changes in `browse.blade.php`, `edit-add.blade.php` and `read.blade.php`, raw HTML issues are already handled [here](https://github.com/the-control-group/voyager/pull/697/files#diff-cf377d287d89be3fc7d2a9c234b13774).

marktopper:
Can you please update the tests to work with this new urls?

ctf0:
@marktopper i know this sounds very idiotic, but i have no idea how to write test files 😢 .

marktopper:
It's okay. I will look into the tests later then when I get to office.

Frondor:
@ctf0 I don't know if you're aware of but you've added once again the code for unescaping every field. I though you already removed it in ab7b74c 

Please, watch out for the escaping :+1: 

Regarding to `Voyager::for`, that means A LOT of DB queries since its inside a loop. I believe that logic should come from `VoyagerBreadController` and query for all BREAD permissions for that user's role. Then return an array to the view and use something like `in_array` to correctly display the action buttons.

This way we are saving (3 * numberOfRows) queries.

EDIT: Actually, more queries because each `Voyager::can` or `Voyager::for` calls a `User::find(Auth::id());` which doesn't make use of `Auth::user()` or `Request::user()` cache. So it's `((3 + numberOfActions) * numberOfRows)`

src/Voyager.php
```diff
@@ -77,6 +77,21 @@ public static function can($permission)
         }
     }
 
+    public static function for($permission)
```
marktopper: NOTE TO SELF: Remove this method and use `Voyager::can` instead when merging to `v0.11`.

Since in `v0.11` the method `Voyager::can` no longer throws exception.

resources/views/bread/browse.blade.php
```diff
@@ -39,9 +41,9 @@
 
                                                     @foreach($data->{$row->field} as $item)
                                                         @if($item->{$row->field . '_page_slug'})
-                                                        <a href="{{ $item->{$row->field . '_page_slug'} }}">{{ $item->{$row->field}  }}</a>@if(!$loop->last), @endif
+                                                        <a href="{{ $item->{$row->field . '_page_slug'} }}">{!! $item->{$row->field} !!}</a>@if(!$loop->last), @endif
```
marktopper: What is your reason for no longer escaping?

resources/views/bread/browse.blade.php
```diff
@@ -39,9 +41,9 @@
...
                                                         @else
-                                                        {{ $item->{$row->field}  }}
+                                                        {!! $item->{$row->field} !!}
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -55,7 +57,7 @@
                                                     {{ $data->{$row->field}->implode($options->relationship->label, ', ') }}
                                                 @endif
                                             @elseif($row->type == 'select_dropdown' && $data->{$row->field . '_page_slug'})
-                                                <a href="{{ $data->{$row->field . '_page_slug'} }}">{{ $data->{$row->field}  }}</a>
+                                                <a href="{{ $data->{$row->field . '_page_slug'} }}">{!! $data->{$row->field} !!}</a>
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +68,35 @@
                                                     <span class="label label-primary">{{ $options->off }}</span>
                                                     @endif
                                                 @else
-                                                {{ $data->{$row->field} }}
+                                                {!! $data->{$row->field} !!}
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +68,35 @@
...
                                                 @endif
                                             @elseif($row->type == 'text')
-                                            <div class="readmore">{{ $data->{$row->field} }}</div>
+                                            <div class="readmore">{!! $data->{$row->field} !!}</div>
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +68,35 @@
...
                                             @elseif($row->type == 'text_area')
-                                            <div class="readmore">{{ $data->{$row->field} }}</div>
+                                            <div class="readmore">{!! $data->{$row->field} !!}</div>
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +68,35 @@
...
                                             @elseif($row->type == 'rich_text_box')
-                                            <div class="readmore">{{ $data->{$row->field} }}</div>
+                                            <div class="readmore">{!! $data->{$row->field} !!}</div>
```
marktopper: Same

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +68,35 @@
...
                                             @else
-                                                {{ $data->{$row->field} }}
+                                                {!! $data->{$row->field} !!}
```
marktopper: Same

resources/views/bread/browse.blade.php
Frondor: https://github.com/the-control-group/voyager/pull/695#issuecomment-277903990
https://github.com/the-control-group/voyager/pull/695#issuecomment-277912731

marktopper:
⚠️  About the `Voyager::for` method. You must come up with another name, since `for` is a reserved word in PHP5.

ctf0:
@Frondor i did solve it, except this one file which have escaped me, solved now.
- changing the current permission checking to use `Auth::user()` means its now using the `App\User` model so we have to include the `VoyagerUser Trait` inside that model or we gonna have errors.

- `VoyagerAdminMiddleware` have the same problem

@marktopper changed to `have`

Frondor:
Well... the 3 queries are still inside a loop. Like I said, the `hasPermissions` query, with multiple permissions should be called from the controller, and pass the resolved data to the view, something like `(array) $breadPermissions` and instead of `Voyager::have('add_'.$dataType->name))` just `@if( in_array('add_'.$dataType->name), $breadPermissions)  )`

ctf0:
maybe we can instead share a view var that gets all the user permissions across the app and save it in an array, then we can check inside it with `in_array` and inside the view we use 
```blade
@if($allow_role('add_'.$dataType->name))
    ...
@endif
```
and now its only one call to the db across the whole session.

src/Http/Controllers/VoyagerDatabaseController.php
```diff
@@ -135,6 +136,24 @@ public function update(Request $request)
             );
     }
 
+    public function cleanOldAndCreateNew($originalName, $tableName)
```
abdgad, line 138: what is this supposed to do?

ctf0: remove the old entry from permissions & data_types tables, create a model for the new name

src/Voyager.php
```diff
@@ -68,12 +68,24 @@ public static function can($permission)
...
+        }
+    }
+
+    public static function have($permission)
+    {
+        // Check if permission exist
+        $exist = Permission::where('key', $permission)->first();
+
+        if ($exist) {
+            $user = User::find(Auth::id());
```
marktopper: This could become `$user =  auth()->user();`.
Then it will no longer require a SQL query.

The the user could eager-load their permissions, that will help limit the amount of queries run when checking for multiple permissions.
```
if (! $user->relationLoaded('permissions')) {
    $user->load('permissions');
}
```

Then the last step is to remove the part that checks if the permission exists and replace it with a method to holds all the permissions. We could for this use the `Voyager` facade, then the Voyager instance could hold all existing permissions. Maybe something like this:
```
// Check if permission exists
$exists = \TCG\Voyager\Facades\Voyager::permissionExists('visit_admin');

// Method in Voyager instance
protected $permissions;
public function permissionExists($key)
{
    if (is_null($this->permissions)) {
        $this->permissions = Permission::all();
    }

    return $this->permissions->where('key', $key)->count() > 0;
}
```

> I now that this code have been here for a while, and is not your fault 👍

marktopper: Maybe the `User` model should check if the relationship is loaded and else load it instead. Then it will also be working for other methods than `Voyager::for`.

marktopper: @Frondor: Don't you think this would do the trick?

abdgad:
Moving the views to Publishable is not a good idea for now because they will be overwritten when an update comes out and custom changes will be lost.

src/Voyager.php
Frondor: @marktopper I totally agree with those changes. We should also cache the permissions and flush them every time they are updated, like https://github.com/spatie/laravel-permission/blob/master/src/PermissionRegistrar.php#L70

ctf0:
i believe it wont affect the user custom changes for the below reasons

1- any file under publishable only gets moved from the package to outside is when those files are not there in the first place, even when re-running `php artisan voyager:install`.

2- when we update the package, the user files are never touched because we are only changing those under vendor.

3- the package prioritize the files outside the vendor folder, ex.`config.voyager`

---

however there is a downside to this, whether we kept it the way it is now or changed
- atm the view files relying heavily on codes written inside the views it self, like https://github.com/the-control-group/voyager/pull/695/commits/1abeda3a204de6bf923535de9a116ac5a9fad4dd#diff-2df669a6818e34b51fd82b2a34adf375 which was cuzing issue if we commented out this line in the config.
this is a small example but u can see how bad idea it is, which means any changes in the underlying code will affect the views and now the user will go in circles trying to find out why when he updated the package some functions in the views stoped working, therefore you should consider using view files as a stateless presenter & not an extension to the controller.

abdgad:
Let's say there is a bug in one of the views, and it was fixed by a PR. If the user's views don't get updated, he will still have the bug unless he manually changes the view to fix it, which is a pain... I hope you got my point.
Voyager right now doesn't support Customizations when it comes to views, classes etc... But in the future it may well do that. But for now, we should not rush. We need to focus on fixing bugs and making small improvements until the package is stable.
After that, we will start working on making customizations easy for everyone. Maybe we'll introduce a Plugin system, so you can add custom data types and views etc... using Plugins. That's just an idea. But we shouldn't worry about it now. It will come later.

ctf0:
i understand & thats exactly what i meant, unless the user creates his own views and rely on the attributes that voyager offer, sooner or later he will fall into the trap of something goes wrong when the package gets updated.

mostly the main reason i moved the files to publishable is to offer all the views to the user to customize as he wants withouts going back and forth between `app/vendor/voyager/res/views` & the one under the `root/resources/views`.

Frondor:
Like @abdgad said, actually there is a lot of work and refactoring to do in those views, they are changing in every release, the user is already free to override them, so in my opinion there's no need to "push harder" for those overrides. I don't even suggest to override them while Voyager is still in v0.*

src/Voyager.php
marktopper: @Frondor: Wonderful idea, then there should however be a helper method to run for manually flushing it, like `php artisan voyager:flush` or `php artisan voyager:flush-permissions`.

ctf0: i believe we can just save the user perms once he logs in into session and flush it out when he leaves without the need for caching, also we wouldn't need a console command for that, as u can listen for the user logout event and flush any session data u want, example from laravel login controller
```php
// Illuminate\Foundation\Auth\AuthenticatesUsers Trait
public function logout(Request $request)
    {
        $this->guard()->logout();

        $request->session()->flush();

        $request->session()->regenerate();

        return redirect('/');
    }
```

marktopper: This would give problems when editing the permissions. So that will be needed to be handled too.

marktopper:
I rebased this to be a pull request against `v0.11`.

I will review the code and if you need I can resolve the conflicts for you.

marktopper (approved):

ctf0:
for the view file, am honestly not sure where exactly is the conflict.
for voyager, i think the one in the pr is slightly better than the one over the branch.

marktopper:
I will fix the conflicts in a commit in a few minutes.

marktopper:
@ctf0 I understand that you are confused about this.
However to explain it, the view have been split up into multiple files.

https://github.com/the-control-group/voyager/tree/release/v0.11/resources/views/formfields

publishable/database/migrations/2016_01_01_000000_add_user_avatar_and_role_id.php
```diff
@@ -6,25 +6,23 @@ class AddUserAvatar extends Migration
 {
```
marktopper: Renaming this file makes a issue with the classname, the classname must match a part of the filename.

publishable/database/migrations/2016_01_01_000000_add_user_avatar_and_role_id.php
```diff
@@ -6,25 +6,23 @@ class AddUserAvatar extends Migration
...
     /**
```
marktopper: Renaming this file makes a issue with the classname, the classname must match a part of the filename.

marktopper (changes requested):

publishable/database/migrations/2016_01_01_000000_add_user_avatar_and_role_id.php
marktopper: Might be `AddUserAvatarAndRoleId`.

marktopper: However, maybe it is not so bad to have the migrations separated like before. Since the name explains exactly what is happening.

Thoughts?

ctf0: actually as those 2 fields are mandatory for the package to work correctly, maybe we can just call it `add_user_extra_fields.php` and ignore the drop part.

also no need for the model/class creation, as the 2 columns are only needed for the interaction between the user and other models.

marktopper: `add_voyager_user_fields` maybe?

marktopper: Btw, I do not think ignoring the drop part is such a good idea? People might use it to uninstall Voyager.

ctf0: well, uninstalling packages should not touch the db in anyway, because what happens if the user did add some data ?
however if the user is going to uninstall it, it will mostly be in testing phase which 99% wont care what happens with the db tables.

marktopper: I see. 👍

marktopper: Any updates on this change? I can do it for you if you do not have time?

ctf0: am currently away from pc, can u change it from ur end ?

marktopper: Sure I can ;)

ctf0: awesome, btw if u have some time in ur hand, check https://www.youtube.com/watch?v=UxFgkCUiRAI i think it could give us some ideas about routing, comm with db, custom templating, etc...

marktopper (approved):

marktopper:
I will look into why TravisCI is failing.

Frondor:
2 things.

1. `Voyager::have` is still inside a loop, this PR may bring important performance issues.
Once again, I recommend an alternative where all the permissions are resolved once, or at least use `Cache` in that query. Mark suggestion is a good one https://github.com/the-control-group/voyager/pull/695#discussion_r100005647

2. I can't remember the 2nd thing.

src/Voyager.php
marktopper: I will make a pull request once this is merged and before releasing `v0.11` with some cache options. Will keep you posted.

marktopper:
@Frondor

> `Voyager::have` is still inside a loop, this PR may bring important performance issues.
Once again, I recommend an alternative where all the permissions are resolved once, or at least use `Cache` in that query.

I will make this in another pull request before the release of `v0.11` 👍 

> I can't remember the 2nd thing.

😂

marktopper:
@Frondor: How about adding a `voyager.cache.driver` configuration that people can change in other to use other caching.
A new cache driver can be made something like this:
```
Voyager::addCacheDriver('key', Handler::class);
```
I will make the following caching drivers available by default.
- `default`, this caching driver uses Laravel's cache system
- `none`, this is best for testing, and simple does not cache anything
And can be flushed using a Artisan method or when changes have been done.
```
php artisan voyager:flush-cache
// and maybe even this for specific cache
php artisan voyager:flush-permissions
// can be call anything else, I am open for suggestions.
```
Then when changing roles, permissions or a user role, the cache will be cleared for the related users.
And the cache key will therefor be something like this:
```
voyager.permissions.{ROLE_ID}
```

Thoughts?

Frondor:
Why we don't let the user handle the cache driver from the framework itself at `config/cache.php` ?
If he needs a different cache for voyager, I think he can create a service provider for that. And we keep the package simpler... I don't know, to be honest.

By prefixing the cache keys we can find them easily on any driver.

marktopper:
Good idea @Frondor. Keep It Simple Stupid 😄

marktopper:
I have just fixed the tests and renamed `Voyager::have` for v0.11's `Voyager::can`, `Voyager::canOrFail` and `Voyager::canOrAbort`.

PR #699, opened 2017-02-08
fix bug with adding enter to textareas in posts edit.add

dyktek:
It looks like this in CMS Panel: 
bug-textareas- 

and every update record was updated with this string


marktopper:
Thanks 👍

PR #705, opened 2017-02-09
Improve the database manager

abdgad:
Ready to merge..
Please notify me if new conflicts arise.


marktopper:
Any ETA for the finishing of this?

Cause I am considering to pull of the release of `v0.11` a few days if that could mean that this pull request could be merged into that first.

abdgad:
@marktopper the main features are ready. I just need to add handling for some edge cases, and some improvements. Also, currently only Mysql types are supported. I need to add types for other database platforms.
If this PR isn't ready for v0.11, we can merge it for v0.11.1 or something. I will fix conflicts if there are any.

marktopper:
My only concern is if this have breaking changes, then it would be nice to have them in before the actual release of `v0.11`.
However this can be merged into it without having all the special cases in it, that might come later as non-breaking changes?

Thoughts?

abdgad:
@marktopper currently the breaking changes I can think of:
- the use Vue 2 instead of Vue 1. Currently Vue 1 is used in the Media manager. We need to update it cause it's broken with Vue 2.
- Remove  DatabaseUpdate and DatabaseQueryBuilder traits: breaking change for people who are using them (which I don't think anyone is using them).
- dropping support for ENUM

That's what I can think of right now. I will look for other potential breaking changes later.
I will do my best to make this PR ready before v0.11.

marktopper:
`v0.11` will already be delayed by up to a week. So there is still time to squeeze in some breaking changes :)

marktopper:
Let me know if you need help fixing the conflicts 👍

PR #708, opened 2017-02-10
fixed browser if row type is date

AsemAlalami:

PR #715, opened 2017-02-12
Relationship fixes regarding naming conventions and displaying relationship data

sidis405:
The issue i am having has to do with setting up the relationships on BREAD according to the documentation here https://the-control-group.github.io/voyager/docs/v0.10/#core-concepts-bread-relationships .

According to this logic, if a have a User, and this user belongs to a Team model (whilst having a team_id in the user table) i would need to create a teamId() relationship.

And that does not sound quite right. I would also add that it's a bit counterintuitive as far as doing things the Laravel-way goes.

If i want to get the user's team, the relationship should be team() and not teamId().
After all i am getting back a Team instance from the relationship and not the id of that instance, which i can access with user->team->id().

So far basically i have the following implemented :

Override method name as @marktopper suggested
```
{
    "relationship": {
        "method": "team"
    }
}
```
Applying relationship data, options and labels correctly on Browse, Read, Create, Edit
Solve field name collision on ManyToMany Relationships by setting the options the following way
```
{
    "relationship": {
        "key": "sports.id",
        "label": "name",
        "method": "sports"
    }
}
```
I don't know how this worked before but it never worked for me correctly as the 'id' field was not unique


sidis405:
@marktopper Voilà

I see here the styleCI has failed and Travis is still building but perhaps if the idea is to your liking we can look at them together.

marktopper:
You can ignore StyleCI. That is automatically fixed on merge.

I will look into TravisCI.

sidis405:
As far as Style goes i made the file PSR2 compliant. Now is just complaining about white lines.
Fixed on another commit the Travis problem.

Ran test locally. They are all green.

resources/views/bread/browse.blade.php
```diff
@@ -54,10 +54,14 @@
                                                 @if ($data->{$row->field} && isset($options->relationship))
                                                     {{ $data->{$row->field}->implode($options->relationship->label, ', ') }}
                                                 @endif
+                                            @elseif($row->type == 'select_dropdown' && property_exists($options, 'options'))
+
+                                                <?php echo $options->options->{$data->{$row->field}};?>
```
marktopper: What exactly does this line do?

Also please use `{{ ... }}` instead of `<?php echo ... ; ?>`

resources/views/bread/edit-add.blade.php
```diff
@@ -164,7 +174,18 @@
...
+                                                    <?php 
+                                                        if(strpos($options->relationship->key, '.'))
+                                                        {
+                                                            list(,$key) = explode('.', $options->relationship->key);
+                                                        }else{
+                                                            $key = $options->relationship->key;
+                                                        }
+
+                                                    ?>
+                                                    {{-- {{logger(json_encode($relationshipOption->{$key}))}} --}}
```
marktopper: Please remove commented out code

resources/views/bread/edit-add.blade.php
```diff
@@ -89,22 +89,32 @@
...
+                                        <?php 
+                                            if(strpos($row->field, '_id')){
+                                                $row->field = str_replace('_id', '', $row->field); 
+                                                $hasId = true;
+                                            }else{
+                                                $row->field = snake_case($row->field);
+                                                $hasId = false;
+                                            }
+                                        ?>
+                                        <?php //$row->field = camel_case($row->field); ?>
```
marktopper: Please remove commented out code.

resources/views/bread/read.blade.php
```diff
@@ -36,6 +36,11 @@
                             @if($row->type == "image")
                                 <img style="max-width:640px"
                                      src="{{ Voyager::image($dataTypeContent->{$row->field}) }}">
+                            @elseif($row->type == 'select_dropdown' && property_exists($rowDetails, 'options') &&
+                                    !empty($rowDetails->options->{$dataTypeContent->{$row->field}})
+                            )
+
+                                <?php echo $rowDetails->options->{$dataTypeContent->{$row->field}};?>
```
marktopper, line 43: This should properly support the `linking` like the one below this comment.

src/Http/Controllers/Traits/BreadRelationshipParser.php
```diff
@@ -23,7 +25,20 @@ protected function getRelationships(DataType $dataType)
             $details = json_decode($item->details);
             if (isset($details->relationship) && isset($item->field)) {
                 $relation = $details->relationship;
-                $relationships[camel_case($item->field)] = function ($query) use ($relation) {
+                // logger($item->field);
```
marktopper: Please remove commented out code.

sidis405:
I'm away from my computer for the moment so it'll take me a bit to clean this up but basically

```
@elseif($row->type == 'select_dropdown' && property_exists($options, 'options'))
                                            <?php echo $options->options->{$data->{$row->field}};?>
```

Handles the case where the input is a dropdown but with static options.
It was not being handled in the view and getting bounced to the final @else thus just showing an integer as opposed to getting that index from the options.

I did not use
```
{{  $options->options->{$data->{$row->field}} }}
```
as the blade compiler kept erroring out. Can you confirm on your end?

Anyhow, i'll add the linking and cleanup the comments (which i thought i already did but apparently i missed some) and then commit again.

Thank you for your feedback

marktopper:
No hurry, I am also busy for the next few hours. 
Will test it out the rendering issue once I'm back along with other tests.

sidis405:
Managed to do it away from my laptop. 
For the php echo business, i found a reference that would allow curly brackets inside blade if prefixed with @ but although i haven't tested it i think results are not what we want.

The tests do pass but may be a false positive.

If this is a blade limitation i don't know how to get around it. 
Perhaps at the beginning of the file, once the json is parsed i could turn it into a collection and
access it as an array underneath to get around this.

Anyhow we'll see.

Speak soon
/best

resources/views/bread/browse.blade.php
```diff
@@ -54,10 +54,19 @@
...
+                                                @else
+                                                    @{{ $options->options->{$data->{$row->field}} }}
+                                                @endif
+    
+
                                             @elseif($row->type == 'select_dropdown' && $data->{$row->field . '_page_slug'})
                                                 <a href="{{ $data->{$row->field . '_page_slug'} }}">{{ $data->{$row->field}  }}</a>
                                             @elseif($row->type == 'date')
-                                            {{ $options && property_exists($options, 'format') ? \Carbon\Carbon::parse($data->{$row->field})->formatLocalized($options->format) : $data->{$row->field} }}
+                                            {{ $options && property_exists($options, 'format') ? \Carbon\Carbon::parse($data->{$row->field})->formatLocalized($options->format) : $dataTypeContent->{$row->field} }}
```
Frondor, line 69: @marktopper @sidis405 ~I can't remember right now, but although I think this is correct, what happened here #708 then?~
Nevermind, just checked. It should be `$data` instead of `$dataTypeContent`. Please, fix it.

sidis405: I was rather confused why only in this file it was called dataTypeContent and not data.
Anyhow i'll clean it up on the next commit.

marktopper (approved):

marktopper:
All tested and works good.
However there seems to have been added some conflicts.
I can look into it if you do not have time for it.

NOTE TO SELF: This should be documented.

sidis405:
@marktopper Ah that's great. 
Well later i got to freshen up the browse.blade.php as discussed with @Frondor and move the $dataTypeContent to $data have a quick look at that curly braces inside curly braces business that is nagging at me :D

That way i can fix the conflicts and push again.

Sometime in the next few hours.

marktopper:
Cool, ping me when ready ;)

sidis405:
@marktopper Turns out i got to go away for work for a couple of days and won't be able to look at the conflict before i come back.
Is there a change you could look into it? If not i'll be happy to do it when i come back.
Whatever is good for you.

Best
sid

marktopper:
I will look into it, enjoy your days off.

SkyL0ver:
Hi.. I have tested the id to name in a fresh laravel 5.3 installation and voyager 0.10.15 but seems still not work.  the id still show id not name or description at browse section.

Frondor:
It works, do `php artisan cache:clear` | `php artisan view:clear` | `php artisan route:clear` | `php artisan config:clear`

bjornfant:
Hi guys. Great that this is fixed. I'm a bit new to Voyager and don't find any documentation on how to use this? I really would like to show a label over an id in the browse view of my system.

fletch3555:
@bjornfant, this PR was merged in 7 months ago.  If you have questions or issues, please make your own issue or ask on our Slack group

chadpriddle:
@bjornfant I'm looking for the same solution. I looked through the commits and merge requests and cannot find anything. Documentation seems lacking if this was fixed.

fletch3555:
@chadpriddle, please don't make me repeat myself.....

PR #697, opened 2017-02-07
BREAD JSON editor & new responsive design for login page

Frondor:
## Contents

- Fix [638#issuecomment-277334908](https://github.com/the-control-group/voyager/pull/638#issuecomment-277334908)
- **Add JSON editor to Database tool.**

- **Login page re-design (way more sexier with material design)**

(actually that's not the default gradient :laughing:, gradient colors can be changed from `config('voyager.login.gradient_a')` or `... gradient_b` )

- Rich text area now shows contents with HTML stripped tags
- Multiple improvements to dashboard UI for small screens UX.
- Dashboard index widgets are now responsive.
- Tables and DataTables are now responsive!!! (even a plugin for [client-side paginated tables](https://i.gyazo.com/fcd420779eb13be7e47d69911353c4f9.gif) can be enabled from voyager's config file)
- Logout is now secured in a POST form.
- Fixed many issues regarding to BREAD `relationship` fields default options
- Fixed `logo` setting. It now displays properly.
- Fixed test case where it was expecting for a fixed value. It now expects the value in your app locale `trans('auth.failed')`
- Moar issue fixes...

PR #717, opened 2017-02-12
Fix #716

### LINKED ISSUE #716: Parent id category warning
- Laravel Version: 5.3.30
- Voyager Version: 0.10.13
- PHP Version: 7.0.14
- Database Driver & Version: mysql 5.5.53
  
Parent id category warning

When I go to edit a category I have this warning

Now I change in models category 
`public function parent_id()`
to
`public function parentId()`

but selected value when I edit category is always NULL then I change this in blade template 
Line 104 `$selected_value = old($row->field);`
to 
`$selected_value = $dataTypeContent->parent_id;`

And now works this is an issue or I'm doing it wrong?

marktopper:


marktopper:
Reference #716

turtuga:
Thanks :) this works now only problem is selectbox doesn't select category when edit second time I think the problem is there $selected_value = old($row->field); in blade template

Frondor:
Sorry I hadn't enough time to make the change log for my actual PR. This was also addressed here https://github.com/the-control-group/voyager/pull/697/files#diff-db973ad26caa434a9afab15a81a3c9bb
But I think it won't generate any conflict, will it?

@turtuga I've addressed some of those glitches, but I'm not sure about what you say right now. Can you edit your issue, or wait until #697 gets merged? At the moment, my fork is working as intended.

turtuga:
@Frondor I will wait but in views/bread/edit.add.blade.php exists an issue I think take this example

Edit category 1 and select parent id category 2 in database change correctly but if you edit again category 1 the parent id selectbox as selected NULL option not category 2.

@marktopper and @Frondor thanks :)

Frondor:
I can confirm that's a bug indeed. I'll see if I can address it later today, if someone else didn't do it already.

turtuga:
@Frondor ok thanks for your time

Frondor:
Just sent a PR for this, @marktopper.
#719 may need to be merged along with this PR.

marktopper:
Merging in both and doing a new release.

Thanks @Frondor 👍

PR #720, opened 2017-02-12
Edit and redirect to the same editable content

turtuga:
I think is more usefull edit and redirect no edit page not to index is it valid?


Frondor:
Although this may be handy for some people, it shouldn't be implemented like that.
We'll have ajax forms in the future, then you'll edit and manually exit the page, but at the moment it should stay as it is, imo.

marktopper:
I agree with @Frondor.
The correct way to resolve this is to save the data with AJAX calls instead.

PR #721, opened 2017-02-12
tinyMCE update and minor fix

turtuga:
- Update tinyMCE to 4.5.3 
- Fix fullscreen modal giphy z-index
- Add table plugin to tinyMCE I think is usefull


marktopper:
Thanks!

hazem1taha:
The current editor doesn't contain the ability to change text color, does it?

marktopper:
Seems like you are right.
It is not configured to show that ability.

We should indeed add this, maybe as a configuration of editor methods.

PR #726, opened 2017-02-13
Refactor/form fields

marktopper:
### Add form field
```
Voyager::addFormField(DateHandler::class);
```
### Handler class
```
class DateHandler extends AbstractHandler
{
    protected $name = 'Date Input';
    protected $codname = 'date';

    public function createContent($row, $dateType, $dataTypeContent, $options)
    {
        return view('voyager::bread.formfields.date', [
            'key' => 'value',
        ]);
    }
}
```
### Show form field
```
{!! Voyager::formField($row, $dataType, $dataTypeContent) !!}
```

Thoughts?

/ping @Frondor


Frondor:
:soccer: Goal! :ok_hand:

marktopper:
I will merge this in now then 😄

resources/views/bread/edit-add.blade.php
```diff
@@ -63,11 +63,11 @@
                                 <div class="form-group">
```
Frondor, line 63: Field views should start at `<div class="form-group">`. Actually the loop is inside a single `.form-group`, which is not right, it should be outside and each item should be wrapped by that class.

src/Voyager.php
```diff
@@ -19,16 +21,64 @@ class Voyager
...
+        });
+    }
+
+    public function addFormField($handler)
+    {
+        if (!$handler instanceof HandlerInterface) {
+            $handler = app($handler);
+        }
+
+        $this->formFields[$handler->getCodename()] = $handler;
```
Frondor, line 56: `return $this` maybe? So you can `addFormField()->addAfterFormField()->addAfterFormField()`

Frondor:
@marktopper ups.. a bit late :laughing:

resources/views/bread/edit-add.blade.php
```diff
@@ -63,11 +63,11 @@
...
                                     <label for="name">{{ $row->display_name }}</label>
 
-                                    @includeIf("voyager::formfields.$row->type")
+                                    {!! app('voyager')->formField($row, $dataType, $dataTypeContent) !!}
```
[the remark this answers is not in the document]
marktopper: We could use facade yea. Not sure what is best with performance.
Basicly it does the same.

src/Voyager.php
marktopper: Good idea, will make a commit for this.

marktopper:
It's never to late to do reviews 😄

resources/views/bread/edit-add.blade.php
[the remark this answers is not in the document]
Frondor: Let's just focus on the feature then, this can be discussed later

resources/views/bread/edit-add.blade.php
marktopper: Your right, thanks for pointing that out. Moving out.

Frondor:
I can't find the logic to automatically add input's descriptions (`addAfterFormField`) by default, when its present in `$options` JSON object... Are you already handling that?

marktopper:
@Frondor: [This](https://github.com/the-control-group/voyager/pull/726/files#diff-36bd69ef7ba5df2912a70baf254a067bR69) triggers the [description](https://github.com/the-control-group/voyager/pull/726/files#diff-fd262e1f9e0149be0ab2e8a863786e65).

Issue #674, opened 2017-02-04
SQL error when trying to enter data types with timestamps


redsoxfan2499:
- Laravel Version: 5.3.30
- Voyager Version: 0.10.10
- PHP Version: 5.6.24
- Database Driver & Version:
mysql
### Description:
following tutorials on youtube. created a products data type with timestamps. but when I try to enter a new product. I leave the created update fields blank since I do not get a pop type bootstrap calendar and I assume that the timestamps will be saves automatically but getting SQL error:

SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '' for column 'created_at' at row 1 (SQL: insert into `products` (`name`, `description`, `created_at`, `updated_at`) values (Laravel Book, A book about Laravel development, , ))

if we have to enter a date for data types then I would assume a jquery calendar or some date formatting hints.  But I then try to enter a data as such:  02/02/2017 and get an error of:
Unexpected data found.
Unexpected data found.
Data missing

### Steps To Reproduce:
following tutorials on youtube. created a products data type with timestamps. but when I try to enter a new product. I leave the created update fields blank since I do not get a pop type bootstrap calendar and I assume that the timestamps will be saves automatically but getting SQL error:

BTW. in the database tool. the timestamps options are marked YES for NULL.

redsoxfan2499:
So today more research. So when I have the created and updated fields marked as Date fields in the BREAD editor I do get a calendar view but of course in the DB it is marked as Timestamps so this does not work. I can leave the fields as Textboxes but I have to make sure that the time is formatted correctly as such:  2017-02-24 00:00:00	If I leave them blank, I get the SQL error. The field is not formatted correctly I get SQL error.  

So I would assume that if I chose the Created and Update fields in my datatype, I would not need to enter anything in the fields and upon submit, it would just save the item in the DB. As of right now, you have to enter a Timestamp formatted correctly or you will get an error. 

Thanks!

marktopper:
Can you please share a screenshot of the database table `products` structure, the BREAD options for `products` and the error that happens.

redsoxfan2499:
so for this example, I enter a name and price and leave the created and updated fields blank since they are timestamps in the database.
screenshots below:
![voy1](https://cloud.githubusercontent.com/assets/1205420/22627745/9fbdbfc2-eb8d-11e6-9c60-168bcb62e9f7.PNG)
[...]

PR #729, opened 2017-02-13
#674 Check for successful timstamp conversion

oschettler:


src/Http/Controllers/Controller.php
```diff
@@ -225,7 +225,11 @@ function (Constraint $constraint) {
             /********** TIMESTAMP TYPE **********/
             case 'timestamp':
                 if ($request->isMethod('PUT')) {
-                    $content = gmdate('Y-m-d H:i:s', strtotime($request->input($row->field)));
+                    if (null === $request->input($row->field)) {
```
marktopper: Please check for empty instead, since we also support Laravel 5.3, which will have this as a empty string instead.
The `empty` function will catch both empty string and `null`.

marktopper (changes requested):

marktopper (approved):

marktopper:
Thanks! 👍

PR #740, opened 2017-02-15
Add date-time type to BREAD table

AsemAlalami:
Add date-time Type to BREAD


marktopper:
What is the different from this and `Timestamp`?

AsemAlalami:
It don't have Timestamp, and when edit-add can add time

marktopper:
Okay thanks. Merging in!

PR #742, opened 2017-02-15
Please don't use short tags

gnovaro:
http://softwareengineering.stackexchange.com/questions/151661/is-it-bad-practice-to-use-tag-in-php

### DIFF
--- a/resources/views/users/read.blade.php
+++ b/resources/views/users/read.blade.php
@@ -13,7 +13,6 @@
 
                 <div class="panel panel-bordered" style="padding-bottom:5px;">
 
-
                     <!-- /.box-header -->
                     <!-- form start -->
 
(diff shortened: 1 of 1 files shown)


marktopper:
I think it would be better to use `{!! $variable !!}` instead.

gnovaro:
Is the same. I prefer php but is a choise.

Frondor:
It's not the same, it's the laravel way of writting unescaped content https://github.com/laravel/framework/blob/5.4/src/Illuminate/View/Compilers/Concerns/CompilesEchos.php#L42
Personally, I prefer blade syntax over pure php on blade views for obvious reasons. Besides that, blade prepares raw content, unlike simply `<?php echo ... ?>`.
This is a small improvement, but can be turned into a serious coding style problem in future versions.

gnovaro:
@Frondor is and admin, and for this case in particular is a content field. maybe the @marktopper solution is better...

Frondor:
That's what I meant.

PR #743, opened 2017-02-15
Fix php short tag

gnovaro:


marktopper:
I have changed this to use blade syntax.

PR #744, opened 2017-02-15
Replace short tag

gnovaro:


Frondor:
Quoting myself:
> It's not the same, it's the laravel way of writting unescaped content https://github.com/laravel/framework/blob/5.4/src/Illuminate/View/Compilers/Concerns/CompilesEchos.php#L42
Personally, I prefer blade syntax over pure php on blade views for obvious reasons. Besides that, blade prepares raw content, unlike simply `<?php echo ... ?>`.
This is a small improvement, but can be turned into a serious coding style problem in future versions.

gnovaro:
@Frondor try with the blade syntax, I can not try it right now, but maybe try to escape something that should not.

marktopper:
The blade syntax `{!! ... !!}` does not escape anything.
`{{ ... }}` escapes.

I have changed this to use blade syntax.

PR #747, opened 2017-02-15
[V0.11] BREAD JSON editor & new responsive design for login page

Frondor:
#697


Frondor:
There is a weird thing going on with `$options` variable for `text` fields... I can't see the error.
I've removed the line that wasn't meant to be there at `edit-add.blade.php`, because `TextHandler` is supposing to be using that variable already... But I can't realize where it is failing... :disappointed:

marktopper:
I will take a look at this.

marktopper:
I think this error was my fault.
I have however fixed this and sent a commit.

Will merge this once test is green. (might be once it's done testing)

PR #749, opened 2017-02-16
Fix small type at login failure

denbatte:


marktopper:
Please also update the test for this: https://github.com/the-control-group/voyager/blob/master/tests/LoginTest.php#L30

marktopper (changes requested):
Test must be changed according to this

ctf0:
all error msgs should be saved in a lang file instead of hardcoding it

Frondor:
Yeah, @ctf0 is right. I wrote `trans('auth.failed'')` because it should support the current `app.locale`.
That error message shouldn't be hardcoded anymore.

If I remember well, this is fine on v0.11

fletch3555:
@denbatte, @Frondor,

Should this be closed?  No activity in 2 months, and I'm not quite sure what the status is

Frondor:
@fletch3555 indeed :+1:

PR #757, opened 2017-02-17
Rendering issue

sidis405:
fixed rendering issue on browse bread #715 

This is a previous pull request of mine that apparently had a bug.

screen shot 2017-02-17 at 15 44 55

Minor tweak. It is now fixed.

screen shot 2017-02-17 at 15 48 22

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -57,9 +57,9 @@
                                             @elseif($row->type == 'select_dropdown' && property_exists($options, 'options'))
 
                                                 @if($data->{$row->field . '_page_slug'})
-                                                    <a href="{{ $data->{$row->field . '_page_slug'} }}">@{{ $options->options->{$data->{$row->field}} }}</a>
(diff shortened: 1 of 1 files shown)


sidis405:
@marktopper i noticed this today after updating to v0.10.15.
it was the matter regarding the usage of php tags in the blade view as the compiler had issues with other curly braces in the expression.

abdgad:
I haven't tested this. But why are you using `{!! !!}` instead of `{{ }}`?

sidis405:
@abdgad Because blade errors out parsing the following

```
{{ $options->options->{$data->{$row->field}} }}
```

abdgad:
I think it's better to refactor and not use `options->options->{$data->{$row->field}}` because it's hard to read and can cause security issues.

sidis405:
I agree there is quite a bit of refactoring needed in those blade files and even generally.
This pull is meant to fix an actual *breaking issue*, but if you have any ideas how to refactor that expression without starting to dig at the core of the package then i'd be happy to help. :)

marktopper:
Thanks

PR #758, opened 2017-02-17
Allow model overwriting

marktopper:
Enable overwriting models like this:

```
Voyager::useModel('user', MyUser::class);
```

Voyager will then use that user class as it's class for any Voyager specific code.


derekbliss:
I am having a tough time figuring out how to add relationships to the Post model without modifying the vendor Post model. Could I use this and how would I use this?

marktopper:
@derekbliss: Make a new model extending the original one and use that for Voyager.
Then you may add the relations you need.

Also you have to edit the model set for the BREAD.

derekbliss:
@marktopper oh wow I just noticed in the BREAD there is a model name field.  I just changed that to App\Post and added my methods to the new Post model. This seems to work.

Thank you

PR #759, opened 2017-02-17
Drop columns only if exists

turtuga:
To avoid errors when try delete column not exists in database.


abdgad:
`src/Http/Controllers/Traits/DatabaseUpdate.php` has been removed in #705

turtuga:
OK sorry :)

PR #783, opened 2017-02-19
Minor ui fix

turtuga:
Thinks I have changed:
- Align center checkbox in table editor
- Table head bg color
- Add class form-control edit bread selectbox
- Remove bootstrapSwitch I think is not used
- Add data-toggle="toggle" to inputs in database manager 
help
- Align font icons to text


abdgad:
Note that this causes a bug, the check boxes no longer work. This is why we shouldn't make changes to the db manager UI right now, cause it can be easily broken.
When the DB manager is stable and has all the features that are needed, then we'll see how to improve its look = )

turtuga:
Can you please make a gif of the bug??

abdgad:
Change one of the check boxes, then update the table.
The change won't take effect. That is because we need to listen for events when the check box is clicked, and react accordingly. Which means we'll add more code just to handle a style thing...
Please let's not worry about these UI issues for now. Cause I will be adding more code to handle database logic, and add functionality features, which is the main point of the DB manager. We'll add code to add styling and UI improvements once all the logic code is implemented.
However, feel free to add and test out any UI improvements you like, I will see what can be merged now and what can be delayed for later :)

turtuga:
ok :D thanks I will talk to @marktopper to close this PR sorry for the confusion

PR #787, opened 2017-02-19
tinymce add forecolor and backcolor

turtuga:
Add plugin textcolor and buttons forecolor and backcolor in editor


marktopper:
Thanks!

PR #788, opened 2017-02-19
v0.11 enhancements

ctf0:
- replace auth:: facade with the function as its already cached and dont need to be re-init on each call
- fixed profile image in user profile page when its a url rather than an image path
- now widgets icon&image follow the value in the config
- add bio to user columns to avoid getting error in profile view
- found solution for user search by uuid instead of id ⚠️  `needs to be added to docs`
- replaced png images with svg & updated views with the new svg image
- removed the sketch file as its not part of the build release
- compressed widgets images to save bandwidth & demo_content images as well
- removed the `?` under db table creation as its not needed
- changed `AI` to `AutoIncrement` for better understanding
- add `form-control` class to drop downs
- add more tables to ignore under config
- add an option to ignore files from media manager ex.`.git` through regex
- remove `nullable` from `role_id` as the whole thing wont work without it
- better check for the user avatar link & fixed a couple of views wasnt showing the avatar


publishable/config/voyager.php
```diff
@@ -94,21 +95,19 @@
...
 
     'widgets' => [
-        [
-            'name'  => 'User',
+        'user' => [
             'icon'  => 'voyager-group',
             'model' => TCG\Voyager\Models\User::class,
             'url'   => 'admin/users',
             'image' => '/images/widget-backgrounds/02.png',
         ],
```
marktopper, line 133: `v0.11` no longer uses this, but uses classes like [this](https://github.com/the-control-group/voyager/blob/master/publishable/config/voyager.php#L117).

Maybe instead for specific configuration for the user, post and page dimmers, you could do something like this:
```
'dimmers' => [
    'user' => [/* ... */],
    'post' => [/* ... */],
    'page' => [/* ... */],
],
```

publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
```diff
@@ -10,8 +10,9 @@ class AddVoyagerUserFields extends Migration
     public function up()
     {
         Schema::table('users', function ($table) {
-            $table->string('avatar')->nullable()->after('email');
             $table->integer('role_id')->nullable()->after('id');
+            $table->string('avatar')->nullable()->after('email');
+            $table->mediumText('bio')->nullable()->after('avatar');
```
marktopper: The new `bio` field should be turned into a new migration since `v0.11` already is out.

resources/views/profile.blade.php
```diff
@@ -14,12 +14,12 @@
...
              style="border-radius:50%; width:150px; height:150px; border:5px solid #fff;"
-             alt="{{ Auth::user()->name }} avatar">
-        <h4>{{ ucwords(Auth::user()->name) }}</h4>
-        <div class="user-email text-muted">{{ ucwords(Auth::user()->email) }}</div>
-        <p>{{ Auth::user()->bio }}</p>
-        <a href="{{ route('voyager.users.edit', Auth::user()->id) }}" class="btn btn-primary">Edit My Profile</a>
+             alt="{{ auth()->user()->name }} avatar">
+        <h4>{{ ucwords(auth()->user()->name) }}</h4>
+        <div class="user-email text-muted">{{ ucwords(auth()->user()->email) }}</div>
+        <p>{{ auth()->user()->bio }}</p>
```
marktopper: Since `v0.11` already is released, please do
```
@if (isset(auth()->user()->bio))
    <p>{{ auth()->user()->bio }}</p>
@endif
```

marktopper (changes requested):

resources/views/profile.blade.php
ctf0: no need for `isset` as the column will be avail in the db, so if no value it will empty

publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
ctf0: it will create extra boilerplate which i was hoping to avoid 😢

publishable/config/voyager.php
ctf0: yeah i saw the new changes, am not sure whats ur plan for widgets but having only the classes in the config is not right, u may even just remove it all together and enforce it through the class, but what if the user wants to make some changes ?

also widgets is easier to understand than dimmer, maybe instead of changing the name we can have a slight more detailed description about what they are for.

publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
marktopper: Nah, keep it this way. It's okay 👍

publishable/config/voyager.php
marktopper: I agree with you there. Maybe we should remake the entire configuration for the `widgets` group instead.
Thoughts?

resources/views/profile.blade.php
marktopper: 👍

publishable/config/voyager.php
marktopper: How would this configuration look ideally.

ctf0: i think the current one looks good & easy to understand
```php
'user' => [ // the name of the widget
     'icon'  => 'voyager-group', // the icon you want this widget to have  
     'model' => TCG\Voyager\Models\User::class, // the class responsible for the widget logic
     'url'   => 'admin/users', // where this widget should re-direct to when clicked
     'image' => '/images/widget-backgrounds/02.png', // background for the widget box
 ],
```

the good thing about this is, if the user don't want any of them he can simply remove/comment it, or if he wants to add a new one he just duplicate and start adding his config,
even the model is not enforced as he just needs to add the full namespace of the class and it will be picked up.

i understand that u want to keep this as minimal as possible but over simplifying things will lead to the exact opposite.

marktopper: Problem with that is that the view for it can't be customized for each widget. Which it can now using the class handlers.

marktopper: I am still open for further ideas for this :)

ctf0: i thought that the model is the one responsible for declaring the view and what logic gets sent to, so even with the new approach , nothing changed except its now been enforced on the user.

however if u absolutely want to go with the new approach, then the user can simply extend the class and make his changes (same with controller, same with models)

marktopper: I will try figure out some other ways to do this in the future and then make a discussion for it.

fletch3555:
In the future, please limit your pull requests to one change at a time.  If we didn't like how one of your 10+ changes were done, we'd have you ask you to undo it manually instead of simply declining that one pull request.  Thanks!

ctf0:
@marktopper do u still need this PR ?

marktopper:
@ctf0 It may be closed

PR #795, opened 2017-02-20
Fix Mysql TIMESTAMP type not accepting NULL as a default value

### LINKED ISSUE #785: issues under v0.11
- Laravel Version: 5.4.12
- Voyager Version: 0.11
- PHP Version: 7.0.15
- Database Driver & Version: 5.7.17

### Description:
1- opening media manager show error in console & it can make clicking the top right avatar un-responding.
screen shot 2017-02-19 at 9 05 14 pm

2- expanding the side panel doesnt persist when navigating away.

3- nested menu with more than one sibling, you cant move the first item unless u started from bottom , to reproduce 
- go to `menu builder`
- click on builder next to `admin menu`
- scroll to item named `more` and try to un-nest the first item "it wont work"
    - now try to un-nest the second item "it will work"

4- cant create timestamp columns ex.`created_at & updated_at` because it needs a default value.
screen shot 2017-02-20 at 3 44 41 pm

5- `enum` field type is not-supported but its avail under `posts > status`, so how we can have the same functionality for other tables ?

abdgad:
Note: I will create a new PR to also handle CURRENT_TIMESTAMP issue.

### DIFF
--- a/src/Database/Types/Mysql/TimestampType.php
+++ b/src/Database/Types/Mysql/TimestampType.php
@@ -11,6 +11,8 @@ class TimestampType extends Type
 
     public function getSQLDeclaration(array $field, AbstractPlatform $platform)
     {
-        return 'timestamp';
+        $null = isset($field['default']) ? '' : ' null';
+
+        return 'timestamp'.$null;
     }
 }


src/Database/Types/Mysql/TimestampType.php
```diff
@@ -11,6 +11,8 @@ class TimestampType extends Type
 
     public function getSQLDeclaration(array $field, AbstractPlatform $platform)
     {
-        return 'timestamp';
+        $null = isset($field['default']) ? '' : ' null';
```
ctf0, line 14: i think it should be `'NULL'`

abdgad: It's case insensitive. But all types are using lower case. It's just a matter of consistency.

ctf0:
@abdgad why u closed it ?

PR #806, opened 2017-02-21
Make app-footer view

marktopper:


hazem1taha:
@marktopper 
Why the file extension is not .blade ? it is shown in the footer as a plain text.

resources/views/partials/app-footer.php
```diff
@@ -0,0 +1,13 @@
+<footer class="app-footer">
+    <div class="site-footer-right">
+        @if (rand(1,100) == 100)
+            <i class="voyager-rum-1"></i> Made with rum and even more rum
+        @else
+            Made with <i class="voyager-heart"></i> by <a href="http://thecontrolgroup.com" target="_blank">The Control Group</a>
+        @endif
+        @php $version = Voyager::getVersion(); @endif
```
hazem1taha, line 8: This `@endif` should be `@endphp`

marktopper: Fixed at [commit (987b09cf219639caeb7f3ecf718ffb44eddf2d51)](https://github.com/the-control-group/voyager/commit/987b09cf219639caeb7f3ecf718ffb44eddf2d51).

PR #836, opened 2017-02-27
Added Save shortcut to bread edit-add page

ghost:
I've added Save shortcut (Ctrl+S or Cmd+S on Mac) to bread edit-add page. I think it's pretty useful feature to save changes with a keyboard shortcut.

### DIFF
--- a/src/Http/Controllers/VoyagerBreadController.php
+++ b/src/Http/Controllers/VoyagerBreadController.php
@@ -156,7 +156,7 @@ class VoyagerBreadController extends Controller
         $this->insertUpdateData($request, $slug, $dataType->editRows, $data);
 
         return redirect()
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/VoyagerBreadController.php
```diff
@@ -156,7 +156,7 @@ public function update(Request $request, $id)
         $this->insertUpdateData($request, $slug, $dataType->editRows, $data);
 
         return redirect()
-            ->route("voyager.{$dataType->slug}.index")
+            ->route("voyager.{$dataType->slug}.edit", ['id' => $id])
```
fergthh, line 159: This break the tests. On the other hand, this change has been discussed earlier in #720.

PR #848, opened 2017-03-02
Added active() scope to Page model

ghost:
Added active() scope to Page model which returns only active pages, example:

`$pages = Page::active()->get();`


marktopper:
Great work, thanks 👍

PR #860, opened 2017-03-02
Error happens when changing public path for "public" to "public_html"

psmedia90:
when changing public path and update the profile image you will get a Local.php Error on Line: 345

changing the code above fixed that error


marktopper:
Thanks!

PR #870, opened 2017-03-04
Post and Page Display

duxor:
Display Post and Page content via functions post($slug, $id, $author)
add page($slug, $id, $author)


turtuga:
Why not use slug ?? Because slug is unique..

Frondor:
We really appreciate your work, but as far as I know, the whole post module is going to be moved into a `hook` in an upcoming version, and then it may expose its own api using its facade for displaying matters.
I'll close this for now, probably @marktopper wants to make use of it in the future.

Thank you once again.

Issue #835, opened 2017-02-27
Category always return to the first category


ergxpr0xy:
- Laravel Version: 5.4.13
- Voyager Version: 0.11.3
- PHP Version: 
- Database Driver & Version:

### Description:
Category always return to the first category (category_id 1)
![2017-02-27 3](https://cloud.githubusercontent.com/assets/19571567/23362351/f8ffebf4-fd27-11e6-8c5b-e9be0162e174.png)
![2017-02-27 1](https://cloud.githubusercontent.com/assets/19571567/23362352/f902fd30-fd27-11e6-8d08-079da4c31676.png)
![2017-02-27 2](https://cloud.githubusercontent.com/assets/19571567/23362353/f904408c-fd27-11e6-8c23-43cd94a80c3c.png)

dimsanchuk:
![image](https://cloud.githubusercontent.com/assets/13986110/23436282/9ee647d0-fe12-11e6-9664-cf810412099b.png)
Field "category_id" must be "Select Dropdown".

ergxpr0xy:
turns out the default input type for it was checkbox, i hope it gets fixed in the next update. thank you!

waelwalid:
1 - Edit bread of post
2 - convert it to Dropdown menu

It works with me

marktopper:
Correct @waelwalid.
However I have left this open since it's still not the default value in the seeders.

I'm open for pull requests optimizing this.

fletch3555:
I'm closing this since it's 4.5 months old now.  We can revisit these issues once 1.0 gets released.

PR #875, opened 2017-03-06
add cagetory id to post  datarow

aronpc:
Fix category id in DataRowTableSeeder

### DIFF
--- a/publishable/database/seeds/DataRowsTableSeeder.php
+++ b/publishable/database/seeds/DataRowsTableSeeder.php
@@ -48,6 +48,21 @@ class DataRowsTableSeeder extends Seeder
             ])->save();
         }
 
+        $dataRow = $this->dataRow($postDataType, 'category_id');
+        if (!$dataRow->exists) {
+            $dataRow->fill([
+                'type'         => 'text',
+                'display_name' => 'Category ID',
+                'required'     => 1,
+                'browse'       => 1,
+                'read'         => 1,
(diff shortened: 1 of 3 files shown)


dimsanchuk:
Why type is "text"? If you are fixing this issue ([#835](https://github.com/the-control-group/voyager/issues/835)), I think type should be "select_dropdown".

publishable/database/seeds/DataRowsTableSeeder.php
```diff
@@ -48,6 +48,21 @@ public function run()
             ])->save();
         }
 
+        $dataRow = $this->dataRow($postDataType, 'category_id');
+        if (!$dataRow->exists) {
+            $dataRow->fill([
+                'type'         => 'text',
```
marktopper, line 54: Please use the `select_dropdown` type, and add the `relationship` JSON option.

marktopper (changes requested):

fletch3555 (commented):
@marktopper, aren't we trying to phase-out the Pages/Posts/Categories stuff...? If so, I don't want these changes in there

Also, @aronpc, these aren't related to the original PR.  Please use separate git branches so you don't accidentally mess up an existing PR while working on something else.

aronpc:
> Also, @aronpc, these aren't related to the original PR. Please use separate git branches so you don't accidentally mess up an existing PR while working on something else.

Thanks

PR #904, opened 2017-03-12
Fix width select 2

turtuga:
Fix select 2 width I don't know why but last changes introduce this bug

select 2


marktopper:
Thanks

PR #908, opened 2017-03-12
scroll on menu more beautifull

turtuga:
Only minor fix to make it more beautiful


fletch3555:
Assigning @thedevdojo to review style changes.

PR #911, opened 2017-03-12
Edit user class

turtuga:
What you think this approach to normal user class not edit your rule id and not edit admin settings?

### DIFF
--- a/resources/views/users/edit-add.blade.php
+++ b/resources/views/users/edit-add.blade.php
@@ -79,7 +79,7 @@
                                 <input type="file" name="avatar">
                             </div>
 
-                            @if (Auth::user()->role_id == "1")
+                            @if (Auth::user()->role_id == 1)
                             <div class="form-group">
                                 <label for="role">User Role</label>
                                 <select name="role_id" id="role" class="form-control">
--- a/resources/views/users/edit-add.blade.php
+++ b/resources/views/users/edit-add.blade.php
@@ -79,7 +79,7 @@
                                 <input type="file" name="avatar">
                             </div>
 
-                            @if (Auth::user()->role_id == 1)
+                            @if (Auth::user()->role_id == TCG\Voyager\Models\Role::where('name', '=', 'admin')->first()->id)
                             <div class="form-group">
                                 <label for="role">User Role</label>
                                 <select name="role_id" id="role" class="form-control">
@@ -111,4 +111,14 @@
             </div>
         </div>
     </div>
+@stop
+
+@section('javascript')
+    <script>
+        $('document').ready(function () {
+            $('.toggleswitch').bootstrapToggle();
+        });
+    </script>
+    <script src="{{ config('voyager.assets_path') }}/lib/js/tinymce/tinymce.min.js"></script>
+    <script src="{{ config('voyager.assets_path') }}/js/voyager_tinymce.js"></script>
 @stop
--- a/resources/views/users/edit-add.blade.php
+++ b/resources/views/users/edit-add.blade.php
@@ -112,13 +112,3 @@
         </div>
     </div>
 @stop
-
-@section('javascript')
-    <script>
-        $('document').ready(function () {
-            $('.toggleswitch').bootstrapToggle();
-        });
-    </script>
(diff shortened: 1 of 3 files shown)


resources/views/users/edit-add.blade.php
```diff
@@ -79,6 +79,7 @@
                                 <input type="file" name="avatar">
                             </div>
 
+                            @if (Auth::user()->role_id == "1")
```
fletch3555: I don't really like using the hard-coded role ID.  Also, I believe the role_id **should** be numerical, not a string.

resources/views/users/edit-add.blade.php
```diff
@@ -112,13 +112,3 @@
         </div>
     </div>
 @stop
-
```
fletch3555, line 115: Please explain why this was removed.  It doesn't seem related to the rest of the changes.

src/Http/Controllers/VoyagerUserController.php
```diff
@@ -0,0 +1,98 @@
...
+        // Check permission
+        Voyager::canOrFail('edit_'.$dataType->name);
+
+        $relationships = $this->getRelationships($dataType);
+
+        $dataTypeContent = (strlen($dataType->model_name) != 0)
+            ? app($dataType->model_name)->with($relationships)->findOrFail($id)
+            : DB::table($dataType->name)->where('id', $id)->first(); // If Model doest exist, get data from table name
+
+        if (Auth::user()->role_id == 2 && $dataTypeContent->role_id == 1) {
```
fletch3555: Again, not too fond of using hard-coded role IDs.

fletch3555:
Please explain what this PR is supposed to fix.  I can't tell from the changes alone.

resources/views/users/edit-add.blade.php
turtuga: because in add or edit user we not use tinymce or bootstraptoggle

turtuga: because in add or edit user we not use tinymce or bootstraptoggle

turtuga:
If I login with "user" in "my profile" I can change my role to administrator this is not supposed occur right ??

resources/views/users/edit-add.blade.php
fletch3555: Fair enough.  Just wanted to make sure it was a conscious change.

marktopper:
In this case a user would **only** be able to change the user role if they are `admin`.

I think that it would be better to make some kind of permission for this.

turtuga:
what do you suggest? Create one role in users section "Allow edit user class" is one option?

marktopper:
That is my idea. But hold on, will quickly take this up on the chat with the other maintainers.

abdgad:
@turtuga if you login with `user`, what happens if you go here: `/admin/users/1/edit`? Are you able to edit the Administrator user?

turtuga:
I have validated `user` to not edit, but is only my opinion...

abdgad:
@turtuga but without your modifications, you can edit other users with a normal user, correct?

turtuga:
yes, and with normal user we can edit user class and I think it's not correct, because normal user can change user class to administrator

marktopper:
Sorry for the late response. I have been chatting with @fletch3555 about how to solve this the proper way, and we came up with a solution for this.

It could be a option to make it possible when editing the BREAD details, to define that a certain BREAD row requires additionally permissions.
For example lets say the for the `users` table, then the permission for editing the `role_id` would have the `users_edit_role_id` permission key.

@turtuga and @abdgad, what are your thoughts on this?

turtuga:
@marktopper yeah I like it. but we have only one little problem I think.. Imagine that one user can edit user class but he not must edit user class to administrator or class above he.

marktopper:
/ping @fletch3555 for the proper solution for this.

fletch3555:
I'm not sure we can officially support that use-case at this time.  For that to work, the Roles would have to be stored as a hierarchy (otherwise there's no way to know what role is "higher").  @turtuga, you're welcome to make it work yourself if it's what you need, but for now, that specific use-case will have to be put on hold until we have a chance to revamp the roles/permissions system.

turtuga:
All right :)

PR #924, opened 2017-03-13
Select 2 like others inputs

turtuga:
I think looks better what you think ?


tnylea:
@turtuga Looks nice. Thanks.

PR #925, opened 2017-03-13
Add prefilled id column when creating a new table

abdgad:
Since most tables will have an `id` column, now when creating a new table, it will have one prefilled to avoid adding it manually.


turtuga:
that's a good one :D thanks 👍

PR #945, opened 2017-03-15
Use laravel url() for paths

bhartvigsen:
I tested out Voyager on an app I'm working on and noticed that all of the resource paths weren't coming through right, so after looking over the code and talking with fletch3555 on Slack, I suggested switching to url() calls for resources instead of the config() calls. This appears to work though I have not extensively tested the app (as I am very new to it). As far as I can tell this is a more robust way to set the paths and have them work with any kind of url/hostname configuration. Please review this thoroughly before merging. Thanks!

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -156,17 +156,17 @@
 
 @section('css')
 @if(!$dataType->server_side && config('dashboard.data_tables.responsive'))
-<link rel="stylesheet" href="{{ config('voyager.assets_path') }}/lib/css/responsive.dataTables.min.css">
+<link rel="stylesheet" href="{{ url('/vendor/tcg/voyager/assets') }}/lib/css/responsive.dataTables.min.css">
 @endif
 @stop
 
 @section('javascript')
     <!-- DataTables -->
     @if(!$dataType->server_side && config('dashboard.data_tables.responsive'))
-        <script src="{{ config('voyager.assets_path') }}/lib/js/dataTables.responsive.min.js"></script>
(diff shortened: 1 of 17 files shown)


turtuga:
I don't like this the config is best way...

abdgad:
Thanks for taking the time to make this PR. The problem with this is that it's hard coding the assets url, and changing it in the future you will have to change it in all places. A better way is to just use `url()` in the config in `/config/voyager.php`. But there is still a problem. With secure sites, using `url()` requires an extra argument to generate `https://`. So this is not portable for all users.
My suggestion is, if there is anyone having issues with the assets url, he can update `/config/voyager.php`, and use `url()` explicitly and customize it to fit his needs. This is a cleaner and easier solution.

bhartvigsen:
I'm fine with that solution, let me try that out. The issue is this thing
is broken out of the box with the current setup.

abdgad:
What is broken?

bhartvigsen:
The app was totally broken the first time I tried it (No CSS or JS due to
broken paths). I then nuked my git working tree to try the technique of
setting paths in the config and re-installed the app. This time around it
worked fine. I'm still having to use the PR method of setting paths in
override views, but the rest of the app is working. I believe there was
probably something the app expected first time around in the DB that wasn't
there, but I'm not sure. I basically installed it twice following
instructions verbatim and the paths only worked the second time I installed
it w/ composer. The only difference between the two installs was that the
second time around I already had the DB in place and I did not use dummy
data.

abdgad:
Thanks for notifying us about this. Another PR has been opened which will handle it #988

PR #947, opened 2017-03-15
Adding new code editor field and code editor ability in the settings

tnylea:
In this PR users can now select 'Code Editor' in the bread dropdown. This means that you can add an Ace Editor to any Data Type.

screen shot 2017-03-15 at 12 00 51 pm

The Ace Editor is also available in the Settings Dropdown and the Settings Additional Options.


tnylea:
@fletch3555 Let me know what you think :)

Thanks.

tnylea:
Thanks Mark 👍  You rock :)

fredhy:
VERY NICE! Good Job. How to use code editor and variable at the same time? Example: Put into Code editor `<div> Hello Word.... {{$foo->bar}}</div>`, my problem is the render laravel. Thanks!

PR #948, opened 2017-03-15
use intervention\image to orientate the uploaded media.

hapheus:
Most portrait oriented pictures are actually landscape pictures with an orientation in their exif data. 
So most browsers will display the pictures rotated by 90°. Fortunately you already requiring  the intervention / image composer library so we can easily use it to orientate the picture during the upload process.


tnylea:
Looks great! Thanks for adding that 👍

PR #950, opened 2017-03-15
Removing the unneccessary ace_editor for the options_editor since it …

tnylea:
Removing the .ace_editor class from the #options_editor, since it already is instantiated as an Ace Editor via the ID, it is getting re-instantiated by having the class .ace_editor.

Removing this class will make it work correctly.

@fletch3555 Sorry man, need a quick update, can you merge this in for me.

I'm simply removing a class that added in the previous commit 👍


fletch3555:
As soon as Travis finishes, sure

tnylea:
Thanks man!

PR #957, opened 2017-03-17
menu on 404 page

turtuga:
When I add a menu() in 404 page we get `Call to a member function getAction() on null`

### DIFF
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -117,7 +117,11 @@ class VoyagerServiceProvider extends ServiceProvider
      */
     protected function addStorageSymlinkAlert()
     {
-        $currentRouteAction = app('router')->current()->getAction();
+        if (app('router')->current() !== null) {
+            $currentRouteAction = app('router')->current()->getAction();
+        } else  {
+            $currentRouteAction = null;
+        }
         $routeName = is_array($currentRouteAction) ? array_get($currentRouteAction, 'as') : null;
 
         if ($routeName != 'voyager.dashboard') {


marktopper:
Thanks!

src/VoyagerServiceProvider.php
```diff
@@ -117,7 +117,11 @@ protected function registerViewComposers()
      */
     protected function addStorageSymlinkAlert()
     {
-        $currentRouteAction = app('router')->current()->getAction();
+        if (app('router')->current() !== null) {
+            $currentRouteAction = app('router')->current()->getAction();
```
Frondor, line 121: You should cache the instantiation of `app('router')` in my opinion. Even `$currentRoute = app('router')->current()` if that's the only use you're giving to it in that function.

marktopper: Good point, will do a PR.

Frondor: But wait! @marktopper I didn't notice earlier but since that's a service provider, can't you simply `$this->app->router->current()` ?

marktopper: I believe you are right.

turtuga: @Frondor I don't understand what you saying can you please explain please ?

marktopper: Maybe doing this:
```
$router = app('router');
if ($router->current() !== null) {
    $currentRouteAction = $router->current()->getAction();
} else  {
    $currentRouteAction = null;
}
```

Frondor: But since your working on a service provider, you have access to `$this->app` hence `$this->app->router` is already instantiated.

marktopper: So would this be ideal?
```
$router = $this->app->router;
if ($router->current() !== null) {
    $currentRouteAction = $router->current()->getAction();
} else  {
    $currentRouteAction = null;
}
```

Frondor: Yeah, I wouldn't create a new variable for that, since its already stored in `$this->app->router` (counts as a variable, stored in memory).

So it'd be

```
if ($this->app->router->current() !== null) {
    $currentRouteAction = $this->app->router->current()->getAction();
} else  {
    $currentRouteAction = null;
}
```

marktopper: 👍 True that, it is a singleton I see.

PR #963, opened 2017-03-17
Disable database migration generation

abdgad:
Generating migrations is not useful at the moment since it's only generating a stub.
Adding full migration generation is on my todo list. So until that is added, we should just disable this.

### DIFF
--- a/resources/views/tools/database/vue-components/database-table-editor.blade.php
+++ b/resources/views/tools/database/vue-components/database-table-editor.blade.php
@@ -25,7 +25,7 @@
 
             <div class="col-md-3 col-sm-4 col-xs-6">
                 <label for="create_migration">Create migration for this table?</label><br>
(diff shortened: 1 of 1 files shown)


connor11528:
If we are not creating a migration file and we click around in the UI to create our database will we need to do the same clicks in our production environment? 

Essentially if there are no migrations how does the database get setup? Thank you!

fletch3555:
@connor11528, This functionality has been disabled for nearly 4 months.  The functionality never fully worked.  It would only create a stub migration file.  This pull request was to disable the functionality until we can fully implement it.

For your question, you have a couple options.

1. You manually reproduce your tables and BREAD through Voyager's UI.
2. You manually create migrations as needed.
3. You do a database backup and restore to port it over to your production environment.

connor11528:
Thank you! Created a PR to remove the button until the feature is implemented #1464

PR #968, opened 2017-03-18
Dynamic routing

gabrieltakacs:
I added a new feature dynamic routing (one of features that I missed in Voyager). A new model PageType is created. Page type defines which controller and action handles rendering of every page of that page type. So you basicly assign a page type to every page and you don't need to write a new route for every single page in your project - you just create a controller (f.e. ArticleController) with an action (f.e. show), create an Article page-type and set ArticleController and show action to this page type.

Feel free to ask me for more details.

### DIFF
--- a/publishable/database/migrations/2017_03_15_200930_create_page_types_table.php
+++ b/publishable/database/migrations/2017_03_15_200930_create_page_types_table.php
@@ -1,0 +1,36 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreatePageTypesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('page_types', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('name', 32);
+            $table->string('controller', 255);
+            $table->string('action', 255);
+            $table->timestamps();
+
+            $table->index(['name']);
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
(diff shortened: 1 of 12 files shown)


abdgad:
Thanks for creating this PR. However, can you please explain further what it does by giving more example use cases?

gabrieltakacs:
#Yes, of course, I'll try.

It allows users to specify, which Laravel controller & action will be used to render 
voyager_page_types
voyager_pages

page. This is reached by adding a new element - **Page types**. Every project has many page types - for example _Homepage_, _Article_, _List of articles_, _Contact page_. We know that for example every `Article` page should be rendered the same way, using the same controller, action, layout and view. So you assign every single article page the `Article` page type and for the Article page type you set that it should be rendered using `ArticleController` and `show` action.

I'm attaching two screenshots. In these pictures you can see 3 page types: `Article`, `List of articles`, `Blog post` and 3 pages: `Articles list`, `My first article`, `My second article`. By assigning a page type to each of these pages, we can ensure that `My first article` and `My second article` will both be rendered by `ArticleController` and `show` action. Likewise we can also set that `Articles list` page will be rendered by `ArticleController` and `index` action. 

In the most CMSs, this is done by templates (you set a template for every page). The advantage of my solution is that the "template" is defined by controller and action, so you can write some custom logic for every "template" or page type. This still enables us to separate logic from the view (other CMSs tend to mix logic with the view in one template file).

If you have any questions, don't hesitate and contact me.

abdgad:
Thanks for the explanation. The thing is, we are planning to remove `Pages`, `Categories` etc.. from Voyager, and include them in a separate package (as a Plugin), because not all people will use them. We want Voyager to only contain core features which will be used by everyone.

gabrieltakacs:
Okay, I understand. Thank you. 

I love Voyager!

PR #975, opened 2017-03-20
Add html5 validation for user create / edit

gnovaro:
if you try to create and user without complete basic required fields

### DIFF
--- a/resources/views/users/edit-add.blade.php
+++ b/resources/views/users/edit-add.blade.php
@@ -49,25 +49,31 @@
                                 <label for="name">Name</label>
                                 <input type="text" class="form-control" name="name"
                                     placeholder="Name" id="name"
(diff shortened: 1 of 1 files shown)


turtuga:
I think the better way is not hard code but by options in bread..

Frondor:
Indeed, it shouldn't be hardcoded. It should make use of the validation capabilities BREAD provides already through `details` JSON api.

In that case you have to encode the details like `$details = json_encode($details)`, then check if `$details->validation` is set and then do something like `if (strpos($details->validation->rule), 'required') { echo ... } ` (which won't work if rules are set as an array).

Besides of that, and since we don't have javascript validation yet, I don't consider it consistent enough to be added, not just yet. 

Server side validation should be enough for your case.

fletch3555:
I'm in agreement with @turtuga and @Frondor.  This should be handled through the BREAD configuration JSON field.

PR #981, opened 2017-03-21
Fix: Relationship value display was not able in (B)READ.

matedon:
For example: In the dummy data "categories" the Browse was not possible for "parent_id".
There was an ERROR when you want to Browse your data with Relationship.
Now it's ok.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -58,7 +58,7 @@
                                                 @if ($data->{$row->field} && isset($options->relationship))
                                                     {{ $data->{$row->field}->implode($options->relationship->label, ', ') }}
                                                 @endif
-                                            @elseif($row->type == 'select_dropdown' && property_exists($options, 'options'))
+                                                @elseif($row->type == 'select_dropdown' && property_exists($options, 'options') && property_exists($options->options, $data->{$row->field}))
 
                                                 @if($data->{$row->field . '_page_slug'})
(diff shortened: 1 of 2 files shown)


Frondor:
Can you please help us reproduce the error you was previously experiencing? To be honest, your editor added so many style changes that it makes a bit painful to read them, and I can't realize what you really wanted to do

matedon:
Ok. I try to explain:

1. Create a "Region" table with these columns:
id, name, description, region_id
2. Setup BREAD for this table, set region_id to "Select Dropdown"
```
{
    "default": "",
    "null": "",
    "options": {
        "": "--"
    },
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
```
3. Region model functions
```
protected $table = 'region';
public $timestamps = false;
public function regionId()
  {
    return $this->belongsTo(self::class);
  }
```
4. Browse the admin http://myapp.local/admin/region
5. Crashes -> I can not see self relational (parental) region id data

Frondor:
The relationship links are getting some big improvements, due to a failed PR that was previously merged.
We may address this PR after that one's fixed.

fletch3555:
@Frondor, have those improvements you mentioned been completed?  If so, can you please review this again?  There are currently merge conflicts that will need to be resolved.

Frondor:
@fletch3555 not really, I dismissed my PR because I became really busy lately and that code demands lots of time.

fletch3555:
Too many formatting changes, and merge conflicts.  Closing this.  If these get fixed, we can reopen and reevaluate it.

PR #985, opened 2017-03-21
store the custom model file into correct path

callect:
if pepole modified models.namespace in config/voyager.php, then add new database table with option 'Create model for this table'  and save it. The model file can save into  a correct path that I configed.


marktopper:
Thanks! 👍

Issue #946, opened 2017-03-15
Can't clear the filled field.


dimsanchuk:
- Laravel Version: 5.4.12
- Voyager Version:  0.11.7
- PHP Version:  5.6
- Database Driver & Version: MySQL

### Description:
It is not possible to save a text field with an empty string, if there is already a record with this field filled.

### Steps To Reproduce:
1. In the context of the blog,  I have post with filled field "Post Content":
![image](https://cloud.githubusercontent.com/assets/13986110/23955440/8bbf774e-09a2-11e7-8174-2b4983a44dc2.png)
2. Now I am clearing the field "Post Content":
![image](https://cloud.githubusercontent.com/assets/13986110/23955510/c0e0e390-09a2-11e7-9996-b3582b642d72.png)
3. Click the button "Update Post" and I see again filled field "Post Content":
![image](https://cloud.githubusercontent.com/assets/13986110/23955697/587faaec-09a3-11e7-962e-7617e18e91b8.png)

dimsanchuk:
There's someone here?

dimsanchuk:
This happens not only when editing a record, but also on the edit page of the BREAD in the "Optional Details" field.

marktopper:
I see. Can you please get the SQL query that is being ran when doing the change?

dimsanchuk:
Saving with empty "Post Content":
![image](https://cloud.githubusercontent.com/assets/13986110/24157218/7367ac8e-0e61-11e7-991d-fb2eba528e01.png)
Saving with changed data in "Post Content":
![image](https://cloud.githubusercontent.com/assets/13986110/24157260/967a037a-0e61-11e7-976a-2dfacf27ee2f.png)
I also found a place in the code that prevents change:
https://github.com/the-control-group/voyager/blob/master/src/Http/Controllers/Controller.php#L76 (lines 76-78)

marktopper:
Thanks, it seems like it does not even trigger the change in the controller just as you said.
Will have to figure out why those lines where added.

marktopper:
Can you share the BREAD configuration for the `posts` table?

dimsanchuk:
![image](https://cloud.githubusercontent.com/assets/13986110/24161048/0c4906c8-0e6c-11e7-91be-b8751383eabb.png)

marktopper:
Can you test if #987 helps?

dimsanchuk:
And nothing has changed. Only I believe that this is a significant issue?
[...]

PR #987, opened 2017-03-21
Allow null values

marktopper:
Possible fix for issue #946.
Have to be tested well before merged.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -71,14 +71,8 @@ abstract class Controller extends BaseController
                 }
             }
 
-            if (is_null($content)) {
-                // Only set the content back to the previous value when there is really now input for this field
-                if (!$request->has($row->field) && isset($data->{$row->field})) {
-                    $content = $data->{$row->field};
-                }
-                if ($row->field == 'password') {
-                    $content = $data->{$row->field};
-                }
+            if (is_null($content) && $row->field == 'password') {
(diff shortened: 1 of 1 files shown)


dimsanchuk:
I checked. It didn't help.
If the value of the parameter is empty string, then $request->has($row->field) returns false. Therefore, current value is overriten by the previous value.

marktopper:
I see. Doing another commit.

marktopper:
@Dimsanchuk: Can you please test this out again, I will also do some proper testing for this in order to make sure that it didn't break anything else.

dimsanchuk:
@marktopper, It works, but... field must be nullable, else error occurs:

> 1048 Column 'body' cannot be null (SQL: update `blog_posts` set `body` = , `updated_at` = 2017-03-21 19:49:48 where `id` = 3)

I don't know if this decision is correct.

dimsanchuk:
I found, that it is reset all setted images (maybe and files too).
We need to process null values based on the type of field. Separately for text fields and file fields.

marktopper:
Good point @Dimsanchuk.

I will try to make a new commit that handle this.

fletch3555:
@marktopper, any update on this?

fletch3555:
Paging @marktopper.  Any update? Otherwise I'm going to have to close this.

marktopper:
Im going to close this, since it's not yet the perfect solution...

PR #988, opened 2017-03-21
Improve Voyager assets handling

abdgad:
- [X] VoyagerServiceProvider now publishes the assets to `config('voyager.assets_path')`
- [X] Add `voyager_asset()` helper function to be used in the views
- [x] Change all views to use `voyager_asset()`


abdgad:
@marktopper this maybe merged.

PR #993, opened 2017-03-22
Added support for theming through introducing a function that checks …

Rndwiga:
1. Added support for theming through introducing a function that checks whether theme support has been switched on, if so, it loads the theme if not it load default views
2. Theme support provided by caffeinated\themes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "illuminate/support": "~5.3.0|~5.4.0",
         "intervention/image": "^2.3",
         "doctrine/dbal": "^2.5",
+        "caffeinated/themes": "^3.0",
         "larapack/doctrine-support": "~0.1.0",
         "arrilot/laravel-widgets": "^3.7",
         "league/flysystem": "~1.0.35"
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -18,6 +18,7 @@
         "illuminate/support": "~5.3.0|~5.4.0",
         "intervention/image": "^2.3",
         "doctrine/dbal": "^2.5",
+        "caffeinated/themes": "^3.0",
```
marktopper, line 21: We can remove this from requirements and put it as a suggestion, and just note users that they need this when enabling the `use_theme` setting.

marktopper (changes requested):
Just a minor change to limit the requirements.

According to this, I have a concern, do we limit the user to use the `caffeinated/themes` library for the admin panel only, or will they be able to run it for both Voyager and their own custom front-end?

fletch3555:
@Rndwiga, please respond to @marktopper's review comments

karneaud:
Is this still applicable?

fletch3555:
This has been closed for over 2 years now. Signficant changes have occurred since then.  You're much better off explaining your use-case and asking for help in our slack group instead

PR #1018, opened 2017-03-27
Fix #1015

### LINKED ISSUE #1015: Question regarding Media Library and dot files
- Laravel Version: 5.3.30
- Voyager Version: 0.11.7
- PHP Version: 5.6.19
- Database Driver & Version: MySQL 5.7.11

---

## Question

- Is there anyway for the Media Library to exclude listing dot files such as `.gitignore`?

turtuga:
Fix https://github.com/the-control-group/voyager/issues/1015


abdgad:
This should be optional, so people who want to view those files can still do that.

turtuga:
@abdgad create a config item is an option?

abdgad:
I am not sure the best way to do this. But I think the easiest way is to add a simple filter on the fronted to show/hide hidden files.

turtuga:
Yes, but if we pretend to end users not access hidden files.

marktopper:
I believe that it should be optional in some way. A configuration would be fine with me.

turtuga:
I have updated with config

fletch3555:
Out of curiosity, wouldn't it be easier/better to simply change the file system permissions for those files? If the webserver doesn't have Read access, it can't show them.

abdgad:
I don't like putting something like this in the config. So to view hidden files, the user will have to modify the config, which is not convenient. This should be handled in the UI. A simple checkbox to show hidden files will be enough imo.

marktopper:
I like approach @abdgad. However then it would be up to the end user to choose wether or not he want to see the hidden dot-files. The developer can not do anything against it. Thoughts?

turtuga:
@mark yeah that's my point

akazorg:
We have 2 types of settings (as @marktopper and @abdgad mentioned):
1. Default Settings (created by developer/voyager admin)

    Imho, we should really think about upgrade Voyager Settings and make it group-able. We would then have Media Settings accessible by a Tab system in the UI, giving permission to admin only.

2. User Settings
    What if we create a User Settings system? Something simple, just a JSON field in users table where we store all the settings we need for the system. Sooner or later we will need it.

My suggestions all involve work, hope you like it ^^

marktopper:
@akazorg: I brought up the idea for group-able settings once before but it was down voted.
However now it could be the time to pull up this vote again.

Also according to this I believe that both could be a option, a developer might however be required to hide specific files for certain reasons. So I think that if dot-files are not hidden by the developer, a later pull request could be allowing this setting for the end-user.
What are your thoughts on this @abdgad?

akazorg:
@marktopper I remember you being down voted when you came up with this idea. In fact it didn't make much sense, but I guess it wasn't the right moment.

You're right, Users Settings can be perfectly released later.

A small clarification, Voyager settings will always prevail over User settings, right? This way admin decides which files should be hidden all the time.

Ideas?

marktopper:
I am gonna merge this in and put a note to self to make it possible for users to have their own settings as well. But this will however make it possible for a developer to force users to not see dot-files, so it's still important to have this option.

PR #1021, opened 2017-03-27
Add basic support for many to many relationship

moe2566:


fletch3555:
I believe we already have support for Many2Many relations (@marktopper, correct me if I'm wrong), but it's a little simplistic.  I'm definitely open to having a UI to edit relations, instead of just the JSON config settings.

One request before we could even consider moving forward with this.  Please remove all of the `Tag` stuff.  We have no plans to expand on the existing Pages/Posts/Categories system in it's current state, so there's no reason to include it right now.  Of course, you're more than welcome to keep it in your own app. :)

moe2566:
Okay, I removed them. They were there just for demonstration.

src/Http/Controllers/VoyagerDatabaseController.php
```diff
@@ -309,6 +309,36 @@ public function updateBread(Request $request, $id)
         }
     }
 
+    public function addBreadRelationship(Request $request, $id)
+    {
+        Voyager::canOrFail('browse_database');
```
fletch3555, line 322: I'm not sure this is the correct permission to use here.  Is this an artifact from you copy/pasting?

src/Models/DataType.php
```diff
@@ -55,6 +55,31 @@ public function deleteRows()
         return $this->rows()->where('delete', 1);
     }
 
+    public function addRelationshipRow($requestData)
```
fletch3555: I'm not 100% sure what this is doing (still reading through the code), but this doesn't seem like it belongs in the model.

src/Http/Controllers/Traits/BreadRelationshipParser.php
```diff
@@ -115,4 +118,16 @@ protected function relationToLink(Model $item, DataType $dataType)
...
+    protected function getField($item, $relationMethod)
+    {
+        $relationBuilder = $item->$relationMethod();
+
+        // original field named was used so we'll return that
+        if ($relationBuilder instanceof BelongsTo) return $relationBuilder->getForeignKey();
+
+        // While adding the relationship we used getRelationName() as field name,
+        // so we'll return that now too.
+        if ($relationBuilder instanceof BelongsToMany) return $relationBuilder->getRelationName();
```
fletch3555: I completely understand ignoring some of the other Laravel/Eloquent relationship types (morphTo, etc), but why only BelongsTo and BelongsToMany? What about their inverses (hasOne, hasMany)? what about the others (i.e. hasManyThrough)?

src/Http/Controllers/VoyagerDatabaseController.php
moe2566: Yes, it is an artifact from copy/pasting

src/Models/DataType.php
moe2566: It simply adds the relationship field to data_rows table. Do you think it belongs in DataRow?

fletch3555: I feel like it belongs in a controller.  Definitely not in the DataType model though.

src/Http/Controllers/Traits/BreadRelationshipParser.php
moe2566: Sorry, I was mainly concerned with BelongsToMany, I'll work on the others too.

fletch3555: Awesome thanks! I wasn't sure if it was a conscious decision or not.  I'd just like it to be as complete as possible.  If you don't have time, by all means say so and we'll get someone else to jump in.

moe2566: Yes please, I can surely use some help.
Thanks

marktopper:
@fletch3555: Nice review, however we currently have a lack of support for Many2Many relations since the ID is not stored on the related table.

fletch3555:
@marktopper, I was thinking we had something, but that it was hacky (like having to do two one2many relations against the pivot table).

src/Http/Controllers/VoyagerDatabaseController.php
```diff
@@ -309,6 +309,36 @@ public function updateBread(Request $request, $id)
...
+                'json',
+                'regex:/^.*(?=.*\brelationship\b)(?=.*\bkey\b)(?=.*\blabel\b)(?=.*\bmethod\b).*$/'
+            ],
+            'type' => 'required',
+        ]);
+
+        if ($validator->fails()) {
+            return back()->with($this->alertException($validator->errors(), 'Update Failed'));
+        }
+
```
moe2566: Just realized , this will fail. 
`alertException()` expects an `Exception`, `$validator->errors()` returns a `MessageBag`

marktopper:
I believe this is a great improvement for the methods we have now.

Great job @moe2566 and sorry for the delay.

fletch3555:
@moe2566 / @marktopper ,  There are a few conflicts that need to be resolved before this can be merged in.  Can one of you please handle it?

Thanks!

tnylea:
Hey @moe2566 

Thanks for this commit. I was testing it out and whenever I try and add a relationship, I get the following message:



Can you send me an example of a format that is valid in this box:



I tried adding something like:

```
{
    "relationship": {
        "key": "id",
        "label": "name"
    }
}
```

but I kept getting that message. Let me know and I would love to get this merged in. Appreciate it.

moe2566:
Hi @tnylea, sorry for the late reply.

You need to add a method property pointing to to Eloquent Model method that sets the relationship.

Eloquent class:
```
public function roles()
{
      return $this->belongsToMany('App\Role');
}
```
BREAD config:
```
{
    "relationship": {
        "key": "id",
        "label": "role",
        "method": "roles"
    }
}
```

tnylea:
Hey, thanks for this commit. I've actually used some of the code that you've written here to implement into the new relationships section coming to 1.0 so I'm going to close out this request.

Take a quick look at the new relationship section in 1.0 here: https://www.youtube.com/watch?v=gnSJ0Z2Fmso&t=53s

Thanks! Talk to you soon.

PR #1022, opened 2017-03-27
Replace php short tag GA for blade syntax

gnovaro:


fletch3555:
A small, but not insignificant, change.  Thanks!

akazorg:
I think this merge crashing Voyager.

PR #1026, opened 2017-03-28
[fix] GA fix

akazorg:
Previous merge was crashing Voyager.


fletch3555:
My fault, good catch!

Issue #1020, opened 2017-03-27
Updated bread permissions in 2 blade templates


jcanchor:
- Laravel Version: 5.4.16
- Voyager Version: 0.11.7
- PHP Version: 7.0.13
- Database Driver & Version: MySQL 5.7.16

### Description: 

I ran tests after installing BREAD system for the various roles I created across existing admin tables in Voyager. I found that having permission for Browse and Read only still allowed Add and Edit buttons to appear, and in View record instances, there was no button to return to the list.

I copied two templates and put them in my `resources/views/vendor/voyager` folder. These have been working for me, and avoid showing buttons that produce errors for the user on permission check.

### bread/read.blade.php

"resources\views\vendor\voyager\bread\read.blade.php"

Here is revised `<h1>` tag in `@section('page_header')` with permission check and return to list button.

        <i class="{{ $dataType->icon }}"></i> Viewing {{ ucfirst($dataType->display_name_singular) }} &nbsp;
        @if (Voyager::can('edit_'.$dataType->name))
            <a href="{{ route('voyager.'.$dataType->slug.'.edit', $dataTypeContent->getKey()) }}" class="btn btn-info">
                <span class="glyphicon glyphicon-pencil"></span>&nbsp;
                Edit
            </a>
        @endif
        <a href="{{ route('voyager.'.$dataType->slug.'.index') }}" class="btn btn-warning">
            <span class="glyphicon glyphicon-list"></span>&nbsp;
            Return to List
        </a>        

### posts/browse.blade.php

"resources\views\vendor\voyager\posts\browse.blade.php"

Here is revised `<h1>` tag in `@section('page_header')` with permission check

        <i class="voyager-news"></i> {{ $dataType->display_name_plural }}
        @if (Voyager::can('add_'.$dataType->name))
            <a href="{{ route('voyager.'.$dataType->slug.'.create') }}" class="btn btn-success">
                <i class="voyager-plus"></i> Add New
            </a>
        @endif

Here are the revised show, edit and delete buttons near bottom of dataTable with permission check.

        <td class="no-sort no-click">
            @if (Voyager::can('delete_'.$dataType->name))
                <div class="btn-sm btn-danger pull-right delete" data-id="{{ $data->id }}">
                    <i class="voyager-trash"></i> Delete
                </div>
            @endif
            @if (Voyager::can('edit_'.$dataType->name))
[...]

PR #1027, opened 2017-03-28
Update permissions on 2 blades

jcanchor:
This is an update for issue #1020 submitted yesterday.

It adds permission check to blades so that Edit, Add, Delete buttons appear correctly based on set permission, especially in cases where user only has Browse and Read permission.

It also adds a "Return to List" button in the `bread/read.blade.php`


marktopper:
Thanks! Wonderful update! 👍

Issue #801, opened 2017-02-21
Avatar is not working, showing broken images


JohanIskandar:
- Laravel Version: 5.4.12
- Voyager Version: dev-release/v0.11"
- PHP Version: 7.1.0
- Database Driver & Version:

Hi,

  I am not sure why the avatar is broken. I recently just upgraded to the latest development version

![image](https://cloud.githubusercontent.com/assets/1010323/23150647/82b15d66-f827-11e6-8b0a-84c039282323.png)

I was wondering whether I have missed a step when upgrading it, it seems to me, no matter what image I picked for the avatar is not working. thank you

regards,

Johan

VladDobrescu:
Check the symlink. Had the same issue in a vagrant-homestead environment, the `symlink` could not be made. If you're in the same boat (huehue), exit the box and run an `ln` command

JohanIskandar:
Hi BlenderBro,

  Thanks for responding. I am not using Vagrant, just on my localhost (Mac OS). Could you point out how to do it, I mean the steps. Sorry, I am just a beginner. Thanks.

regards,

Johan

VladDobrescu:
Open a terminal and cd into your project folder. Run `php artisan storage:link`. If that fails, read this #722

marktopper:
Please use the released version of `v0.11.1` instead of the dev branch. This branch will be deleted soon.

bab55z:
i tried everything but nothing works! please help needed. i'm runnnig v0.11.1 on laravel 5.4, (Windows)

marktopper:
@bab55z: Please make a new issue with as many details as possible.

sweggggg:
i just fresh install voyager, i still have the problem.

i tried php artisan storage:link in my project folder, it says "The "public/storage" directory already exists." still have the same problem

any solutions?

sikhub:
@sweggggg You need to specify `APP_URL` in `.env` file

apiaget:
I had the same problem and it is now resolved, all I did was replacing the following in my .env file :
```APP_URL=localhost:8000```
replaced by
```APP_URL=http://localhost:8000```
I was running Laravel 5.4.16 with voyager 0.11.8 that I installed by following your guide.
The network tab in chrome's dev tools was listing the error as : ```net::ERR_UNKNOWN_URL_SCHEME```

rudiroed:
Yes me too, please help.
I have tried to change `APP_URL` run this `php artisan storage:link` but still not working for me.
[...]

PR #1033, opened 2017-03-29
[Fix] broken images network error

apiaget:
Hi,

That small change repaired my avatars so it seems usefull to have it in the installation guide.

Any feedbacks are welcome.

Related issue : #801


marktopper:
Thanks

PR #1039, opened 2017-03-31
Changing function name to minimize overridden functionality

fletch3555:
Minimal change, but straight-forward enough.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "require-dev": {
         "phpunit/phpcov": "~3.0",
         "phpunit/phpunit": "~5.7.14",
-        "laravel/framework": "~5.3.8|5.4.*",
+        "laravel/framework": "~5.3.29|5.4.*",
         "orchestra/testbench": "3.3.*|3.4.x-dev",
         "orchestra/database": "3.3.*|3.4.x-dev",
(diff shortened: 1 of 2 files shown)


marktopper:
TravisCI fails on Laravel 5.3.
Was the old function name used in Laravel 5.3?
Cause then I think we should make that function just call the new function in order for this to work.

fletch3555:
I don't see why... The failure is a "route not found" for `/home`, not the function not being found.  Github shows that the function hasn't changed between 5.3 and 5.4: https://github.com/laravel/framework/blob/5.3/src/Illuminate/Foundation/Auth/RedirectsUsers.php

----

Edit: Just tested locally with Laravel 5.3 and Voyager v0.11.8.  Works fine

marktopper:
I have found it now.
It's because this function first came to Laravel in v5.3.29.
And TravisCI is not testing the newest Laravel 5.3 version.

Solution is to require at least Laravel v5.3.29.

marktopper:
This should fix the TravisCI tests.
https://github.com/the-control-group/voyager/pull/1039/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L28

marktopper:
And seems like it fixed it.
Merging in.

PR #1085, opened 2017-04-06
Add method for Post to get image thumbnails

emchooo:
We have option in **Optional Details** to crop image in different sizes, and because image is saved in image field as full path I think we should have method if we need to get specific cropped version of image.

In Blade we can use **$post->image** to get full image path, but it is problem to get thumbnail versions of that image.

With this method we can get thumbnail version by using **$post->thumbnail('cropped')** to get cropped version of saved image.


marktopper:
Great idea 👍

PR #1088, opened 2017-04-07
New display validation fields and selecting routes in menu item

AdrianKuriata:
Hi, i did new validation. Currently validation working really basicaly. Just sending us message error, but we don't know which field is bad. I did validation with Ajax Request and displaying messages under fields if is bad, check this YouTube Video for Example how it work. I wanted to say this is validation for every created Bread, rules you must add like in the Voyager documentation. I hope we discuss about this and this will be add in the next update, because i think this is really better validation like is now.

[Check example on YouTube!](https://www.youtube.com/watch?v=GIe_4ACN8JA&feature=youtu.be)

//Edit

I added selecting routes in edit-add menu item, check example on youtube
[Example on YouTube!](https://www.youtube.com/watch?v=caYGSJirOcQ) (Sorry for the music...)

### DIFF
--- a/publishable/assets/js/val.js
+++ b/publishable/assets/js/val.js
@@ -3,9 +3,8 @@ $(document).ready(function(){
 		$($(this).parents('form')).submit(function(e){
 			e.preventDefault();
 			var form = $(this);
-			var method = $(form).find('[name="_method"]').attr('value');
-
-			if(method != 'DELETE' || method == undefined)
+			var method = $(form[0]['innerHTML']).find('.form-group');
+			if(method.length > 0)
 			{
 				var url = $(this).attr('action');
 				var data = new FormData(this);
--- a/publishable/assets/js/val.js
+++ b/publishable/assets/js/val.js
@@ -1,29 +1,42 @@
 $(document).ready(function(){
-	$(".form-edit-add").submit(function(e){
-		e.preventDefault()
+	$("[type='submit']").on('click', function(e){
+		$($(this).parents('form')).submit(function(e){
+			e.preventDefault();
+			var form = $(this);
+			var url = $(this).attr('action');
+			var data = new FormData(this);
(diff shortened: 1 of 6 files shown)


AdrianKuriata:
~~1. This validation not working properly with FILE fields, i will work hard for repair it.~~ Fixed

~~2. I changed validating fields from "form" tag to .form-edit-add - this is a problem with files, i getting 500 error with admin/upload if i want to use form or [method='POST'] - currently this is working just with created breads which has .form-edit-add class into form tag and this is working perfectly(without file field).  So this is workign with Categories, Posts, Pages and every new created bread by you.~~ Fixed

//Edit
Ohh, this is not working with form[method='POST'], because on every page is more forms than one. Need to add class to every form which will connect all forms from BREAD(like .form-edit-add).

//Edit

I think i fixed all issues and everythink working perfectly, you can validate files by mimes etc. Now is not problem with couple forms in view, just check it and test!

davodavodavo3:
@AdrianKuriata do you check validation attempt if field has a description option in JSON?

AdrianKuriata:
@davodavodavo3 What do you mean? I added description to one field and everything working ok.

davodavodavo3:
I mean helper descrition to field. Like this.
```
{
    "description": "A helpful description text here for your future self."
}
```
If you mean that too all right)))

AdrianKuriata:
@davodavodavo3 Sure, i added this into field and validating does not interfere. Do you have any ideas how can i check everything is good?(missing issues what i wrote before).

davodavodavo3:
For now i dont have any ideas where cant ba a issue.

AdrianKuriata:
I need to close this pull request, i will better work at this in few next days and add working perfectly validation and selecting routes for menu items, sorry guys, this is deprecated.

PR #1098, opened 2017-04-08
Added lock screen and css improvements

davodavodavo3:
Unit of measure 'px' is redundant 0px = 0pt = 0em = 0 :)
There is many CSS properties, which are being overwritten by default.I dont tuch them.

Change container custom paddings .
Lock screen work with session and Voayger middleware
Lock screen page is with animation and use only Voayger libraries(js,icons)


davodavodavo3:
Old container:
old

New container:
accuracy

Lock Screen Link: Can be added as button in future 
lockscreen link

Lock Screen page for admin panel users: 
lock screen

AdrianKuriata:
Awesome!

marktopper (changes requested):
Looks good @davodavodavo3. Great job.
One thing, please make a configuration for enabling/disabling the lock feature 👍

src/Http/Controllers/VoyagerController.php
```diff
@@ -24,6 +24,35 @@ public function logout()
...
+        $password = $request->password;
+        $users = Voyager::model('User')->find($id);
+
+        if (\Hash::check($password, $users->password)) {
+            session(['admin_lock' => 0]);
+
+            return redirect()->route('voyager.dashboard');
+        }
+
+        return redirect()->route('voyager.lock')->with($this->alertSuccess("Password is wrong!"));
```
fletch3555: `alertSuccess("Password is wrong!")` is counter-intuitive.. I'm sure you just used what was already there, but can this be changed somehow?  Should it be? @marktopper?

davodavodavo3:
@marktopper Why add to the configuration, a functional that does not interfere with the user, it does not automatically turn on, only by clicking on it.

davodavodavo3:
I can do changes in lockScreen method , if config lockscreen is true, or if in config->dashboard->navbar_items has item LockScreen

akazorg:
A configuration option would be great, maybe using an option type of time, where we define how many seconds before lock screen would turn active, while for disabling this feature we would set it to 0.

DODMax:
Up! Can we get this awesome feature?

I kind of agree that we don't need a setting since there is **no** auto locking, if user really don't want the *lock* link to show they can comment it in the `voyager.php` config file.

```php
    'dashboard' => [
        // Add custom list items to navbar's dropdown
        'navbar_items' => [
            /* ... */

            //'Lock' => [
            //    'route'      => 'voyager.lock',
            //    'icon_class' => 'voyager-lock',
            //],
            'Logout' => [
                'route'      => 'voyager.logout',
                'icon_class' => 'voyager-power',
            ],
        ],
```

akazorg:
Maybe this is a silly question, but why do we need a Lock screen, when we have a logout?

I like the idea of the auto lock - or should we call it auto-logout?

I know this is somehow unrelated, but the user navbar dropdown should be handled by the Voyager Menu, or we loose the multilingual system for that nav. Besides, anyone can change the menu without the need of change the config file.

Ideas? Feedback?

AdrianKuriata:
@akazorg i think lockscreen is good when you need to go shop for 10minutes and you dont want logout from your account, but in house is someone who can do something bad on your account, this is enhacment, you know, this protect your account when you are away from keyboard.

fletch3555:
I also don't understand the need for this.  Websites are intended to be stateless, with no single request being wholly dependent on another.  They should not be concerned about account security in this manner (and should instead just use session timeouts to force user logout for security).  A "lock screen" is more of an "OS" function, not one for a website.

This is equivalent to logout functionality (already exists) with the functionality to return to the last page you were on (arguably not needed if an app is designed well enough).

Consider this a down-vote from me.

AdrianKuriata:
I think you can lock it because davodavo wont change anything in this PR, he wanted help with improvement Voyager but time which is waste to update PR's is horrible, same for me so i dont help too.

PR #1112, opened 2017-04-11
bug fix

itgoldman:
should call isBreadTranslatable with $dataTypeContent variable, and not $model which might not be defined


marktopper:
Thanks

PR #1114, opened 2017-04-11
Small post view improvements.

gnovaro:
Move sluglify js below script definition, Remove php inside blade syntax


AdrianKuriata:
Maybe better idea will be add Laravel Collective and using it for creating forms?

PR #1115, opened 2017-04-11
Excerpt field nullable

gnovaro:


abdgad:
Existing migrations should not be modified, instead new migrations should be created to make the necessary changes.

PR #1116, opened 2017-04-11
Create 2017_11_04_000000_alter_post_nullable_fields_table.php

gnovaro:
Make some posts fields nullable by default


marktopper:
I believe you have to do like
```
$table->text('excerpt')->nullable()->change();
```

fletch3555:
@gnovaro, please update the PR with @marktopper's suggestion

PR #1125, opened 2017-04-13
make multiple images preview like in settings

turtuga:
Make multiple images preview like in settings
multiple


marktopper:
The blur is just a part of the image, right?

turtuga:
yes 👍  sorry

PR #1128, opened 2017-04-14
Redirecting to intended URL after Login

itgoldman:
This is done by setting the `url.intended` session key in the middleware prior to redirecting to voyager.login.
I'm not sure this is the best way to do this, but it works.
If it's cool, I can also add a config option for this feature.

### DIFF
--- a/src/Http/Middleware/VoyagerAdminMiddleware.php
+++ b/src/Http/Middleware/VoyagerAdminMiddleware.php
@@ -24,6 +24,12 @@ class VoyagerAdminMiddleware
             return $user->hasPermission('browse_admin') ? $next($request) : redirect('/');
         }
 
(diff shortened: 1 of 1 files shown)


fletch3555:
I don't think this is the best way to handle that issue.  Let me try to think up a better one.

marktopper:
For me it's seems like a good approach.
However did you come up with anything @fletch3555?

itgoldman:
Hi
I'm kinda new here. Got some questions:
What am I looking for @ fletch? I assume it's about the post login
redirect?
Should I build it locally? Is that potentially risky for example having
malicious code?
Thanks, Itay

איתי גולדמן
www.itgoldman.com - אינטרנט ומערכות מידע

Issue #1144, opened 2017-04-18
"Optional Details" Field updates race condition with multilingual


chrisboustead:
- Laravel Version: 5.4.*
- Voyager Version: 0.11.9
- PHP Version: 7.1.x
- Database Driver & Version: MySQL 5.7.14

### Description:
When multilingual support is enabled, and a user attempts to edit the "Optional Details" fields in the add/edit bread screens, there is a race condition which occurs, and sometimes will not save the "Optional Details" values.

This occurs because the JSON is parsed for validity on form submission, then updated the nearby textarea element.  When multilingual support is also enabled, the jQuery plugin transforms the translations into JSON for the nearby hidden inputs of each field, then immediately submits the form.

So the form is usually submitted by multilingual plugin when the "Optional Details" JSON has not yet been fully processed.

### Steps To Reproduce:
* Enable Bread translations in `app/voyager.php`
* Edit Bread for any translatable schema in `Tools > Database`
* Add some JSON to the "Optional Details" field for one column
* Submit the form.
* The JSON added to "Optional Details" was not saved.

akazorg:
Hi @chrisboustead, 

In fact it's kind of tricky because we need to update all hidden inputs (containing each translation), before submitting the form, otherwise the latest form changes would not be affected.

I will check this.

chrisboustead:
Agreed, I think the best solve to avoid future issues would potentially be adding a validation library to all forms.  Obviously a heavy lift for what is only a small issue right now. 

For now maybe removing the `submit` event from multilingual.js and manually calling `prepareData` from the javascript section of each multilingual view with a form.

chrisboustead:
If there is a validation library you've identified as a good candidate for voyager, I'd be happy to lead the effort in incorporating into the project.

akazorg:
On the multilingual.js, we can define a optional callback function to take care of this issue (and future ones). I have to check that and see what's happening.

About the validation it sounds a great idea, but I can't say much because I haven't work with that, however, I've seen some activity with this topic.

chrisboustead:
Submitted a quick pull request which fixes the issue, though I do think a better long-term solution is reusable validation library which can be hooked by multilingual, ace editor, etc. 

https://github.com/the-control-group/voyager/pull/1147

akazorg:
@chrisboustead join us on slack channel! :)

PR #1147, opened 2017-04-18
Bread edit-add race condition - Issue/fix-1144

chrisboustead:
Address issue #1144

### DIFF
--- a/resources/views/tools/database/edit-add-bread.blade.php
+++ b/resources/views/tools/database/edit-add-bread.blade.php
@@ -276,16 +276,16 @@
 
             formTextEditors.each(function () {
                 var textarea = $(this),
-                        mode = textarea.data('editor'),
-                        editDiv = $('<div>', {
-                            position: 'absolute',
-                            width: 250,
-                            resize: 'vertical',
-                            class: textarea.attr('class')
-                        }).insertBefore(textarea),
-                        editor = ace.edit(editDiv[0]),
-                        _session = editor.getSession(),
-                        valid = false;
+                mode = textarea.data('editor'),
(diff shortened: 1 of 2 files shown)


resources/views/tools/database/edit-add-bread.blade.php
```diff
@@ -328,24 +332,56 @@ class: textarea.attr('class')
...
+                            // Field is empty
+                            $textArea.val('');
                         }
+                    });
+
+                    // If multilingual, prepare data manually.
+                            @if ($isModelTranslatable)
+                    var multilingual = $('.side-body').data().multilingual;
+                    if(multilingual) {
+                        multilingual.prepareData();
```
akazorg, line 374: Honestly, I would rater see the prepareData method being handled in the multilingual.js, besides, we may need to handle this on other views and we would be repeating code.

chrisboustead: I think it's fine for `prepareData` to be handled in multilingual.js, but I don't think multilingual should be submitting the form.  What do you suggest as an alternative?  Happy to take a different approach

akazorg: we're talking on Slack, something will come up here.
Just writing this to let others know about the progress.

akazorg:
@marktopper, please don't merge this, I've sent PR #1148 fixing this.

@chrisboustead thanks for your help!

marktopper:
Should this then be closed now that #1148 is merged?

akazorg:
Yes @marktopper, you can close this.

PR #1165, opened 2017-04-23
[Added] Sortable BREAD Items

### LINKED ISSUE #994: New Feature Suggestion - Order Fields
- Laravel Version: 5.4.
- Voyager Version: 0.11.7
- PHP Version: 7.0.15 & 7.1.1
- Database Driver & Version: MySQL 5.6.35 & MariaDB

### Description:
First of all thank you for creating one of best CRUD system I've used for Laravel. Respect for the people who build it!! 
One of the things I would suggest **is to order fields, just like the menu items**. 

Or some new param we could use in the details json, and use it for the order. 
```
{
  "order": "1"
}
```

Until this feature is available what I did was add a new column to this table_rows  (order) ,  and edit the query to return the fields order by **order**. But as you can imagine it's not a solution...

Regards

akazorg:
Making BREAD items sortable, this changes fields order on views Browse/Edit/Add.

Fixes #994.

### DIFF
--- a/resources/views/tools/database/edit-add-bread.blade.php
+++ b/resources/views/tools/database/edit-add-bread.blade.php
@@ -135,105 +135,110 @@
                     </div><!-- .panel -->
 
                     <div class="panel panel-primary panel-bordered">
-
                         <div class="panel-heading">
                             <h3 class="panel-title">Edit the rows for the {{ $table }} table below:</h3>
                         </div>
 
                         <div class="panel-body">
-                            <table id="users" class="table table-hover">
-                                <thead>
-                                    <tr>
-                                        <th>Field</th>
-                                        <th>Field Info</th>
-                                        <th>Visibility</th>
-                                        <th>Input Type</th>
-                                        <th>Display Name</th>
-                                        <th>Optional Details</th>
-                                    </tr>
-                                </thead>
-                                <tbody>
-                                @foreach($fieldOptions as $data)
+                            <div class="row fake-table-hd">
+                                <div class="col-xs-2">Field</div>
+                                <div class="col-xs-2">Visibility</div>
+                                <div class="col-xs-2">Input Type</div>
+                                <div class="col-xs-2">Display Name</div>
+                                <div class="col-xs-4">Optional Details</div>
+                            </div>
 
-                                    @if(isset($dataType->id))
-                                        <?php $dataRow = TCG\Voyager\Models\DataRow::where('data_type_id', '=',
-                                                $dataType->id)->where('field', '=', $data['field'])->first(); ?>
-                                    @endif
-                                    <tr>
-                                        <td><h4><strong>{{ $data['field'] }}</strong></h4></td>
-                                        <td>
-                                            <strong>Type:</strong> <span>{{ $data['type'] }}</span><br/>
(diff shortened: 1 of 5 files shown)


resources/views/tools/database/edit-add-bread.blade.php
```diff
@@ -270,6 +275,23 @@
                 });
             @endif
 
+            /**
+             * Reorder items
+             */
+            $('#bread-items').sortable({
```
akazorg, line 281: I ended up using jQuery, hope it's ok.

Ping @abdgad @marktopper

Rostom22:
How can i change the order of my fields in the edit-add view please ?

akazorg:
Hi @Rostom22 when you edit a BREAD item, you can easily order fields by drag-and-drop rows.
Don't forget to hit save, because this isn't working with AJAX.
If you're using a older version of Voyager, make sure you have the latest assets / migrations.

Good luck!

Rostom22:
Hi, i tried to drag-and-drop but it'sn't working, also in te edit-add view the field appears in different order. thanks for your help.

akazorg:
Hello, when you say it doesn't work, the drag-and-drop UI effect is visible/working?
It seams you're not using the latest version.

Please check if you:
- published Voyager views before, if so, you need to make sure you update them.
- run the migration, if you did, you should see a new field named `order` in `data_types` table.

I suggest you to join us on Slack, it may be faster to help you.

Happy coding :)

Rostom22:
Sorry i'm a begginer and i don't understand very well, i installed voyager with composer 5 days ago following the videos of devdojo. what's the Slack ? Thanks again !

akazorg:
It's ok, we all started somewhere :)
Please keep in mind that you need to know the basis of Laravel, otherwise it's hard to jump on a project like Voyager.

Slack channel is were the community hangs out. You can get support from there.
https://laravel-voyager.slack.com/

src/Models/DataType.php
```diff
@@ -141,7 +142,17 @@ public function fieldOptions()
...
 
-        $fieldOptions = SchemaManager::describeTable($table);
+        // Get ordered BREAD fields
+        $orderedFields = $this->rows()->pluck('field')->toArray();
+
+        $_fieldOptions = SchemaManager::describeTable($table)->toArray();
+
+        $fieldOptions = [];
+        $f_size = count($orderedFields);
+        for ($i = 0; $i < $f_size; $i++) {
```
DODMax, line 152: I think we are loosing new fields here (added through DB edit but not yet in data_rows)

akazorg: I'm not sure. Can you test that please?

DODMax: The problem is that we iterate on $orderedFields which do not contains new fields. We should work on $_fieldOptions instead.
I've added a PR for that https://github.com/the-control-group/voyager/pull/1210

Rostom22:
Hello, the drag-and-drop UI effect is not visible, to get the new version i just run composer update ? how can i get an invitation for slack ? thanks.

src/Models/DataType.php
akazorg: Great job. Thank you!!

PR #1169, opened 2017-04-24
Allow to order BREAD fields

DODMax:
This is my attempt at allowing ordering of BREAD fields.
It relies on an `order` column in the `data_rows` table (requires to run migrations).

Most is done in the `edit-add-bread` view (could be improved with drag and drop) with a bit of sorting done in `$dataType->fieldOptions()`. Also added the dataRow object as part of the `fieldOptions` array as we're fetching it there to avoid querying it in the view.

Issues:
- New columns are added at the end instead of their actual place in DB (if added using 'after' in phpmyadmin for example)


DODMax:
Looks like I wasn't the only one working on that :D (#994)

marktopper:
This have already been added at #1165 😄

PR #1176, opened 2017-04-25
This change allows the inclusion of additional CSS and JS assets

SUPERNOVI:
This change allows the inclusion of additional CSS and JS assets to be added to the master.blade.php through the voyager config.

The real benefit here is it removes the need for copying the voyager master.blade.php into your views folder avoiding issues if a future update makes changes to this file.

I hope this is acceptable.

Thanks,
Tim

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -218,4 +218,13 @@ return [
 
     'show_dev_tips' => true, // Show development tip "How To Use:" in Menu and Settings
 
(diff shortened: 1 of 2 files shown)


SUPERNOVI:
Just a comment on the codeclimate errors line exceeding 120 characters.
I had to do this because the formatting when rendered was making the `<link>` be tabbed in twice and I thought that would annoy people.

marktopper:
Ignore CodeClimate for now 👍

PR #1190, opened 2017-04-27
allow number formfield to accept decimals

hotsaucejake:
- Reason(s):
  - the validator on the number formfield does not allow decimals

- Change(s):
  - added step="any" to the number formfield

- Reference(s):
  - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number


fletch3555:
👍 Thanks!

PR #1194, opened 2017-04-28
[fix] Boot User VoyagerServiceProvider

akazorg:
This fixes user boot at VoyagerServiceProvider, when running `php artisan voyager:admin name@domain.tld`. It happens when using another table name for `users`.

It wasn't using the config from `voyager.user.namespace`.


fletch3555:
I'm okay with this if someone else can test it.  @marktopper maybe?

fletch3555:
Has anyone had a chance to test this? @marktopper? @akazorg? @abdgad?

PR #1197, opened 2017-04-28
Enforce sortBy PK id

akazorg:
When using a different name for `id` in a table, we have to enforce that the sortBy is using the PK defined in the model `protected $primaryKey = 'custom_id'`.


fletch3555:
Thanks!

PR #1199, opened 2017-04-28
Added ability to configure datepicker for date fields

DODMax:
**Current behaviour:**
Date picker is up to the browser (`<input type="date"/>`).

**Suggested behaviour:**
- Default: Date picker is up to the browser if available, otherwise fallback to BootstrapDatePicker (as used for timestamps)
- Custom: If `datepicker` is defined on BREAD field's option: Use BootstrapDatePicker with provided parameters

Example BREAD option:
```
{
    "datepicker": {
        "format": "YYYY-MM-DD",
        "defaultDate": "1990-01-01"
    }
}
```

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -125,6 +125,15 @@
 
         $('document').ready(function () {
             $('.toggleswitch').bootstrapToggle();
+            
+            //Init datepicker for date fields if data-datepicker attribute defined
+            //or if browser does not handle date inputs
+            $('.form-group input[type=date]').each(function (idx, elt) {
+                if (elt.type != 'date' || elt.hasAttribute('data-datepicker')) {
+                    elt.type = 'text';
+                    $(elt).datetimepicker($(elt).data('datepicker'));
+                }
+            });
 
             @if ($isModelTranslatable)
                 $('.side-body').multilingual({"editing": true});
--- a/resources/views/formfields/date.blade.php
+++ b/resources/views/formfields/date.blade.php
@@ -1,3 +1,4 @@
 <input type="date" class="form-control" name="{{ $row->field }}"
        placeholder="{{ $row->display_name }}"
-       value="@if(isset($dataTypeContent->{$row->field})){{ gmdate('Y-m-d', strtotime(old($row->field, $dataTypeContent->{$row->field}))) }}@else{{old($row->field)}}@endif">
+       value="@if(isset($dataTypeContent->{$row->field})){{ gmdate('Y-m-d', strtotime(old($row->field, $dataTypeContent->{$row->field}))) }}@else{{old($row->field)}}@endif"
+       @if(isset($options->datepicker))data-datepicker="{!! htmlentities(json_encode($options->datepicker), ENT_QUOTES, 'UTF-8') !!}"@endif>


akazorg:
Nice job!

fletch3555:
I'm okay with this, so long as it's tested.  @akazorg, would you mind testing this?

akazorg:
yes @fletch3555, I'll do it later.

akazorg:
This is good to merge. I've test it.

reficul:
Can I suggest also to add a reference to "locale" and an option to add or remove time?
Something like:
`{

    "datepicker": {

        "format": "YYYY-MM-DD",

        "defaultDate": "1990-01-01",

        "locale": "it",

        "display": "date" //or "datetime" or "time"

    }

}`

acicovic:
Hi @DODMax and thank you for your work.

Is this still relevant? Does it work? no matter what I do I don't seem to be able to change the format of DatePickers (I've used the format like in your first post but also strftime formats as described in #1407). The only option listed by the current documentation is "format" but this doesn't change DatePickers. I am using 1.2. Do you know if I'm doing something wrong or this just doesn't work anymore?

Thanks!

DODMax:
Hi @acicovic ,

I think the code has been removed in v1 but you can still have the feature work by override the date formfield.

That would be done by creating the following file in your `resources/views/vendor/voyager/formfields/date.blade.php`

```php
<input type="{{ isset($row->details->datepicker) ? 'text' : 'date' }}" id="{{ $row->field }}" class="form-control" name="{{ $row->field }}"
       placeholder="{{ $row->display_name }}"
       value="@if(isset($dataTypeContent->{$row->field})){{ gmdate('Y-m-d', strtotime(old($row->field, $dataTypeContent->{$row->field}))) }}@else{{old($row->field)}}@endif">


@section('javascript')
<script>
    $(function () {
        @if(isset($row->details->datepicker))
            //Init datepicker for date fields if data-datepicker attribute defined
            //or if browser does not handle date inputs
            $('input[name={{ $row->field }}]').datetimepicker({!! json_encode($row->details->datepicker) !!});
        @endif
    });
</script>
@append
```

Hope this helps

acicovic:
Hi @DODMax, thank you so much for the code!
I'll look into it.

acicovic:
@DODMax this is user friendlier than the built-in functionality in my opinion.

The concept works but it doesn't translate dates correctly. For example, an input with value of 2019-01-06 will display 20-01-2006 instead of 06-01-2019 for a DD-MM-YYYY format. In other words, it doesn't understand the format of the source date.

Do you know if there's a setting or function we can pass so the DatePicker will understand the correct date, as well as save it correctly?

In the past I've used $("input[name='date_from']").data('datepicker').getFormattedDate('yyyy-mm-dd') in similar scenarios but this would be only for saving and I would not know where to integrate this within Voyager

Any input is greatly appreciated! Thanks!

DODMax:
If I remember well you have to pass the source format as an extra format to datetimepicker. Your database field configuration would look like
```
{
    "datepicker": {
        "format": "DD-MMM-YYYY",
        "extraFormats": [
            "YYYY-MM-DD"
        ]
    }
}
```

acicovic:
Hi, @DODMax, unfortunately this code tries to insert the dates with the latter format so it results in an exception. However I don't want to bother you with this any further. Thank you for your time and your prompt response which were greatly appreciated!

DODMax:
Indeed. I chose to solve that by using a setter on my model:
```
public function setBirthdateAttribute($value)
    {
        $this->attributes['birthdate'] = Carbon::parse($value);
    }
```

PR #1213, opened 2017-05-01
Scroll first error into view after ajax validation

DODMax:
**Current behavior**
On form submit, in case of validation error problematic fields are highlighted

**Suggested behavior**
Current behavior + scroll first error into view (if any)


AdrianKuriata:
Nice idea, i didnt think about add it, thanks for this, i hope @marktopper add it, this is good improvement.

PR #1214, opened 2017-05-01
fixed the resize width and height missing for multiple image field type

imknight:


fletch3555:
Where did the 1800 and null values come from? Did you make them up because they worked for you, or calculate them somehow?

imknight:
those options are there for single image, whoever add on the code for multi image section didn't copy that part which is the reason causing the error.

PR #1219, opened 2017-05-02
settings page logs every time options editor content changes

mpge:
This spams console.log - not sure this is required.


fletch3555:
You are 100% correct.  That definitely shouldn't be there.

PR #1228, opened 2017-05-05
updated FILE type browse view and added "where" advanced BREAD option

gkmk:
- updated FILE type browse view
- added advanced option "where" for select_dropdown
sample usage
`{
    "relationship": {
        "key": "id",
        "label": "name",
        "page_slug": "admin/users",
        "where": [
            "role_id",
            2
        ]
    }
}`


marktopper:
Great improvement!

Issue #1172, opened 2017-04-24
Checking permission returns 'true' if permission doesn't exist


hotsaucejake:
- Laravel Version: 5.4.19
- Voyager Version: 0.11.9
- PHP Version: 7.1
- Database Driver & Version:

### Description:
Created a custom controller for one of my routes.  In testing, I checked to see if it had a permission without actually creating that permission in the DB.  **It always returns 'true' if the permission does not exist.**  However, when I create the permission in the DB and don't assign it to the user then it returns false (as expected).

### Steps To Reproduce:

My ````routes/web.php```` file has this:

````php
Route::get('/home', 'HomeController@index');
````

My ````app/Http/Controllers/HomeController.php```` has the following:

````php
namespace App\Http\Controllers;

use Illuminate\Http\Request;
use TCG\Voyager\Facades\Voyager;

class HomeController extends Controller
{
    /**
     * Create a new controller instance.
     *
     * @return void
     */
    public function __construct()
    {
        $this->middleware('auth');
    }

    /**
     * Show the application dashboard.
     *
     * @return \Illuminate\Http\Response
     */
    public function index()
    {
      dd(Voyager::can('blahblahblah'));
      // return view('home');

    }
}
````

**This will return true.**  However, if I create a permission and don't assign it to any users, it returns false (as expected).  Only when the permission does not exist it returns true.

akazorg:
Thanks for reporting that, this is a serious security issue.

marktopper:
This is fixed in #1285 for `v1.0-pre-release` since this is a breaking change.

Issue #1168, opened 2017-04-23
Menu shows media/tools/settings items for user wihout permission


pjeutr:
- Laravel Version: 5.4.19
- Voyager Version: 0.11.9
- PHP Version: 5.6.2
- Database Driver & Version:

### Description:
After I updated from 0.10 to 0.11, a custom role I created shows the media/tools/settings in the admin menu. In the created role I only ticked _Browse Admin_ to make the dashboard viewable. 
A click gives me an UnauthorizedHttpException so the system knows the user isn't authorized
Are more people seeing this?

![screen shot 2017-04-23 at 23 33 04](https://cloud.githubusercontent.com/assets/611968/25317661/4be2238e-287d-11e7-9ad3-d43c2b6e4440.png)
![screen shot 2017-04-23 at 23 34 15](https://cloud.githubusercontent.com/assets/611968/25317666/659d2bfc-287d-11e7-94b7-32225e833fbb.png)

### Steps To Reproduce:
Add a role, only tick _Browse Admin. 
Add the role to an user and login.

nunomsh:
It shows, but wont open. I had the same problem, and the solution was to create different menus for each role, and load the menu according the user permissions. 
Something like

```
switch (Auth::user()->role_id) 
{
            case 1:
                echo menu('admin','admin_menu');
                break;
            case 3:
                echo menu('manager','admin_menu');
                break;
            case 5:
                echo menu('seller','admin_menu');
                break;
            default:
            echo menu('seller','admin_menu');
}
```

pjeutr:
I dove a little deeper and removed 
`&& is_null($value->table_name) `
in admin_menu.blade.php:64
Now things work for me again, but it's in Voyager code...

Can't really figure out how it should work, But there seems to be something wrong with permissions, maybe related to https://github.com/the-control-group/voyager/issues/1172

marktopper:
This is fixed in #1285 for `v1.0-pre-release` since this is a breaking change.

PR #1240, opened 2017-05-10
Fix missing permission returning true

emptynick:
Permission returned true if the permission does not exist as mentioned
in #1172 and #1168.
Also inversed the check so function can() only returns true if the user
!= null and the permission is assigned to this user - not vise versa

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -61,7 +61,7 @@
                             @endif
 
                             @foreach($dataTypeRows as $row)
-                                <div class="form-group @if($row->type == 'hidden') hidden @endif col-md-{{ $row->col_width }}">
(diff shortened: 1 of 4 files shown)


marktopper:
The idea is that if a permission does not exists in the system, everyone can actually do it, else even the admin wouldn't have the permission.
I however am open for changing this behavior upon request.

I am opening this for discussion.

marktopper:
I think the best solution would be to throw an exception if the permission does not exists.
And then merge that to v1.

Thoughts @emptynick.

akazorg:
I like throwing an exception.

marktopper:
@akazorg. Me too, I think we should do this for v1.

akazorg:
I've sent PR #1285 for fixing this.

marktopper:
Merged into `v1.0-pre-release`.

PR #1246, opened 2017-05-12
Field sort not suitable

hysdan:
I change some codes for oracle db. It's work well. 
There is one problem is field sort not suitable. How to change field sort?
p1
p2
p3
p4

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -37,60 +37,60 @@
                                         <td>
                                             <?php $options = json_decode($row->details); ?>
                                             @if($row->type == 'image')
-                                                <img src="@if( strpos($data->{$row->field}, 'http://') === false && strpos($data->{$row->field}, 'https://') === false){{ Voyager::image( $data->{$row->field} ) }}@else{{ $data->{$row->field} }}@endif" style="width:100px">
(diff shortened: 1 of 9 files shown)


resources/views/bread/browse.blade.php
```diff
@@ -37,60 +37,60 @@
                                         <td>
                                             <?php $options = json_decode($row->details); ?>
                                             @if($row->type == 'image')
-                                                <img src="@if( strpos($data->{$row->field}, 'http://') === false && strpos($data->{$row->field}, 'https://') === false){{ Voyager::image( $data->{$row->field} ) }}@else{{ $data->{$row->field} }}@endif" style="width:100px">
+                                                <img src="@if( strpos($data->{strtolower(strtolower($row->field))}, 'http://') === false && strpos($data->{strtolower($row->field)}, 'https://') === false){{ Voyager::image( $data->{strtolower($row->field)} ) }}@else{{ $data->{strtolower($row->field)} }}@endif" style="width:100px">
```
fletch3555, line 40: There is absolutely no reason to `strtolower()` that twice, please fix it

fletch3555:
I'm not sure I understand the issue.  Looks like the issue is that your column names are capitalized, which is not an OracleDB requirement (as far as I know).  I'm not against using `strtolower()` for comparison purposes wherever that may be needed, but not for display.

resources/views/bread/browse.blade.php
hysdan: @ Fletch3555
This is an Oracle db issue.
Field names in the table are in upper case, but in the eloquent model they are lower case.
I just try to change some code to avoid this issue.
Thank you very much.

PR #1250, opened 2017-05-14
BREAD field-width feature

emptynick:
Enables the user to set the width of a form-field in the add/edit page.
To set the width write in the optional details field:
`{
    "width": "half"
}`

Possible values are 
- full
- half
- onefourth
- threefirth
- onethird
- twothird

Default is full


tnylea:
Closing this PR since it is being implemented in version 1.0 :)

PR #1256, opened 2017-05-15
Formfield width in BREAD with percentage

emptynick:


tnylea:
Thanks, waiting for the builds to finish.

tnylea:
@emptynick Hey man, just want to let you know that I did change it back to the grid system for the width. You were right on that part to have the 12 grid column instead of percentages, I ran into a few problems with the percentage issue.

Anyway, I've already added it to our version 1.0 release docs here: https://laravelvoyager.com/docs/1.0/#core-concepts-bread-display-options

Thanks again and sorry I didn't end up sticking with your first solution.

Hope you're doing well. Talk to you soon.

PR #1257, opened 2017-05-16
Clean-Up

emptynick:


tnylea:
Sweet! Looks good. Thanks @emptynick

fletch3555:
Why was column_width removed? I don't know why it was added in the first place, but I want it documented why it was removed, since this PR has zero information in it

PR #1275, opened 2017-05-24
Cache settings in memory to avoid querying multiple times

DODMax:
**Current behavior:**
The DB is queried each time a setting is retrieved with `Voyager::setting()`

**Proposed behavior**
Locally cache settings in a private array on the Voyager object so that their are each queried only once.

### DIFF
--- a/src/Voyager.php
+++ b/src/Voyager.php
@@ -52,7 +52,7 @@ class Voyager
         'User'       => User::class,
     ];
 
-    private $setting_cache = [];
+    public $setting_cache = null;
 
     public function __construct()
     {
@@ -138,13 +138,11 @@ class Voyager
 
     public function setting($key, $default = null)
     {
-        if (! isset($this->setting_cache[$key]))
-        {
-            $setting = Setting::where('key', '=', $key)->first();
-            $this->setting_cache[$key] = isset($setting->id) ? $setting->value : $default;
+        if ($this->setting_cache === null) {
+            $this->setting_cache = Setting::pluck('value', 'key');
         }
 
-        return $this->setting_cache[$key];
+        return $this->setting_cache->get($key) ?: $default;
     }
 
     public function image($file, $default = '')
--- a/src/Voyager.php
+++ b/src/Voyager.php
@@ -52,6 +52,8 @@ class Voyager
         'User'       => User::class,
     ];
 
+    private $setting_cache = [];
+
     public function __construct()
     {
         $this->filesystem = app(Filesystem::class);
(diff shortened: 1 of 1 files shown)


akazorg:
Hi @DODMax. I'm ok with that.
I just don't see any handler for refresh the cache system when we update the database. Did I missed something or you forgot to handle it?

DODMax:
Hi @akazorg !

So yeah at first I thought of using one of Laravel's cache providers but it seems the only one available out of the box is file caching, which I guess for single values like settings is as fast as DB (if not slower).
So the title is a bit misleading it's not a proper cache, the only thing it does is keeping the setting's value in memory while building the page, but it will still be queried each time you request a page (but at least once per setting not every time).

I also thought of caching the whole setting table in memory in the same way to avoid querying each setting individually. I think it would be ok in most cases (~32KB) but I'm a bit worried if some users decided to store large object in their settings, like let's say a very very long *terms & conditions* text..

akazorg:
Hello @DODMax :)

I thought the cache system would work on memory.
It doesn't make sense to store long text like `terms & conditions` in settings.

Please let me know your thoughts on this.

»»»» edited »»»»
I was thinking on the cache system, it would be interesting if we add it as a BREAD option, this way we can make it available on every BREAD item.

Spatie wrote a package related to this: [laravel-responsecache](https://github.com/spatie/laravel-responsecache), do you know it? their work is quite inspiring. I would definitely dig on their code and write our implementation.

I'm not sure what others have to say on this.
Ping @marktopper, @fletch3555, @abdgad, please provide your feedback.

DODMax:
Hi,

I agree that a page caching system could be useful for blog like websites, however at the moment it doesn't serve my use case so I won't have time to implement such feature.

My current problem is that I use settings to store display settings (such as date format) and that if I use it in a loop the setting is queried **every time** (eg. if I display 100 rows the same setting with the same value is queried 100 times).

Currently displaying any Voyager page makes at least 5 queries to the *settings* table so in my opinion we should store all the settings in memory on the first call to `Voyager::settings()` then read from memory. But yeah it might be a problem for users with a huge *settings* table

image

Edit: @akazorg 
Just to make sure we are on the same page: since PHP is stateless the settings would be stored in memory just for the time of the request anyway, so no need to refresh anything it will just happen for every request.

AdrianKuriata:
So you should do global variable or aliast which will be load in every request like a Auth::user(), it this case all settings could be load in one time and just display with Setting alias.

akazorg:
Hello.

@DODMax, indeed, implementing a cache system requires some work, but on another hand, writing this cache "only for settings", imho will contribute, somehow, on bloating Voyagers code.
I have serious doubts whether we should go with this, or wait for a global cache system that deals with any BREAD item.

If we decided to go, @AdrianKuriata, is right, we should use a single db query and load all settings.
Another issue, like you mentioned, if a user store a lot of data, total memory used may be an issue, perhaps we could add a config option for toggling this feature?

I'll ask for others feedback.
Thanks guys!

DODMax:
Ok so I updated the PR to query all the settings at once and keep them in memory while the request is running.  This save *at least* 4 DB queries for any Voyager page.

Regarding the BREAD caching discussion I think it's a completely different topic as here we are not talking about caching **between** requests, we just avoid querying the DB each time `Voyager::setting()` is called.

marktopper:
Thanks @DODMax

akazorg:
I missed your reply @DODMax. Yes, you're right, we should handle a global cache system on different issue.
Thanks for the PR.

marktopper:
Indeed we should get a global cache system, but I think we will handle that later.
For now this is fine 👍

PR #1284, opened 2017-05-29
Fix Delete Images

akazorg:
When deleting data (using BREAD), images were not being removed from server.


marktopper:
Great work 👍

PR #1285, opened 2017-05-29
Fix Missing Permission

akazorg:
Fixes missing permission, now returns an Exception.

Related:
- #1168 (not sure if gets fixed)
- #1172 
- #1240 (PR)


marktopper:
Thanks 🎉
Rebased this to `v1.0-pre-release` since this is a breaking change.

PR #1293, opened 2017-05-31
Redirect to index after BREAD insert/update

akazorg:
After using BREAD Insert/Update, it should redirect to the index view, and not to the edit.
Thanks @hotsaucejake for pointing this out.


hotsaucejake:
Ack!

akazorg:
I have no idea why Travis is failing...

turtuga:
Why update redirect to index ? I like to see my data but ok..

hotsaucejake:
@turtuga it's a better user experience.  Do you want to edit your entry as soon as you enter it?  It's more intuitive to redirect to the index or to a create new screen.  If you want to submit several entries, it's 2 extra clicks to "add new."

akazorg:
hello @turtuga, why would you need to update information you have just added/edited?

turtuga:
When I edit only I like view my data but is ok like I said

fletch3555:
TravisCI is failing because of the tests.  When redirecting to the index view, you no longer need to pass in the user ID.  Since the route doesn't expect it, it's appending it as a querystring, which is making the URL `/admin/users?id=1` instead of `/admin/users`.

@akazorg, please remove the `id` parameter from the redirects you modified in VoyagerBreadController (https://github.com/the-control-group/voyager/pull/1293/files#diff-ddbbbb24b4631fc07f7236273ceb9739R180 and https://github.com/the-control-group/voyager/pull/1293/files#diff-ddbbbb24b4631fc07f7236273ceb9739R247), then this should be good to go.

akazorg:
Thanks @fletch3555 I'll do it.

akazorg:
The [error log](https://travis-ci.org/the-control-group/voyager/jobs/241025488) on TravisCI is different but still happening. Any idea of that @fletch3555? Thanks

donmbelembe:
Actually it should return to the index, because if you are entering multiple row one by one you need to go first to index then click new item, long process

Elyahou:
@akazorg Did you removed some translation (Successfully Updated) in this PR ?

marktopper:
@Elyahou, thanks for seeing this.
This have been fixed in #1431 🎉

akazorg:
Thank you @Elyahou.
It's fixed.

PR #1299, opened 2017-06-02
Set laravel/framework required at composer.json

akazorg:
The `laravel/framework` is now required at composer.json. Some cleanup done.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,22 +15,21 @@
         }
     ],
     "require": {
-        "illuminate/support": "~5.3.0|~5.4.0",
-        "intervention/image": "^2.3",
-        "doctrine/dbal": "^2.5",
-        "larapack/doctrine-support": "~0.1.0",
         "arrilot/laravel-widgets": "^3.7",
+        "doctrine/dbal": "^2.5",
+        "intervention/image": "^2.3",
+        "larapack/doctrine-support": "~0.1.0",
+        "laravel/framework": "~5.3.29|5.4.*",
         "league/flysystem": "~1.0.35"
     },
     "require-dev": {
-        "phpunit/phpcov": "~3.0",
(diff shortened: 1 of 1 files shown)


johnss:
How about instead of deleting it change it to `laravel/laravel` so `tcg/voyager` can be installed without installing laravel first.

akazorg:
What do you suggest @laravelish?
We need to define the dependency for laravel version "~5.3.29|5.4.*".

fletch3555:
We already have illuminate/support which generally meets this need.  Not sure if laravel/framework is better or not.

ghost:
@akazorg well you don't need it because laravel already requires it https://github.com/laravel/laravel/blob/master/composer.json

akazorg:
@fletch3555 How do you define which versions of Laravel are supported by Voyager?

@laravelish, that didn't answer the question. **Any suggestion?**

akazorg:
@marktopper @fletch3555, your feedback is required, please.

marktopper:
Well, it's not required to `require` the `laravel/laravel` package, however I think it's a good idea to do so.
Just my opinion.

fletch3555:
@akazorg, the `illuminate/support` package follows Laravel versioning since it's one of the core packages within `laravel/framework`, so as I said before, we technically already have the version requirements you mentioned.

Now, as for whether `laravel/framework` is a better dependency to list, I don't really have an opinion.  Might be worth asking somewhere more official (larachat? laracasts? the laravel github repo?)

marktopper:
The mod of Larachat says that we should never depend on the framework as your package will be installed somewhere else which will have framework as a depdency.
Instead the best way to saveguard the required versions of Laravel Framework is to depend on an Illuminate component of that version.

akazorg:
Great info @marktopper. Now we know it :)

PR #1315, opened 2017-06-06
[Modify]It will not copy Voyager::routes() into routes/web.php, if th…

rushairer:
It will not copy Voyager::routes() into routes/web.php, if that already copied in the file when reinstall voyager.


akazorg:
Great job @rushairer, finally someone does it!

marktopper:
Great job @rushairer

PR #1335, opened 2017-06-13
fix optional relationship on Browse

hugosvent:
When having nullable relationship, it returns undefined object.

So I check for optional relationship,
if null, just show nothing


marktopper:
Thanks! Great work.

I just fixed the merge conflicts, will merge it once Travis is done testing.

marktopper:
It seems like there are some issues with this.

Can you please merge in the origin?
https://github.com/thishugo/voyager/pull/1

PR #1341, opened 2017-06-14
Added field name in "label for" and "input id"

fernandonicacio:
[Forms] When you enter more than one radio field of the same value, a conflict occurs because they have the same id.

Example:

Field: Display
Options: Yes (id: option-Yes) | No (id: option-No)

Field: Featured
Options: Yes (id: option-Yes) | No (id: option-No)


marktopper:
Thanks! 👍

Issue #1331, opened 2017-06-12
Rich text editor doesn't allow all content to be deleted.


gavinj1999:
- Laravel Version: 5.4.25
- Voyager Version: 0.11.12
- PHP Version:5.6.3
- Database Driver & Version: 10.0.30-MariaDB,

Rich text editor won't save empty content

### Steps To Reproduce:

Created database table
Added Bread
configured form fields

I have a form with a rich text editor, I can add / edit the content but, if I delete all the content and save, the content remains.

Any idea?

Thanks

akazorg:
Hi @gavinj1999,
That is a known bug.

gavinj1999:
thanks for the reply, any workarounds?

AdrianKuriata:
Try to set validation for this field with sometimes and nullable parameters. "sometimes|nullable"

akazorg:
Does that work @AdrianKuriata?
I don't know any workaround atm @gavinj1999 :/ Please let us know if Adrian solution worked out.

AdrianKuriata:
@akazorg i hope it will work, i doing it in laravel when i want save null for any field. When you try save empty field with laravel, into $request this field will not figure so he cant save empty field. When you add sometimes and nullable, sometimes will be present and can be empty and will be only validating when you put something into field. You can set present, this will let you be empty but if ypu add parameter something like min:3 this will be validate so better solution is set sometimes|nullable if your field is not required. Check laravel docs. 

[present](https://laravel.com/docs/5.4/validation#rule-present)
[sometimes](https://laravel.com/docs/5.4/validation#conditionally-adding-rules)

Better check documentation, i bad explained solution, but i did in my projects same i explained. If you want save null, you have to add nullable attribute.

akazorg:
Thank's Andrian, will check that PR asap.

PR #1352, opened 2017-06-16
Fix for saving field like a null

AdrianKuriata:
I fixed possibility to saving field like a null. I thougth rules validation like "`sometimes|nullable`" will be enough but wasn't. When we tried to save empty field, it was not active into $date variable so we need remove this part of code for let us saving field like a empty(null) value. I tested it and i don't see any issues, but if someone could test it again, i will be appreciated.

@akazorg I tested issue with rules and it was not worked so i did a PR for fix it.
#1331

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -61,10 +61,6 @@ abstract class Controller extends BaseController
             }
 
             if (is_null($content)) {
-                // Only set the content back to the previous value when there is really now input for this field
-                if (is_null($request->input($row->field)) && isset($data->{$row->field})) {
-                    $content = $data->{$row->field};
-                }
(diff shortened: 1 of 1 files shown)


akazorg:
Great job, I'll test that out ASAP!!

akazorg:
Merged

MujtabaFR:
Hello,
When performing the suggested change (deleting the few lines).. I found an issue with the **image field**
If I'm editing a post which already has an image uploaded for it. When saving it (without changing the image field) .. The image becomes null and will be removed from the post and the database

Guess it will be the same issue with the **File field**, but I didn't test it yet

akazorg:
The feature for delete an existing image/file is currently missing. We need to place a delete button and handle that procedure.

PR #1360, opened 2017-06-18
Add Arabic language file

mbdwey:
It's more than 90%


AdrianKuriata:
Guys why you don't wait for official information about translations. They will group it so it wont be 'generic_email' but

```
'generic' => [
'email'
];
```

Give them some time for do translations...

marktopper:
We use groups for translations.

Please see https://github.com/the-control-group/voyager/blob/i18n/publishable/lang/en/voyager.php

PR #1383, opened 2017-06-22
$row is better then $rows

wuwx:


akazorg:
Thanks for the PR @wuwx!

PR #1387, opened 2017-06-22
Fix Migration Date

akazorg:
[This migration](https://github.com/the-control-group/voyager/blob/master/publishable/database/migrations/2017_11_04_000000_alter_post_nullable_fields_table.php) was dated for November 2017, are you playing back to the future? :D

Migration was renamed to: `2017_06_20_000000_alter_post_nullable_fields_table.php`.

A duplicated migration may happen on the database, however, if the migration runs twice it's harmless.


akazorg:
This will be fixed using the future Upgrade System.

PR #1393, opened 2017-06-24
Remove dummy text from menu builder

akazorg:
Clear useless string I left at menu `builder.blade.php`.


fletch3555:
I wholeheartedly approve of this change.

PR #1395, opened 2017-06-25
Two add_timestamps buttons on Database Helpers

gabrieljaime:
Hello, I see that in the helpers buttons of the database appears the add_timestamps button repeated instead of the softdelete button.

Regards


akazorg:
Thank you!

PR #1409, opened 2017-06-29
Add additional where option dropdown combo

gabrieljaime:
Hello, I want to make a change in the conditions of the  dropdown for the "where" option taking into account the 3 options that gives datatable component.
1 variable looking for the id directly 

> ```
> {
>     "relationship": {
>         "key": "id",
>         "label": "name",
>         "where": [
>             "1"
>         ]
>     }
> }
> ```

```
$model= DB::table('table')
                 ->where('id',  1)
                 ->get();
```

2 variables to compare two fields with "=" (Existing option)

> ```
> {
>     "relationship": {
>         "key": "id",
>         "label": "name",
>         "where": [
>             "parent_id",
>             "1"
>         ]
>     }
> }
> ```
```
$model= DB::table('table')
                 ->where('parent_id',  1)
                 ->get();
```
 and another option of 3 variables to add the type of condition.

> ```
> {
>     "relationship": {
>         "key": "id",
>         "label": "name",
>         "where": [
>             "name",
>             "like"
>             "Star%"
>         ]
>     }
> }
> ```
> 
$model= DB::table('table')
                 ->where('name', 'like', 'Star%')
                 ->get();
```

Thank!!!

### DIFF
--- a/resources/views/formfields/select_dropdown.blade.php
+++ b/resources/views/formfields/select_dropdown.blade.php
@@ -29,11 +29,27 @@
                 $relationshipOptions = $dataTypeContent->$relationshipListMethod();
             } else {
                 $relationshipClass = $dataTypeContent->{camel_case($row->field)}()->getRelated();
-                if (isset($options->relationship->where)) {
-                    $relationshipOptions = $relationshipClass::where(
-                        $options->relationship->where[0],
-                        $options->relationship->where[1]
-                    )->get();
+               if (isset($options->relationship->where)) {
+                    if (count($options->relationship->where) == 2) {
+                        $relationshipOptions = $relationshipClass::where(
+                                $options->relationship->where[0],
+                                $options->relationship->where[1]
+                        )->get();
+                    } else {
+                        if (count($options->relationship->where) == 3) {
+                            $relationshipOptions = $relationshipClass::where(
+                                    $options->relationship->where[0],
+                                    $options->relationship->where[1],
+                                    $options->relationship->where[2]
+                            )->get();
+                        } else {
+                            $relationshipOptions = $relationshipClass::where(
+                                    'id',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by marktopper** [comment]
Why not just do this in the relation?

```
class User extends Model
{
    public function posts()
    {
        return $this->hasMany(Post::class)->where('status', 'published');
    }
}
```

NOTE: Ignore CodeClimate

**Comment by gabrieljaime** [comment]
Hi, I thought on it. But if the functionality (the where filter) is already enabled why it don't give it complete with the different variables? Because although it might filter the relationship maybe just want to show in combo a group of possibilities. 
I think that more is better. But is only a suggestion. 
Regards

**Comment by marktopper** [comment]
Thanks for your comment, will get back to you when I have talked to the other guys. 😄

**Comment by marktopper** [comment]
Hey @gabrieljaime.

It will become kinda a lot for Voyager if we should support where statements, since there will then also be a need for doing where's and orWhere's combined and in subqueries.

So instead of making this complicated, we have agreed that it would be much nicer to have a set of scopes to run through.

Let me give you an example, lets say we have the Post model which includes some scopes.

```
class Post extends Model
{
    public function scopePublished($query)
    {
        return $query->where('status', 'published');
    }
}
```

Then the JSON for this could look something like this:
```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "scopes": [
            "published"
        ]
    }
}
```

gabrieljaime:
Thank you very much! I think it's an excellent idea !!! Better than my original proposal. 

Great job. 

Regards

fletch3555:
I'm going to decline this PR for now.  @gabrieljaime, you're welcome to modify your work to use scopes instead and resubmit the PR, or we'll get to it eventually.

Thank you for the original suggestion!

PR #1416, opened 2017-06-29
Change reference to Voyager Font Class

gabrieljaime:
I change the reference to Voyager Font Class from fonts/voyager/icons-reference.html to fonts/icons-reference.html, and de icons fonts directory.

I do not know if it is the right place to make these changes, but now it works correctly.

Regards


akazorg:
Thank you!

PR #1417, opened 2017-06-30
Update login.blade.php and Spanish translation

gabrieljaime:
Change the call to loginin to match the descriptions


akazorg:
Nice job. Thank you!

PR #1418, opened 2017-06-30
I18n

stone512:


akazorg:
Thank you @stone512.
I will close because it will be merged with another PR.

PR #1421, opened 2017-07-01
Can x

cotrudel:
Adds a layer for custom permissions in models in the form "canX", in the case of complex permissions on the model.


fletch3555:
Reviewed this on slack and decided on a different approach.  Closing this for now.

PR #1425, opened 2017-07-01
Add support for dynamically changing view parameters through events

marktopper:
### Load a view with support for dynamically changing view parameters

```
Voyager::view('voyager::hello', [
    'name' => 'Mark',
]);
```

### Add event to do something when that specific view is loaded

```
Voyager::onLoadingView('voyager::hello', function ($name, array $parameters) {
    // This triggers when the "voyager::hello" view is loaded
});
```

### Change view and/or parameters

```
Voyager::onLoadingView('voyager::hello', function (&$name, array &$parameters) {
    $name = 'something::else';
    $parameters['foo'] = 'bar';
});
```

/ping @akazorg

### DIFF
--- a/src/Http/Controllers/VoyagerAuthController.php
+++ b/src/Http/Controllers/VoyagerAuthController.php
@@ -4,6 +4,7 @@ namespace TCG\Voyager\Http\Controllers;
 
 use Illuminate\Foundation\Auth\AuthenticatesUsers;
 use Illuminate\Http\Request;
+use TCG\Voyager\Facades\Voyager;
 
 class VoyagerAuthController extends Controller
 {
@@ -11,7 +12,7 @@ class VoyagerAuthController extends Controller
 
     public function login()
     {
-        return view('voyager::login');
+        return Voyager::view('voyager::login');
     }
 
     public function postLogin(Request $request)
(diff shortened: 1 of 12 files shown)


fletch3555:
@marktopper, what prompted this? It's somewhat contrary to "the Laravel way"

akazorg:
Great job Mark, testing now.

@fletch3555 this will allow us to load custom views without the need of extending Voyager Controllers. I'm not sure what is the "Laravel way" of doing this, but events sounds a nice approach.
Would you handle it differently?
Please, share your ideas with us.

fletch3555:
I guess I don't understand the need.  What "custom views" would you be loading that would require overriding the controller?

akazorg:
For example, from VoyagerBreadController, at the show() method we have:
```
    public function show(Request $request, $id)
    {
        $slug = $this->getSlug($request);
        ....
        $view = 'voyager::bread.read';

        if (view()->exists("voyager::$slug.read")) {
            $view = "voyager::$slug.read";
        }

        return view($view, compact('dataType', 'dataTypeContent', 'isModelTranslatable'));
}
```
If we want to use a view that doesn't match the BREAD slug, it would require to overwrite the Controller.

PR #1433, opened 2017-07-04
Spatial column type support using a Google map 

lampdev:
# Setup:

- Install Voyager. 

- Create a table with at least one column with any of Geometry datatype (currently only POINT supported). Select "Generate Model" for the table

- Create BREAD for the newly created table. For the Geometry column select 'coordinates' field type. Don't forget to allow all actions for the column under Main menu -> Roles.

- Add the following trait to your model class:

```
use TCG\Voyager\Traits\Spatial;

class Mymodel extends Model
{
    protected $spatial = ['coords'];

    use Spatial;
}
```

where `coords` is the name of your Geometry field you added on step 2. 

- Add Google maps config section in your .env file with the following keys:

```
GOOGLE_MAPS_KEY=AIXXXXXXXXXXXXXXXXXXXXXXXXXqjrA
GOOGLE_MAPS_DEFAULT_CENTER_LAT=50.3083568
GOOGLE_MAPS_DEFAULT_CENTER_LNG=8.2091757
GOOGLE_MAPS_DEFAULT_ZOOM=11
```

obviously, replacing center coordinates and the zoom level with the values for your needs

# Browse:

Navigate to the browse url of your BREAD entity , you will see the coordinates in POINT() constructor in the corresponding column

# Add/Edit:

if you click Edit link of any instance you will see a Google map with a draggable marker , you can drag and release and click Save and the new coordinates will be stored in the database

# Read:

if you click View link of any instance you will see a Google map with a non-draggable marker. Quite similar to the Edit portion, but you can't move it and save

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -228,12 +228,12 @@ return [
     ],
 
     'googlemaps' => [
-        'key' => env('GOOGLE_MAPS_KEY'),
+        'key'    => env('GOOGLE_MAPS_KEY'),
         'center' => [
             'lat' => env('GOOGLE_MAPS_DEFAULT_CENTER_LAT'),
             'lng' => env('GOOGLE_MAPS_DEFAULT_CENTER_LNG'),
         ],
         'zoom' => env('GOOGLE_MAPS_DEFAULT_ZOOM')
-    ]
+    ],
 
 ];
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -230,10 +230,10 @@ return [
     'googlemaps' => [
         'key'    => env('GOOGLE_MAPS_KEY'),
         'center' => [
-            'lat' => env('GOOGLE_MAPS_DEFAULT_CENTER_LAT'),
-            'lng' => env('GOOGLE_MAPS_DEFAULT_CENTER_LNG'),
+            'lat' => '',
+            'lng' => '',
         ],
-        'zoom' => env('GOOGLE_MAPS_DEFAULT_ZOOM'),
+        'zoom' => '',
     ],
 
 ];
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -227,4 +227,13 @@ return [
         //'js/custom.js',
     ],
 
+    'googlemaps' => [
+        'key' => env('GOOGLE_MAPS_KEY'),
+        'center' => [
+            'lat' => env('GOOGLE_MAPS_DEFAULT_CENTER_LAT'),
+            'lng' => env('GOOGLE_MAPS_DEFAULT_CENTER_LNG'),
+        ],
+        'zoom' => env('GOOGLE_MAPS_DEFAULT_ZOOM')
+    ]
+
 ];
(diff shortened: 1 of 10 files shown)


publishable/config/voyager.php
```diff
@@ -227,4 +227,13 @@
         //'js/custom.js',
     ],
 
+    'googlemaps' => [
+        'key'    => env('GOOGLE_MAPS_KEY'),
+        'center' => [
+            'lat' => env('GOOGLE_MAPS_DEFAULT_CENTER_LAT'),
```
fletch3555: I don't think lat/lng/zoom need to pull from `env()`.  key makes perfect sense though

fletch3555:
Great work @lampdev.  I'm not sure we NEED this in Voyager core, but I'm certainly intrigued by it.  I'll let others chime in

fletch3555:
Some concerns mentioned in Slack:

1. This is only supported by MySQL.  We need to make sure it doesn't break under SQLite/PostgreSQL.
2. I mentioned this already in a file comment, but the lat/lng/zoom configuration should be in the config file, not `env()`
3. Looks like Voyager can already interpret MySQL's geometric field types (https://github.com/the-control-group/voyager/tree/master/src/Database/Types/Mysql).  This should make use of that (I haven't looked deeply enough to verify whether it does)

src/FormFields/CoordinatesHandler.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace TCG\Voyager\FormFields;
+
+class CoordinatesHandler extends AbstractHandler
+{
+    protected $supports = [
+        'mysql',
```
lampdev: adding only pg- and mysql as supported RDBMS

lampdev:
1. tested on MySQL 5.6 and postgresql 9.6 + postgis 2.3. Note that Geometry type is only added to postgres with postgis extension. See [here](https://stackoverflow.com/questions/6850500/postgis-installation-type-geometry-does-not-exist). So it throws now an error (core voyager functionality) if you try to add a geometry column with postgres with no postgis installed. The spatial functions in raw SQL queries have been modified to match both MySQL and postgeSQL syntax. 

Also, "coordinates" BREAD field type has been enabled only on 'pgsql' and 'mysql' engines:

```
class CoordinatesHandler extends AbstractHandler
{
    protected $supports = [
        'mysql',
        'pgsql',
    ];
```

2. done:

```
    'googlemaps' => [
        'key'    => env('GOOGLE_MAPS_KEY'),
        'center' => [
            'lat' => '',
            'lng' => '',
        ],
        'zoom' => '',
    ],
```

3. These are SQL DB types, here in this PR we're talking about new BREAD field type called "coordinates"

4. added static google maps image when browsing in BREAD: 

test_static_map

src/FormFields/CoordinatesHandler.php
```diff
@@ -0,0 +1,23 @@
...
+        'pgsql',
```
lampdev, line 9: adding only pg- and mysql as supported RDBMS

tnylea:
@lampdev This is awesome!

Testing it out now before merging the pull request.

Thanks.

tnylea:
@lampdev I just made a few updates and tested it out.

Looks really good. Merging now. Thanks again.

lampdev:
Thanks too for the great product!

PR #1435, opened 2017-07-04
Role hasMany User

wuwx:


fletch3555:
Voyager currently only supports Users having a single Role.

PR #1437, opened 2017-07-06
role hasMany users, not belongsToMany

wuwx:


fletch3555:
Voyager currently only supports Users having a single Role.  Please stop submitting these requests.  This is the 3rd you've submitted in 2 days, and I will keep declining them.

wuwx:
can you have a look conscientious
it's $role->users
not $user->roles
ok ?

PR #1452, opened 2017-07-08
Fix custom primary key in browse bread

viktor42b:
It was impossible to edit or view a model if we were not using the `id` primary key for a model. This pull request should use the default primary key set by Laravel inside the Model class.

A model class with a custom primary key should look like:
```php
class MyModel extends Model
{
  protected $primaryKey = 'myCustomPrimaryKey'; // otherwise simply set to id
}
```

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -99,8 +99,16 @@
                                     @endforeach
                                     <td class="no-sort no-click" id="bread-actions">
                                         @if (Voyager::can('delete_'.$dataType->name))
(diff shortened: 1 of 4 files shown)


fletch3555:
Thank you for this.  I'm okay with it, but can you please verify that there aren't any custom views using this same pattern, and update them if there are?

Thanks!

viktor42b:
@fletch3555 done!

What about code climate warnings about HTML lines that are way over 120 characters?

fletch3555:
You don't have to worry about codeclimate.

tnylea:
@victorboissiere Awesome. Thanks, this looks really good. Merging it in now :)

PR #1458, opened 2017-07-11
Switched dimmer text to use unescaped braces.

dattz:
Escaped braces are basically to escape the data sent out by the user. Here the data is generally from the database like statistics etc.

There are scenarios where you might want to use html tags in dimmers. For example:

- `<strong>` etc tags.
- `<div>` tags to add classes or IDs.
- `<span class="fa fa-abc">` tags for Font Awesome or similar libraries.

Only these scenarios are on the top of my mind right now, there may be more.


tnylea:
Thanks for this PR :)

PR #1464, opened 2017-07-13
Delete the create migrations button

connor11528:
Remove the button since the feature is not implemented

### DIFF
--- a/resources/views/tools/database/vue-components/database-table-editor.blade.php
+++ b/resources/views/tools/database/vue-components/database-table-editor.blade.php
@@ -18,12 +18,7 @@
                 <input type="checkbox" name="create_model" data-toggle="toggle"
                        data-on="{{ __('voyager.generic.yes_please') }}" data-off="{{ __('voyager.generic.no_thanks') }}">
             </div>
-
-            <div class="col-md-3 col-sm-4 col-xs-6">
-                <label for="create_migration">{{ __('voyager.database.create_migration') }}</label><br>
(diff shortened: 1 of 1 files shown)


akazorg:
Thank you @connor11528!
We have decided to comment the button instead of removing it. I had updated your code and merged.

connor11528:
👍

michaelmano:
Any idea on when this feature will be implemented as its currently the only reason we are not using voyager.

akazorg:
Hi @michaelmano, there is no prevision for that feature.
If you really need migrations, you can always write them. I hope this is viable for you.
Happy codding!

michaelmano:
@akazorg yeah no worries. Thanks for the reply. I was more or less worried about if I write my migrations would they show up in voyagers back end or if I'm going to have to write a migration then create a dB with voyager and if any data I write in voyager is migratable. I will test it out with a staging server.

akazorg:
There are a few options:
1. You can write migrations, and skip using the DB manager of Voyager.
2. Use Voyager DB manager, skip using migrations and dump your DB schema.

This will all depend on your workflow and dev requirements.

The thing of migrations is tricky because we have to generate a new migration for every change we do at the DB, and this will require some time to implement.

Just use Voyager and don't loose all the fun :D

fletch3555:
@michaelmano, you could also look into adding this package(https://github.com/Xethron/migrations-generator) and just run it periodically to get all the changes you've created with Voyager's database editor.  It only does full tables, so you may have to edit the migrations after they're generated, but it would certainly save you a bit of typing

Issue #1184, opened 2017-04-26
[Feature]Add option not to rename uploaded files


taiwanleaftea:
Now the voyager renames uploaded images as slug/monthyear/random.ext, which isn't useful. It would be nice to add any option or some way to customize what to do with such names.

BTW. If I want to change this behavior, should I edit Controller.php and ViyagerController.php?

akazorg:
You need to extend `Controller.php` and maybe `VoyagerMediaController.php`.

lancepioch:
Labels: feature

Close (pull request merged)

PR #1467, opened 2017-07-14
Add option to preserve image filename on upload

ghost:
Currently when an image is uploaded, its filename is set to a random 20 character string.

This pull request adds an option to the field details that will preserve the filename of the uploaded file. Since the new behavior is opt-in it will not introduce any breaking changes.

To use, you only need to add the option to the field options like so
```json
{
    "preserveFileUploadName": true
}
```

This addresses issue #1184

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -260,14 +260,14 @@ abstract class Controller extends BaseController
                     $file = $request->file($row->field);
                     $options = json_decode($row->details);
 
-                    $filename = basename($file->getClientOriginalName(), '.' . $file->getClientOriginalExtension());
+                    $filename = basename($file->getClientOriginalName(), '.'.$file->getClientOriginalExtension());
                     $filename_counter = 1;
-                    
+
                     $path = $slug.'/'.date('F').date('Y').'/';
-                    
+
                     // Make sure the filename does not exist, if it does make sure to add a number to the end 1, 2, 3, etc...
-                    while( Storage::disk(config('voyager.storage.disk'))->exists( $path.$filename.'.'.$file->getClientOriginalExtension() ) ) {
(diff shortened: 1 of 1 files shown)


tnylea:
Thanks @david-mk 

Will this overwrite a file with the same name? I can still merge it, but I think it would be cool if we preserved the name by default but appended a random string or a number if the file already exists.

Let me know your thoughts :)

Thanks.

fletch3555:
I'm leaning toward maintaining the file name in the database and keeping the random string on the filesystem... This essentially removes the chance for conflict, and still makes the file accessible by it's original name.  I think this is what we already do (or at least very close to it).  What's the use-case for NEEDING the files to have the same name on the filesystem?

ghost:
I found that in the Media view, its difficult to manage images with random file names. This is especially true when you have multiple thumbnails per image and its not obvious at a glance which thumbnail you're seeing without clicking through each of them. Another case is if you have multiple images that appear similar.

Here is an example where I have the main image and 2 thumbnails. Its not clear without clicking through each of them which one is which.
screen shot 2017-07-28 at 12 50 13 pm

Also, the database does not preserve the filename. It uses the randomly generated name.
screen shot 2017-07-28 at 12 54 50 pm

I also think there's an argument to be made for allowing voyager to give me control of my file names in the filesystem if I want it to, which is why I think its good as an opt in feature and not the default behavior.

However I do take your point that it increases the possibility for file name clashing. So I think a good solution would be to use @tnylea's suggestion and append the random string which would give the best of both worlds.

tnylea:
@david-mk I've added the ability to preserve the origin image name by default. If there is already an image with that name it will add +1 to the end so for the following image name `image.jpg`, if it already exists the following will be used `image1.jpg`, if that exists, it will be `image2.jpg`, and so on 👍 

Thanks for the PR.

PR #1468, opened 2017-07-15
Fix #906

### LINKED ISSUE #906: [Feature Request] Menu Builder Route Prefixes
After playing around with everything since the last time I took a look, I wanted to check and see if this would be something that people might want.

As it stands, installing Voyager will drop a small bit of code into the Routes file with a nice Prefix that a developer can go in and change at their leisure. The problem is that after installation if somebody wants to change this prefix, all seeded admin menu items won't reflect the new change:

1. Install Voyager
2. Change the route prefix in routes/web.php to something else
3. Log into Voyager
4. Click a Menu Item

The only way a developer can change the route prefix is by first changing each menu item route by hand (*before* changing the prefix), and then adjusting the prefix in the routes file.

For new users, this approach may not seem apparent and could cause confusion when changing the prefix first and being presented with a NotFoundHttpException after clicking any of the admin menu items.

This can be remedied fairly easily by a few approaches:

1. Add an additional prompt during the installation process to allow the user to install Voyager on a nonstandard route with `admin` being the default
2. Add an optional `prefix` input under the Menu Editor that will prefix all **(?)** menu items with the designated menu prefix (should one exist)
3. Add an additional artisan command that will change the `routes/web.php` file as well as the seeded menu items to reflect the new change (ie: `voyager:routes:prefix`)

I see Item 1 being a good start, and perhaps the easiest in terms of getting Voyager set up with the desired routes.

I see Item 3 being more complicated to implement due to rewriting the `routes/web.php` reliably, but would ultimately streamline the prefixing process nicely

The only one I'm unsure about is Item 2, as this may have unintended consequences for users who want to include prefixed routes with external URLs in their menus.

Thoughts?

fletch3555:
Changes MenuItems Seeder to use routes instead of URLs.

Fixes the remaining outstanding issue in #906

### DIFF
--- a/publishable/database/seeds/MenuItemsTableSeeder.php
+++ b/publishable/database/seeds/MenuItemsTableSeeder.php
@@ -21,6 +21,7 @@ class MenuItemsTableSeeder extends Seeder
             $menuItem = MenuItem::firstOrNew([
                 'menu_id'    => $menu->id,
                 'title'      => 'Dashboard',
+                'url'        => '',
                 'route'      => 'voyager.dashboard',
             ]);
             if (!$menuItem->exists) {
@@ -36,6 +37,7 @@ class MenuItemsTableSeeder extends Seeder
             $menuItem = MenuItem::firstOrNew([
                 'menu_id'    => $menu->id,
                 'title'      => 'Media',
+                'url'        => '',
                 'route'      => 'voyager.media.index',
             ]);
             if (!$menuItem->exists) {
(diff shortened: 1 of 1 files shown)


akazorg:
Hello @fletch3555, I was looking to the code, I guess that is missing?
https://github.com/the-control-group/voyager/blob/master/src/Models/MenuItem.php#L84

akazorg:
Merged. TY!!

PR #1472, opened 2017-07-16
Fix message when deleting table

akazorg:
When deleting a table, the name was not displayed at modal message.


fletch3555:
Thanks!

PR #1473, opened 2017-07-16
Fix empty Modal when edditing BREAD

akazorg:
Error when using the BREAD for a table with a empty Modal.
@fletch3555 @marktopper is this supposed to be working?

Please don't merge this. I've found another case it's failing. (will update this message later)


akazorg:
Closing PR, as Voyager requires a Model for each BREAD item.

PR #1488, opened 2017-07-21
Changed blade files to use `getKey()` instead of assuming `id` is the primary key

edmandiesamonte:
Changed blade files to use `getKey()` instead of assuming `id` is the primary key


marktopper:
Thanks 👍 
Will just see if I can fix conflicts.

resources/views/bread/edit-add.blade.php
```diff
@@ -47,7 +47,7 @@ class="form-edit-add"
 
                             <!-- Adding / Editing -->
                             @php
-                                $dataTypeRows = $dataType->{(isset($dataTypeContent->id) ? 'editRows' : 'addRows' )};
+                                $dataTypeRows = $dataType->{($dataTypeContent->getKey() ? 'editRows' : 'addRows' )};
```
fletch3555, line 50: The logic for this ternary was changed.  You should keep the `isset()`

fletch3555: This applies to a number of changes here.  `->getKey()` returns the key value, not a boolean value

marktopper: Well, it could be that the primary key is not `id`.
So I think we should check for `$dataType->exists` (boolean) instead.

marktopper:
A fix for this have already been done at #1549

PR #1491, opened 2017-07-21
Bug fix for incorrect rows being used

edmandiesamonte:
Inside the `create()` method of the `VoyagerBreadController`, `editRows` was being used inside the forloop for `addRows` instead of the `addRows`itself.


akazorg:
Thanks @edmandiesamonte. Merging now.

PR #1499, opened 2017-07-24
Refactoring Authorization

fletch3555:
This is up for discussion and review.  Have at it guys!


src/VoyagerServiceProvider.php
```diff
@@ -237,6 +251,24 @@ public function registerConfigs()
         );
     }
 
+    public function registerGates()
+    {
+        $dataTypes = VoyagerFacade::model('DataType')::get();
+
+        foreach ($dataTypes as $dataType) {
+            $policyClass = isset($this->policies[$dataType->model_name]) ? $this->policies[$dataType->model_name] : BasePolicy::class;
```
fletch3555: @cotrudel, I'd leave this as BasePolicy all the time, and set the "PolicyName" field in the BREAD for those two through the seeder.

tnylea:
Hey @fletch3555 and @cotrudel 

This looks awesome!
Can you make a change to: `src/VoyagerServiceProvider.php` on line 258 and change:

```
$dataTypes = VoyagerFacade::model('DataType')::get();
```

to:

```
$dataTypes = VoyagerFacade::model('DataType')->get();
```

I believe this is making the Travis-CI tests fail. If we can get those tests to pass I would love to merge this in 👍 

Thanks!

cotrudel:
Just a note: this is also missing a database migration necessary for the custom policy to be entered. It is not stable enough to merge yet. I'll try to take a look tonight.

akazorg:
Congrats, nice job. I like it.
I haven't tested yet, and have one question:
We will need to create a policy file for each model we deal with?

cotrudel:
@akazorg There is a BasePolicy which the system defaults to. It makes use of the database permissions. If someone wants to specifically customize on top of those database permissions, they extend the BasePolicy (as per the example PostPolicy or UserPolicy) and enter their new policy location under policy_name in the database.

akazorg:
That's great @cotrudel, I had that idea but wasn't totally sure.

marktopper (approved):

akazorg (approved):

fletch3555:
I'm okay with it being a breaking change since it's only going into v1.0,
which is a new major release

handiwijoyo:
@cotrudel @fletch3555 I've create a PR here https://github.com/cotrudel/voyager/pull/1 and the tests should pass now

marktopper:
The tests are now working locally for me, however I cannot get the pull request to trigger a test at TravisCI.

fletch3555:
I also can't figure out how to trigger Travis to retest.  I'll fix the merge conflicts today and that should do it.

fletch3555:
@marktopper / @tnylea / @akazorg / @handiwijoyo / @cotrudel, please give this a once-over.  This should be good to go.

I even added in the interface for the User model, so we'll need to add that to the documentation.  We no longer need to REQUIRE use of Laravel's built-in Auth scaffolding.  We can support any user model that extends TCG\Voyager\Contracts\User, though this would require a bunch of it's own testing and a full smoke test to verify that it's properly used everywhere.  I only made it work for the Policies.

PR #1514, opened 2017-07-29
Fixing issue where the previous image field was being set to null if …

tnylea:
…no new image was selected


victorlap:
Is there a way to remove an image right now?

PR #1521, opened 2017-08-01
Crop images with preview 

alexeykostenko:
Hi! If you need to crop an image with a preview by coordinates, then paste Optional Details in BREAD for Image, for example:
```js
{
    "crop": [
        {
            "name": "avatar",
            "size": {
                "name": "max",
                "width": "800",
                "height": "400"
            },
            "resize": [
                {
                    "name": "norm",
                    "width": "600",
                    "height": "null"
                },
                {
                    "name": "min",
                    "width": "400",
                    "height": "200"
                }
            ]
        },
        {
            "name": "catalog",
            "size": {
                "name": "max",
                "width": "300",
                "height": "200"
            },
            "resize": [
                {
                    "name": "norm",
                    "width": "200",
                    "height": "null"
                },
                {
                    "name": "min",
                    "width": "100",
                    "height": "null"
                }
            ]
        }
    ]
}
`

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -105,3 +105,5 @@ php artisan voyager:admin your@email.com --create
 ```
 
 And you will be prompted for the users name and password.
+
+<!-- yakov -1 -->
```
fletch3555, line 109: Please remove this.  It's an unnecessary change.

src/Traits/CroppingImages.php
```diff
@@ -0,0 +1,104 @@
...
+
+namespace TCG\Voyager\Traits;
+
+use Illuminate\Http\Request;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\Model;
+use Intervention\Image\Facades\Image;
+use File;
+
+trait CroppingImages
```
fletch3555, line 11: I'm not sure I like the naming.  Most Laravel traits use the `-able` suffix (i.e. `Authenticatable`).  I'm open to suggestions, but `CroppingImages` seems weird.

src/VoyagerServiceProvider.php
```diff
@@ -88,6 +91,11 @@ public function boot(Router $router, Dispatcher $event)
             $this->addStorageSymlinkAlert();
         });
 
+        //Added a listener on event after update data
+        $event->listen('voyager.insert_update_data', function ($request, $slug, $rows, $data) {
```
fletch3555, line 95: So... we're calling `cropImages()` for ALL insert/update calls, regardless of whether images are used at all? I'm not sure I like that.  Was this tested in those cases?

fletch3555: I'm also not sure I understand why you're using a trait on the ServiceProvider for this.  Is this a trait you're expecting to use elsewhere as well?

PR #1531, opened 2017-08-02
crop images with preview

alexeykostenko:
I made changes to the code. I'm sorry but I could not think of a name for a trait with a -able suffix. I am also open to suggestions

If you need to crop an image with a preview by coordinates, then paste Optional Details in BREAD for  Image

*For example:*

```javascript
{
    "crop": [
        {
            "name": "article_main",
            "size": {
                "name": "max",
                "width": "800",
                "height": "400"
            },
            "resize": [
                {
                    "name": "norm",
                    "width": "600",
                    "height": "null"
                },
                {
                    "name": "min",
                    "width": "400",
                    "height": "200"
                }
            ]
        },
        {
            "name": "article_for_list",
            "size": {
                "name": "max",
                "width": "300",
                "height": "200"
            },
            "resize": [
                {
                    "name": "norm",
                    "width": "200",
                    "height": "null"
                }
            ]
        }
    ]
}
```


tnylea:
@alexeykostenko bad ass 👍  Testing this out now and I'll merge it in shortly :)

tnylea:
@alexeykostenko This is awesome, thanks for submitting it. Is there a way we can make the default location the storage folder and then crop folder. Additionally, I see that you're able to crop the image and it gets saved the first time, but after that it doesn't seem to let you edit the cropped image, would that be possible?

Let me know :) I would love to have this in the latest release.

Appreciate it 👍

PR #1535, opened 2017-08-03
Form builder

alexeykostenko:
If you need to create a custom form, then use form builder for BREAD

*For example:* 

```javascript
[
    {
        "class": "col-md-8",
        "panels": [
            {
                "class": "panel",
                "title": "translation.text_information",
                "fields": [
                    "title",
                    "excerpt",
                    "body",
                    "slug"
                ]
            },
            {
                "class": "panel panel-bordered panel-info",
                "title": "voyager.post.seo_content",
                "fields": [
                    "meta_keywords",
                    "meta_description"
                ]
            }
        ]
    },
    {
        "class": "col-md-4",
        "panels": [
            {
                "class": "panel panel-bordered panel-primary",
                "title": "translation.image",
                "fields": [
                    "image",
                    "status"
                ]
            }
        ]
    }
]
```

### DIFF
--- a/resources/views/bread/partials/form-builder.blade.php
+++ b/resources/views/bread/partials/form-builder.blade.php
@@ -1,0 +1,116 @@
+@php
+    $dataTypeRows = $dataType->{(isset($dataTypeContent->id) ? 'editRows' : 'addRows' )};
+    $formBuilderFields = collect([]);;
+@endphp
+<!-- form start -->
+<form role="form"
+        class="form-edit-add"
+        action="@if(isset($dataTypeContent->id)){{ route('voyager.'.$dataType->slug.'.update', $dataTypeContent->id) }}@else{{ route('voyager.'.$dataType->slug.'.store') }}@endif"
+        method="POST" enctype="multipart/form-data">
+    <!-- PUT Method if we are editing -->
+    @if(isset($dataTypeContent->id))
+        {{ method_field("PUT") }}
+    @endif
+
+    <!-- CSRF TOKEN -->
+    {{ csrf_field() }}
+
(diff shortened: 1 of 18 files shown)


resources/views/bread/partials/form-builder.blade.php
```diff
@@ -0,0 +1,116 @@
...
+        {{ method_field("PUT") }}
+    @endif
+
+    <!-- CSRF TOKEN -->
+    {{ csrf_field() }}
+
+    @foreach($formBuilder as $key => $htmlRow)
+        <div class="{{ $htmlRow->class }}">
+            @foreach($htmlRow->panels as $key_panel => $panel)
+                <div class="{{ $panel->class ?? '' }}">
```
tnylea: Great job :) 

I think this is only available in PHP 7 and up, If you change it to use the old ternary operator I'm sure all the tests will pass and we can merge this in. https://lornajane.net/posts/2015/new-in-php-7-null-coalesce-operator

Thanks.

resources/views/bread/partials/form-builder.blade.php
```diff
@@ -0,0 +1,116 @@
...
+                    </div>
+                    @endif
+                </div>
+            @endforeach
+        </div>
+    @endforeach
+
+    @php
+        $formBuilderFields = $formBuilderFields->collapse()->toArray();
+        $notFormBuilderFields = $dataTypeRows->whereNotIn('field', $formBuilderFields)
```
fletch3555: This line is breaking the TravisCI tests.  `whereNotIn()` is not valid in Laravel 5.3, which we still support.

PR #1537, opened 2017-08-04
Added auto filled validation and auto form type

zaimlvmry:
Auto form type and validation rule when creating bread

For example:
**varchar** will have text auto selected
**integer** will have number auto selected with numeric validation rule
**required** field will have auto filled required rule

### DIFF
--- a/resources/views/tools/database/edit-add-bread.blade.php
+++ b/resources/views/tools/database/edit-add-bread.blade.php
@@ -254,26 +254,28 @@
                                             {{ __('voyager.json.invalid') }}
                                         </div>
                                         <?php
-                                            $detail = !empty($dataRow->details) ? json_decode($dataRow->details) : new \stdClass;
-                                            if (!isset($detail->validation)) {
+                                        $detail = !empty($dataRow->details) ? json_decode($dataRow->details) : new \stdClass;
+                                        if (!isset($detail->validation)) {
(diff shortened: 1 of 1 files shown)


resources/views/tools/database/edit-add-bread.blade.php
```diff
@@ -1,3 +1,12 @@
+<?php
+    $types = [
+        'number' => ['smallint', 'mediumint', 'integer', 'float', 'double', 'decimal'],
+        'date' => ['datetime', 'date'],
+        'textarea' => ['bigtext', 'text', 'mediumtext', 'longtext'],
```
marktopper, line 5: This would be better stored in the form field object.
Example for the `textarea` it could be stored here: https://github.com/the-control-group/voyager/blob/master/src/FormFields/TextAreaHandler.php

marktopper (changes requested):

resources/views/tools/database/edit-add-bread.blade.php
fletch3555: Agreed.  Also, the field types listed are DBMS-specific.  Mysql and Postgres for example have different field names.  If we later choose to implement Mongo, MS SQLServer, etc, those will be different as well.

tnylea:
Closing due to inactivity, please open if you want to bring this back up 👍 
Thanks.

PR #1539, opened 2017-08-04
Fixing file upload is null

secondnetwork:
Fixing issue where the previous file field was being set to null if no new file was selected


tnylea:
@secondnetwork Thanks man. I just fixed styleCI issues. When the test finishes I'll merge this in 👍

PR #1544, opened 2017-08-08
Settings grouping

emptynick:
Lets you group Settings with a custom text/name.

groups

### DIFF
--- a/publishable/database/migrations/2017_08_05_000000_add_group_to_settings_table.php
+++ b/publishable/database/migrations/2017_08_05_000000_add_group_to_settings_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddGroupToSettingsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->string('group')->nullable()->after('order');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->dropColumn('group');
+        });
+    }
+}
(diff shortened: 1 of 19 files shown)


tnylea:
@emptynick dude! This kicks ass 👍 

I've actually wanted this for quite awhile. I'm testing it out right now. I'm wondering why Travis CI is failing. Looking into it further.

tnylea:
@emptynick, I love this addition. I just pushed a change so that way the Travis CI tests will pass.

I also added 4 different permissions, which are

 - read_settings
 - edit_settings
 - add_settings
 - delete_settings

We already had the `browse_settings` so I made sure to add the other ones 👍

tnylea:
Also, I would like to add something for version 1.0,

I would like for each setting key to be saved with a prefix of each group. For instance, let's say that we had a group of `admin` and we then created a new setting called `title`, for this setting to be referenced it would be nice if the user can use the following code:

```php
echo setting('admin.title');
```

I think that would be easier to read and understand, let me know your thoughts :)

Thanks.

emptynick:
I thought about this also, but accessible as an array, so you could do something like this:
`foreach (setting('admin') as $value)`
Could be really useful if you have depending settings (e.g. slides for a slideshow)

The only thing we have to make sure is that the setting-key is unique in the group, not in the whole settings.

tnylea:
@emptynick, yeah totally. That would be amazing. I was thinking that as we save the setting we can prefix it with the group, so it will automatically lowercase the group name (ex 'Admin' => 'admin') and then prefix it with the key, so if you add the key 'title', it will save the key as 'admin.title', I'll be working on this a bit today 👍

emptynick:
No worries, already done it.
Im just trying to figure out how to stay backward-compatible

tnylea:
Ok, cool. Good Idea. We can always add this to the Upgrade Guide. It will be something that users need to update in order to upgrade to the latest version.

fletch3555:
👍

emptynick:
I'm still waiting for your suggestions if the user should be able to change the keys, or if its an permission/config option.

Please let me know, I have 3 versions ready to pull!

marktopper:
Well, if it is going to be possible to change the key, the user should be warned about it, since it can have consequences.
However I don't think it's a completly bad idea to let them change it, we just need to make sure they always get a warning.

marktopper:
Let's start with not having it an option for the user to change the key, we can always add this option later.

Merging in now since the test have been fixed.

PR #1546, opened 2017-08-10
Fixing the primary key issue that was introduced after PR 1452

tnylea:
In this PR https://github.com/the-control-group/voyager/pull/1452

A new feature was added that allowed users to specify their own `primaryKey`, this will cause it to break if there is not a `primaryKey` specified.

This PR will default to the `id` if primaryKey is not specified in the model.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -136,13 +136,14 @@
                                         </td>
                                     @endforeach
                                     <td class="no-sort no-click" id="bread-actions">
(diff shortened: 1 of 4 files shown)


victorlap:
Why not use the default laravel way of accessing the primarykey? Either `$this->getKey()` or `$this->getKeyName()`

fletch3555:
@tnylea, @victorlap is correct, this should be `->getPrimaryKey()`, which does that check for you since it's just an overridden property on the model.

Also, please don't merge your own PRs.

PR #1555, opened 2017-08-14
Fixing error of empty details of multiple_select field in bread

mo7amed-3bdalla7:
reslove error appear when field of a bread selected as select_dropdown without options in detail with throw exception
`First parameter must either be an object or the name of an existing class`


tnylea:
Thanks man! Looks good.

PR #1559, opened 2017-08-14
fixed plugin datetimepicker

alexeykostenko:


fletch3555:
So.... what is this fixing...?

fletch3555:
dupe of #1562 and #1563?

alexeykostenko:
At the moment, the plugin datetimepicker does not work in the master branch, I fixed it

fletch3555:
It looks like the "fix" isn't a fix, it's a replacement to use a different plugin.  Was there an issue that this is fixing? I have no context here.  What was the original issue?

PR #1562, opened 2017-08-15
fix format datetimepicker

alexeykostenko:


fletch3555:
dupe of #1559 and #1563?

PR #1564, opened 2017-08-16
Allow raw select for relationships labels

DODMax:
I often find myself wishing I could customize the select statement of a relationship.

**Problem:** displaying member.firstname and member.lastname on a Voyager relationship

**Proposed solution:**
Allow user to provide their own select statement on the relationship details.
```
   "relationship": {
        "method": "member",
        "key": "id",
        "label": "name",
        "selectRaw": "id, CONCAT(firstname, ' ', lastname) AS name"
    },
```

Although I guess this solution would allow users with Database Tool access to perform **SQL injections**.. no?

Happy to hear your thoughts!

### DIFF
--- a/src/Http/Controllers/Traits/BreadRelationshipParser.php
+++ b/src/Http/Controllers/Traits/BreadRelationshipParser.php
@@ -39,7 +39,11 @@ trait BreadRelationshipParser
 
                 $relationships[$method] = function ($query) use ($relation) {
                     // select only what we need
-                    $query->select($relation->key, $relation->label);
+                    if (isset($relation->selectRaw)) {
+                        $query->selectRaw($relation->selectRaw);
+                    } else {
+                        $query->select($relation->key, $relation->label);
+                    }
                 };
             }
         });
--- a/src/Http/Controllers/Traits/BreadRelationshipParser.php
+++ b/src/Http/Controllers/Traits/BreadRelationshipParser.php
@@ -8,7 +8,7 @@ use TCG\Voyager\Models\DataType;
 
 trait BreadRelationshipParser
 {
-    protected $patchId;
+    protected $relation_field = [];
 
     /**
      * Build the relationships array for the model's eager load.
@@ -27,20 +27,15 @@ trait BreadRelationshipParser
                 $relation = $details->relationship;
                 if (isset($relation->method)) {
                     $method = $relation->method;
-                    $this->patchId[$method] = true;
+                    $this->relation_field[$method] = $item->field;
                 } else {
                     $method = camel_case($item->field);
-                    $this->patchId[$method] = false;
-                }
-
-                if (strpos($relation->key, '.') > 0) {
-                    $this->patchId[$method] = false;
                 }
 
                 $relationships[$method] = function ($query) use ($relation) {
                     // select only what we need
-                    if (isset($relation->selectRaw)) {
-                        $query->selectRaw($relation->selectRaw);
+                    if (isset($relation->method)) {
+                        return $query;
                     } else {
                         $query->select($relation->key, $relation->label);
                     }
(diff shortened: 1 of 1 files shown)


fletch3555:
Yes this certainly would allow for SQL injections (though if you have database access already, I'm not certain that's not already possible haha).  I'm not sure I like this particular implementation, but I can see the desire for it.  Seems like it'd be better in code as an accessor, but you'd still have to have a field to display it.  I think someone added the ability to add arbitrary (non-database) fields to the BREAD views as part of v1.0.  That, along with an accessor, would be a better solution I think

DODMax:
@fletch3555 Yes I agree this implementation is kind of rough :-)
I like the arbitrary field approach but we are still missing a way to add the database fields to the select query because as far as I can see the `getRelationships()` method in the master branch is still only selecting `label` and `key`:
```
$relationships[$method] = function ($query) use ($relation) {
    // select only what we need
    $query->select($relation->key, $relation->label);
};
```

marktopper:
Can't this already be done like this:
```
   "relationship": {
        "method": "member",
        "key": "id",
        "label": "name"
    },
```

```
public function member()
{
    return $this->belongsToMany(Member::class)->select(['id'], \DB::raw('CONCAT(firstname, \' \', lastname) AS name']);
}
```

Or however it would be done in eloquent?

marktopper:
Maybe instead of a JSON setting for it, it could become a model function which can handle any selection at all?

DODMax:
So what about we use this existing `"method": "member"` attribute on the relationship?

**The idea:** if `method:` is defined then use it as the relationship (without adding any `select` to the query), otherwise proceed as usual.

**Currently:**
If `method:` is not defined the method is assumed to be `camel_case($field)`, here  `memberId()`.
If `method:` is defined it is used as the relationship, however there is a limitation: later in  `relationToLink()` it is assumed that the field name is `snake_case($method).'_id'`

This means that the only acceptable value for `method:` is `member`, so I'm not sure what it was designed for other than allowing to skip defining `memberId()` on my model.

**Solution:**
Refactor BreadRelationshipParser to allow any custom method name.
This way in the json:
```
 "relationship": {
        "method": "memberCustomSelect",
        "key": "id",
        "label": "name",
    },
```
In the model:
```
public function memberCustomSelect(){
        return $this->member()->select(['id', \DB::raw('CONCAT(firstname, \' \', lastname) AS name')]);
}
```

DODMax:
Here is a commit illustrating those changes.
Note: I haven't tested that extensively.

Also, really not sure about this chunk of code I have removed on line 36:
```
if (strpos($relation->key, '.') > 0) {
    $this->patchId[$method] = false;
}
```
It doesn't seem like `key` would support nested relationships. Basically if key had a `.` then the field would be assumed to be `snake_case($method)` instead of `snake_case($method).'_id'` later in `relationToLink()` :confused:

tnylea:
I can merge this in, but the way that relationships are going to work in 1.0 are different then the previous way so this change may not even be relevent, but I'll merge in in nonetheless 👍 

Thanks.

bimo13:
Hi everyone, hi @DODMax and @tnylea ,
I wonder, is there any documentation for this feature ? I am currently in need to display custom-query on a browse page. What I need to do is to display a COUNT() query on the parent table from related table rows.

DODMax:
Hi @bimo13,

So in theory if this still works you would need to manually edit the details field of relationship in the data_rows table to add `method:myCustomDetailsQuery` such as it looks like:
```
 "relationship": {
        "method": "parentDetailsQuery",
        "key": "id",
        "label": "parent_count",
    },
```

Then in the model you are displaying the browse page for:
```
public function parentDetailsQuery(){
   return $this->parentRelation()->selectRaw('id, COUNT(field) AS parent_count');
}
```

Theorically...

bimo13:
Hi @DODMax ,
So, currently on my `details` field in the `data_rows` table, I have:
`{
    "model":"App\\ParentModel",
    "table":"parent_table",
    "type":"hasMany",
    "column":"id",
    "key":"id",
    "label":"id",
    "pivot_table":"active_guests",
    "pivot":"0"
}`
for the `parent_hasmany_child_relationship` row.

Do I need to append the current json, or just replace it with the method you show me ? Thank you.

Vulgar:
I see that this was implemented, is it still active in the current version?
If I have options that belong to an attribute, it doesn't really make sense to duplicate the attribute() function as attributeId() if laravel syntax already assumes an "_id" suffix.

PR #1565, opened 2017-08-16
fix rich_text_box

alexeykostenko:
if column not null then textarea required but hint is hidden


fletch3555:
I need you to add more context when you open PRs.  Referencing an existing issue would suffice as well.  Your comment doesn't seem to make sense with the change that was made.  Please describe the issue and how your fixing it.

marktopper:
I am unsure why removing the `required` part when it in fact is required?

PR #1568, opened 2017-08-17
Do not store default avatar in database

marktopper:


fletch3555:
Awesome, thanks @marktopper!

PR #1583, opened 2017-08-22
(WIP) Backward support

marktopper:
Re-adding the support for Laravel 5.4 ~~and 5.3~~.


fletch3555:
Can we get rid of the `phpunit/phpcov: dev-master` and replace it with a released version?  It should fix the PHP 5.6 errors in Travis.  All later versions seem to be failing due to an implicitly abstract class (TestCase), so it just needs to define the missing method.

PR #1587, opened 2017-08-22
Update Controller.php

ya-tmch:
Update **getContentBasedOnType** method.


fletch3555:
I can see what change you're making, and I'm okay with it, but in the future, we require all PRs include a complete description of the issue and what you're doing to fix it.  If the PR is to fix an already-reported issue, then linking to that issue would meet this requirement.

marktopper:
What @fletch3555 said still applies.
However I am merging this in now, great job.

PR #1591, opened 2017-08-23
Fixed translation queried twice

DODMax:
Fixed translate() method loading translations twice


fletch3555:
Good catch! Inverted logic issues can be hard to find.  Thanks!

PR #1596, opened 2017-08-24
Fixed incorrect field name

ya-tmch:
There was an error in the field name. Instead of "field" you need to use "type". Corrected on three lines.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -64,16 +64,16 @@ abstract class Controller extends BaseController
             if (is_null($content)) {
 
                 // If the image upload is null and it has a current image keep the current image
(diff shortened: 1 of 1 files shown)


fletch3555:
Huh..... This is the first I've heard of this being an issue.  I imagine if it was a problem, it would've been reported prior to this.  @marktopper, any thoughts?

marktopper:
I haven't heard anyone complaining about it, but when looking at this code it all makes sense using `type` instead of `field`.

I will do some proper testing of this locally.

ya-tmch:
Apparently, the name of the field usually coincides with its type. I made a field like "file" with the name "video", and noticed that after the update the content disappears.

marktopper:
Makes sense @ya-tmch.

Just tested, works perfectly.
Thanks, merging in.

PR #1631, opened 2017-08-31
fix: BREAD fails for custom created_at field.

greut:
My models inherit from an old Symfony application, hence all the names are f-up.


tnylea:
Thanks for this PR. Merging it in now 👍

PR #1640, opened 2017-09-01
Problem with timestamp in getContentBasedOnType

saibotd:
Content is always null if type is timestamp and method != PUT


tnylea:
Thanks man!

PR #1642, opened 2017-09-01
change fire to handle

tbence94:
Since Laravel 5.1 `fire` method is replaced by `handle`. This fixes the commands in the newer versions.

If you want to be backwards compatible, we could create a `handle` method which calls `fire` or the other way around to make sure every laravel version finds the command logic. But i think this is better.


fletch3555:
This has already been resolved in the `backwards-compatible` branch. It just hasn't been merged into master yet

PR #1648, opened 2017-09-02
remove unnecessary code in template file

xcaptain:
`//{camel_case    ($row->field)}()->pluck($options->relationship->key)->all() : array(); ?>`
would result an early closed tag for php, makes `$relationshipOptions = app($options->model)->all();` not evaluated, so remove all such commented code in this file.


tnylea:
Thanks man.

PR #1654, opened 2017-09-03
[Update] Russian translation

Kleindberg:
Add missing strings from English translation and correct some strings.

P.S. Why translation on site not update after voyager.php editing? Showing old texts.
P.P.S. Too many strings still untranslatable (not represent in voyager.php file).


publishable/lang/ru/voyager.php
```diff
@@ -17,37 +17,39 @@
         'all_done'               => 'Готово',
         'are_you_sure'           => 'Вы уверены',
         'are_you_sure_delete'    => 'Вы точно хотите удалить',
-        'auto_increment'         => 'Auto Increment',
+        'auto_increment'         => 'Счётчик',
```
dmekhov: Don't translate this field: 'auto_increment' => 'Счётчик', 
Auto Increment is an established term

publishable/lang/ru/voyager.php
```diff
@@ -74,16 +76,15 @@
...
+        'showing_entries'        => 'Показана от :from до :to из :all запись|Показано от :from до :to из :all записей',
         'submit'                 => 'Отправить',
         'successfully_added_new' => 'Успешное добавление',
         'successfully_deleted'   => 'Успешное удаление',
         'successfully_updated'   => 'Успешное обновление',
         'timestamp'              => 'Временная метка',
         'title'                  => 'Название',
         'type'                   => 'Тип',
-        'unsigned'               => 'Unsigned',
+        'unsigned'               => 'Не задано',
```
dmekhov: This is wrong translate: 'unsigned' => 'Не задано',
we can use 'unsigned' as established term or 'Беззнаковое'

publishable/lang/ru/voyager.php
```diff
@@ -97,31 +98,27 @@
...
+        'usage_help'           => 'Чтобы получить значения параметров, используйте в шаблоне код ',
         'save'                 => 'Сохранить настройки',
-        'new'                  => 'Новые настройки',
-        'help_name'            => 'Название параметра, например, Admin Title',
-        'help_key'             => 'Ключ параметра, например, admin_title',
-        'help_option'          => '(необязательно, применяется только к определенным типам,'.
-                                  ' таким как dropdown box или radio button)',
+        'new'                  => 'Создать настройку',
+        'help_name'            => 'Название параметра (например, Мой параметр)',
+        'help_key'             => 'Ключ параметра (например, admin_title)',
```
dmekhov: If you use "Мой параметр" as example of `help_name` - you must use "my_param" instead "admin_title" in `help_option` translation

dmekhov (changes requested):

publishable/lang/ru/voyager.php
Kleindberg: If this sring is not translatable, it would be better to remove it from translations.

publishable/lang/ru/voyager.php
Kleindberg: I don't know, that 'unsigned' in this translation is a modificator's name. It would be better to remove it from all translations.

marktopper:
To see the translations edit the configuration `locale` in `config/app.php`.

Kleindberg:
I solve a problem with outdated translations: just copied it from vendor folder to the public. I can't understand why we need a translations table in database if we use the files?

marktopper:
The translations table is for translating BREAD. Since BREAD is dynamicly inserted, the translations for that needs to be dynamic as well.

PR #1656, opened 2017-09-05
BREAD Index: eager loading of translations

DODMax:
**Currently**
The translation table is queried for each line on BREAD index (for translatable models).

**Solution**
Eager load translations in VoyagerBreadControlle index() method.


fletch3555:
Nice job, thanks!

PR #1660, opened 2017-09-05
Fix wrong compare type of null value. string instead null given

dmekhov:
If we dont have `$options->resize->height` param on
https://github.com/the-control-group/voyager/compare/master...dmekhov:master#diff-bf550c9ef0c15a627ba85f6bca552e12L195
`$resize_height` is set to  `null` `$resize_height = null;` and then on https://github.com/the-control-group/voyager/compare/master...dmekhov:master#diff-bf550c9ef0c15a627ba85f6bca552e12L222  `$thumb_resize_height = $resize_height;`

But on https://github.com/the-control-group/voyager/compare/master...dmekhov:master#diff-bf550c9ef0c15a627ba85f6bca552e12L339 we are compare `$thumb_resize_height` with string and we get 0 in `$thumb_resize_height` and than we get `division by 0` error from resize method of Intervention\Image

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -221,11 +221,11 @@ abstract class Controller extends BaseController
                                     $thumb_resize_width = $resize_width;
                                     $thumb_resize_height = $resize_height;
 
-                                    if ($thumb_resize_width != 'null') {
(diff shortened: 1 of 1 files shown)


marktopper:
Thanks

PR #1662, opened 2017-09-05
Fixes for1.0

tnylea:
Fixing functionality where new Fields were not showing up in the bread builder after the model was created. Also adding fix for the images so that way if someone forgets to add the `APP_URL` in their .env file the images will still show correctly.

@fletch3555 When these tests pass can you merge this in for me 👍

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -66,7 +66,7 @@ return [
     */
 
     'storage' => [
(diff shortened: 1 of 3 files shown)


tnylea:
@brianchristopherpeck Can you merge this in for me.

brianchristopherpeck (commented):
Code looks good

PR #1664, opened 2017-09-05
Fixes for1.0

tnylea:
Adding additional fields for posts Edit-Add template.
Adding style fixes for the add and bulk delete buttons.
Removing unnecessary posts/browse.blade.php, since we can now use the default browse view.


brianchristopherpeck (commented):
Code looks solid

PR #1671, opened 2017-09-06
Update SettingsTableSeeder.php

erayusta:


marktopper:
For this change to be useful, we should also include a setting for it here: https://github.com/the-control-group/voyager/blob/master/resources/views/master.blade.php#L12

marktopper (changes requested):
For this change to be useful, we should also include a setting for it here: https://github.com/the-control-group/voyager/blob/master/resources/views/master.blade.php#L12

marktopper:
Closed due to inactivity.

PR #1672, opened 2017-09-06
Fixing Auto-discovery database connection error

tnylea:
When running the `php artisan package:discover` Voyager will throw an error because it tries to make a connection to the database, but there is no database connection yet.

Adding this try catch will catch the error when the user does `composer require tcg/voyager` because the `php artisan package:discover` will be able to run without error.

See screenshot of error:

screen shot 2017-09-06 at 11 22 33 am


marktopper:
Great job

tnylea:
Thanks man!

PR #1673, opened 2017-09-06
Add Some Feature

erayusta:
- Favicon settings
- Turkish Lang
- add "additional fields" translatable code


fletch3555:
This includes changes from #1671 as well as whatever you were hoping for this PR to be.  Please avoid opening PRs for multiple unrelated changes as it forces us to either accept all of none of them.

Please separate out each concern (Settings changes vs. translation changes) into their own distinct change set and open a PR for each.

PR #1674, opened 2017-09-07
Adding the icon fixes, login styles, and timestamp bug

tnylea:
Fixing the broken icons adding some more login styles and fixing the timestamp bug where it was not being Parsed correctly when being updated.


tnylea:
@fletch3555 @brianchristopherpeck If one of you guys could merge this in when the tests all pass that would be awesome.

Thanks 👍

marktopper:
Looks good! 👍

Issue #1669, opened 2017-09-06
Setting model can not update


JesonCkon:
- Laravel Version: 5.5
- Voyager Version: dev-master
- PHP Version: 7.0
- Database Driver & Version: mysql5.6

### Description:

`/vendor/tcg/voyager/src/Http/Controllers/VoyagerSettingsController.php`
when i call the update function laravel $request post content 'admin_title' but `getContentBasedOnType` get $request input by the settings->key 'admin.title' so always return null....
### Steps To Reproduce:
`
if($content === null){
     $content = $request->input(str_replace('.','_',$setting->key));
}`

erayusta:
Maybe this is a better solution

$content = $this->getContentBasedOnType($request, 'settings', (object) [
               'type'    => $setting->type, 
                **'field'   => str_replace('.','_',$setting->key),**
                 'details' => $setting->details,
                'group'   => $setting->group,
            ]);

erayusta:
This is working code VoyagerSettingsController.php  (starting from line 86)

$content = $this->getContentBasedOnType($request, **'public/settings'**, (object) [
                'type'    => $setting->type,
                **'field'   => str_replace('.','_',$setting->key),**
                'details' => $setting->details,
                'group'   => $setting->group,
            ]);

megaket4up:
Also sqlite, pgsql, mysql not work. I can not save anything.

php: 7.1
laravel/framework: 5.5.*
tcg/voyager: dev-master

vshapovalov:
Hi all, 
php replace 'dot' with underscore ('admin.title' -> 'admin_title') in form field names. 

simple replacing may be incorrert with 2 different keys, for example:
user added colors.main and colors_main as setting keys :], when settings will be saved, both of them will have one value.

lancepioch:
Labels: invalid

Close (PR merged + dev-master is not stable)

PR #1675, opened 2017-09-07
fix Update settings

megaket4up:
https://github.com/the-control-group/voyager/issues/1669


src/Http/Controllers/VoyagerSettingsController.php
```diff
@@ -82,9 +82,9 @@ public function update(Request $request)
         $settings = Voyager::model('Setting')->all();
 
         foreach ($settings as $setting) {
-            $content = $this->getContentBasedOnType($request, 'settings', (object) [
+            $content = $this->getContentBasedOnType($request, 'public/settings', (object) [
```
marktopper: Why the `public/` path if I may ask?

megaket4up: To upload into `storage/app/public/settings`.

marktopper: Thanks, will test this today 👍

megaket4up: Ok

fletch3555: I think we need a bit more cleanup if this is going to work this way.  That `public` path gets passed in as `$slug`, which is no longer accurate if we're passing a full path.  `$slug` is only used as a prefix to the storage path, so it should still work correctly.  I'd like to know why it broke, though... why this fix is needed in the first place.

Assuming this fix is actually necessary, i propose undoing this so it remains `$slug` (for future use), then adding the `public` prefix to the file and image path in `Controller::getContentBasedOnType()`

src/Http/Controllers/Controller.php
```diff
@@ -167,7 +167,7 @@ public function getContentBasedOnType(Request $request, $slug, $row)
                     $filesPath = [];
                     foreach ($files as $key => $file) {
                         $filename = Str::random(20);
-                        $path = $slug.'/'.date('F').date('Y').'/';
+                        $path = $slug.'/public/'.date('F').date('Y').'/';
```
fletch3555: Wouldn't you want `public/` before `$slug`, not after?

megaket4up: oOps. I'm sorry. Fixed

fletch3555: Thanks!

tnylea:
Just checked this branch and made sure settings are working. Thanks @megaket4up, I did remove the public reference, you will see it should be working now.

I've also fixed the dashboard analytics with this commit 👍 

I'll merge in tomorrow or anyone can merge in after tests have passed.

Thanks :)

PR #1683, opened 2017-09-08
Fix column for index route on database page

Iphoning:


tnylea:
Awesome. Thanks man!

src/Http/Controllers/VoyagerDatabaseController.php
```diff
@@ -24,12 +24,15 @@ public function index()
     {
         Voyager::canOrFail('browse_database');
 
-        $dataTypes = Voyager::model('DataType')->select('id', 'name')->get()->pluck('id', 'name')->toArray();
+        $select = Voyager::model('DataType')->select('id', 'name', 'slug')->get()->toArray();
```
fletch3555, line 27: Is `$select` used for anything other than the `collect()` call 2 lines down?  If not, you should be able to combine the 2 lines like below:

```
$dataTypes = Voyager::model('DataType')->select('id', 'name', 'slug')->get()->keyBy('name')->toArray();
```

@tnylea, you merged it.  Thoughts?

tnylea: Yeah, I think that would be great adding it to one line. You're right it would definitely make it cleaner 👍

dmytro-afonin: It is a metter of taste. I don't like to build "trains" with code. It is hard to debug in future.

fletch3555: Okay, committed the change.  Thanks guys!

fletch3555: I don't think this is really a "train".  The `->toArray()` and `collect()` were redundantly converting back and forth, and were therefore unnecessary

PR #1690, opened 2017-09-09
allow empty fields in voyager settings

mo7amed-3bdalla7:
sometimes we need to save setting field with an empty string or remove a setting value

### DIFF
--- a/src/Http/Controllers/VoyagerSettingsController.php
+++ b/src/Http/Controllers/VoyagerSettingsController.php
@@ -89,7 +89,10 @@ class VoyagerSettingsController extends Controller
                 'group'   => $setting->group,
             ]);
 
-            if ($content === null && isset($setting->value)) {
+            if ($content === null && $setting->type !== 'image') {
+                $content = '';
+            }
+            if ($content === null && $setting->type === 'image') {
                 $content = $setting->value;
             }
 


src/Http/Controllers/VoyagerSettingsController.php
```diff
@@ -89,7 +89,10 @@ public function update(Request $request)
                 'group'   => $setting->group,
             ]);
 
-            if ($content === null && isset($setting->value)) {
+            if ($content === null && $setting->type !== 'image') {
+                $content = '';
+            }
+            if ($content === null && $setting->type === 'image') {
```
fletch3555, line 95: I don't really like the code style here.  It's difficult to read.

But beyond that, you've changed the logic by removing `isset()`.  Also, you've seemingly added an arbitrary check for the "image" type, with no explanation why that particular type is excluded.

mo7amed-3bdalla7: thanks for your review
i excluded image because image in update action will return with a null input but image may have a value in database if i remove this check the value in database will be removed

fletch3555: That makes sense for images (and is what I expected your answer to be), but are you certain there are no other types that need to be excluded?

mo7amed-3bdalla7:
```
<select name="type" class="form-control" required="required">
                            <option value="">{{ __('voyager.generic.choose_type') }}</option>
                            <option value="text">{{ __('voyager.form.type_textbox') }}</option>
                            <option value="text_area">{{ __('voyager.form.type_textarea') }}</option>
                            <option value="rich_text_box">{{ __('voyager.form.type_richtextbox') }}</option>
                            <option value="code_editor">{{ __('voyager.form.type_codeeditor') }}</option>
                            <option value="checkbox">{{ __('voyager.form.type_checkbox') }}</option>
                            <option value="radio_btn">{{ __('voyager.form.type_radiobutton') }}</option>
                            <option value="select_dropdown">{{ __('voyager.form.type_selectdropdown') }}</option>
                            <option value="file">{{ __('voyager.form.type_file') }}</option>
                            <option value="image">{{ __('voyager.form.type_image') }}</option>
 </select>
```


from code there is two types file and image

PR #1694, opened 2017-09-10
Update multiple_images.blade.php

marcelosantos89:
In case of model casting check if its string or array.
This way makes it easier for people to get data from database

Ex: Model\Page.php

public function getImageAttribute($value)
{
    return json_decode($value);
}

### DIFF
--- a/resources/views/formfields/multiple_images.blade.php
+++ b/resources/views/formfields/multiple_images.blade.php
@@ -1,6 +1,12 @@
 <br>
 @if(isset($dataTypeContent->{$row->field}))
-    <?php $images = json_decode($dataTypeContent->{$row->field}); ?>
+    <?php
+      if(is_array($row->field)){
+        $images = $dataTypeContent->{$row->field};
+      } else {
+        $images = json_decode($dataTypeContent->{$row->field});
+      }
+    ?>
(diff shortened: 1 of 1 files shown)


marcelosantos89:
This screws with the browse.blade.php but i can't find a solution for it.

fletch3555:
This feels like the wrong approach, but I can't pinpoint why.  I also don't have a better suggestion at the moment.  Can you describe the problem it's trying to solve for me?

marcelosantos89:
@fletch3555 Lets say you have a table Works and in Works you have image field. In voyager you say that field is multiple images. So far so good. In Work Model using Eloquent i can parse the attributes with getImageAttribute and return json_decode($value) and doing that i don't get the json from the field image but an array with the images to use in Blade. Did you understand? I can provide my source code if you want.

For example to use this image field in blade i have to @foreach(json_decode($works->image) as $image) and i wanted to use @foreach($works->image as  $image) but it's the same. i didn't want to parse data in the view that was all. But it turned out to give me more problems than it helped

fletch3555:
This is old and the original repo seems to have been deleted.  We're going to be refactoring all of the image stuff anyway.  Closing

PR #1710, opened 2017-09-14
Use translation methods instead of direct property

marktopper:
Fix #1646

Issue #1720, opened 2017-09-15
Error in Voyager 1.0


vitormicillo:
- Laravel Version: 5.5.4
- Voyager Version: 1.0
- PHP Version: 7.0
- Database Driver & Version: mysql server 5.7

### Description:
Error with users when enabling multilanguage

### My changed model:
`
namespace App;

use Illuminate\Notifications\Notifiable;
use Illuminate\Foundation\Auth\User as Authenticatable;
use \TCG\Voyager\Traits\Translatable;

class User extends \TCG\Voyager\Models\User
{
    use Notifiable;
    use Translatable;

    protected $tables = ['translations'];

    protected $fillable = [
        'name', 'email', 'password',
    ];

     protected $hidden = [
        'password', 'remember_token',
    ];
}
`
![image](https://user-images.githubusercontent.com/1921806/30483545-8fd81212-9a27-11e7-9ebb-571bc228dd7e.png)

vaggelis2018:
Did u added more columns to user table or role table?

vitormicillo:
problem solved with this command
composer require tcg/voyager:dev-fix-1719
Thanks @marktopper

PR #1721, opened 2017-09-15
Check if method exists before running it on models

### LINKED ISSUE #1719: Error 403 on saving settings
- Laravel Version: 5.5.0
- Voyager Version: 1.0.0
- PHP Version: 7.1.8
- Database Driver & Version: Maria DB 10.1

### Description:
When i try to save a setting i get an error 403. Also all translatable models returns Call to undefined method TCG\Voyager\Models\User::translatable()
I think problem is relationships
### Steps To Reproduce:
Just migrated to new laravel and voyager versions and enabled translations.

marktopper:
Fix #1719 and #1720 
This happens after doing another fix #1710

### DIFF
--- a/src/Helpers/helperTranslations.php
+++ b/src/Helpers/helperTranslations.php
@@ -13,8 +13,9 @@ if (!function_exists('is_field_translatable')) {
             return;
         }
 
-        return $model->translatable() &&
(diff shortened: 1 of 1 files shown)


fletch3555:
👍

PR #1726, opened 2017-09-16
User self role control

berkayoztunc:
İ found a logic for fix this ,in this case ;

if user have browse_admin permision , after try to update profile he can and change his role by this path.


fletch3555:
This permission should be managed by the UserPolicy, not accessed directly.

Issue #1716, opened 2017-09-15
Question about Relation


davodavodavo3:
- Laravel Version: 5.4
- Voyager Version: 0.11

### Description:
I want to show referer relationship in users edit(add) page.
User can be affiliated by other user, and i need to show who affiliate him.
Now i create relation in dropdown select, but there is showing the name of current user like 

![screen](https://user-images.githubusercontent.com/10462563/30464591-08838b6c-99db-11e7-9aec-7a983634b411.jpg)

Self registered - need to show when the user registered himself
Admin - everything is fine here
User - exclude current user name from relation , need to show only other users 

My relation in user model
```
public function referredBy()
{
     return $this->hasOne($this, 'affiliate_id', 'referred_by');
}
```

Can anyone do this, can someone help?

davodavodavo3:
Auuuuuuu is there anyone can halp me?))))
@fletch3555 ?

lancepioch:
Labels: question, enhancement

marktopper:
Please share the BREAD settings for this field.

fletch3555:
Closing due to inactivity.  If this is still a problem, please answer @marktopper's question and we'll reopen it.

PR #1730, opened 2017-09-16
Limit access to the admin area by ip's

davodavodavo3:
You can enable blocker functionality in Voyager config file and block ip's which are not specified in list of ip's.
Blocked ip's can not enter the admin area.
I think will be useful to create maintenance mode switcher in Admin Settings area.
And with allowed ip's user can access site in maintenance mod.
What do you think?


akazorg:
Great job, I like this idea. Code looks good.
The multiple emojis with your own post - are priceless :D

davodavodavo3:
Thanks @akazorg )))))))

davodavodavo3:
@akazorg can you help me with one question?
#1716

fletch3555:
I'm not sure how I feel about this being in the core... Seems like a great Hook though.

In either case, the whitelist config should be an array of strings, not just a string.  You shouldn't be calling `explode(",")` on that config, it should be presented to you appropriately.

We'll discuss this and get back to you

davodavodavo3:
Good come back do not be shy)))))
Change it to array and call in_array, all fine.
This is for the security of the admin from spam and other similar and not similar attacks.
Online shop's, web sites with payments ,user wallets in databse, or privat pornographic web site created on voyager))).
This is little security candy.))))

fletch3555:
We've discussed this, and believe it would be best built as a hook instead. https://larapack.io/

Thank you!

davodavodavo3:
@fletch3555 how to crate hook?
I cant install any hooks from https://larapack.io/.
If i try it show me
```
Could not find package voyager-polls at any version for your minimum-stability (stable). Check the pac
  kage spelling or your minimum-stability
```

fletch3555:
@davodavodavo3, that's an unrelated issue that should be opened as it's own issue

davodavodavo3:
@fletch3555 there is an open issue in voyager-polls, but no one answered

PR #1736, opened 2017-09-17
Fix coordinates field

Nox04:
I don't know if this is the best way to do or maybe I was using  this field in the incorrect way, but I was getting an error because the coordinates formfield was trying to get coordinates from the database.

If there's a way to use correctly this formfield, please close my PR and tell me how to do it.

Thanks in advance.


fletch3555:
The form-fields are all intended to be used for display and input of database-driven fields.  Sounds like you may have been using it incorrectly.

Nox04:
Yeah I know that. I think I didn't explain myself good enough. (Sorry for my english).

What I was trying to say it's the coordinates field was trying to read the value from the database in the ADD view (I know add and edit share the view) but I was checking the others form-fields and every single one validate if there's some value in the database before trying to display it.

`@if(isset($dataTypeContent->{$row->field}))`

Coordinates form field is always requesting data from database. I got this error every time (in the add view).

deepinscreenshot_select-area_20170918222314

Anyway, I just added the `@if(isset($dataTypeContent->{$row->field}))` and it works like a charm. Maybe I was using it incorrectly but maybe this helps someone else.

PR #1738, opened 2017-09-17
Added missing language keys to spanish version

Nox04:
I just added some language keys to spanish translation.


marktopper:
Thanks

PR #1748, opened 2017-09-19
Change video demo to new video tutorial

handiwijoyo:


fletch3555:
An excellent suggestion, thanks!

PR #1752, opened 2017-09-19
Hide developers tips in settings

mo7amed-3bdalla7:
The developer tip beside every input in settings appear even if `show_dev_tips` in config file equal false


marktopper:
Thanks

mo7amed-3bdalla7:
Thanks for awesome package

PR #1757, opened 2017-09-19
Replaces real domain (admin.com, email.com) with example.com

shrink:
For documentation and dummy accounts a [reserved domain](https://www.iana.org/domains/reserved) should be used, e.g: `example.com`.

> As described in RFC 2606 and RFC 6761, a number of domains such as example.com and example.org are maintained for documentation purposes. These domains may be used as illustrative examples in documents without prior coordination with us. They are not available for registration or transfer.

The use of a real domain ([admin.com](http://admin.com), [email.com](http://email.com)) means that if a malicious actor had control over that domain they could take over accounts, and unintended emails will be sent to that mail server which will be considered spam, which may harm the sender's mail server reputation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ View the Voyager Cheat Sheet: https://voyager-cheatsheet.ulties.com/
 
 Laravel Admin & BREAD System (Browse, Read, Edit, Add, & Delete), made for Laravel 5.3.
 
-After creating your new Laravel application you can include the Voyager package with the following command: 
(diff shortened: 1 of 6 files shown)


fletch3555:
Excellent, thank you! I've been meaning to do this myself, so you just saved me 5 minutes

fletch3555:
Reverting this in #1760.  I forgot many of our videos mention that email directly.  This will be reconsidered next time we make new video documentation

Issue #1731, opened 2017-09-16
Fix Cancele Delete Settings Field


Akopean:
- Laravel Version: 5.4.*
- Voyager Version: 0.11.14
- PHP Version: 5.6.4
- Database Driver & Version:

### Description:
```
<form action="http://laravel/admin/settings/???" <- id not updating id="delete_form" method="POST">
                        <input type="hidden" name="_method" value="DELETE">
                        <input type="hidden" name="_token" value="Rj0CIfoJOVPHmiV9n8eZ5CUj3jiBf4tD8b7GBgkY">
                        <input type="submit" class="btn btn-danger pull-right delete-confirm" value="Yes, Delete This Setting">
                    </form>
```
### Steps To Reproduce:
1.Cancele Delete Setting Item
2.Select Next setting field and delete
3. Id  of the deleted field is not updated

davodavodavo3:
Agree there is issue

davodavodavo3:
Also when modal is open click on trash icon in modal header and booooom. There is another bug

lancepioch:
Labels: possible bug

PR #1764, opened 2017-09-20
Fix for deleting settings

davodavodavo3:
Fix deleting settings
Also fixed a bug if you click on trash icon in delete modal
issue #1731

### DIFF
--- a/resources/views/settings/index.blade.php
+++ b/resources/views/settings/index.blade.php
@@ -419,7 +419,7 @@
                     </h4>
                 </div>
                 <div class="modal-footer">
-                    <form action="{{ route('voyager.settings.delete', ['id' => '__id']) }}" id="delete_form" method="POST">
+                    <form action="{{ route('voyager.settings.delete', ['id' => '0']) }}" id="delete_form" method="POST">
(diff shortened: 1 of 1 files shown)


resources/views/settings/index.blade.php
```diff
@@ -444,11 +444,13 @@
...
 
-            $('.voyager-trash').click(function () {
+            $('.panel-actions .voyager-trash').click(function () {
                 var display = $(this).data('display-name') + '/' + $(this).data('display-key');
-
+				var form = $('#delete_form')[0];
+				
                 $('#delete_setting_title').text(display);
-                $('#delete_form')[0].action = $('#delete_form')[0].action.replace('__id', $(this).data('id'));
+                form.action = form.action.match(/\/[0-9]+$/) ? form.action.replace(/([0-9]+$)/, $(this).data('id')) : form.action + '/' + $(this).data('id');
```
fletch3555, line 452: This fix doesn't support non-numeric id values.  While we don't explicitly make use of them within Voyager, there's nothing saying we won't in the future

handiwijoyo:
@fletch3555 Do we still want to merge this or wait for update?

PR #1773, opened 2017-09-20
Support for POST method in content timestamp. Fixes #1706.

### LINKED ISSUE #1706: Timestamp not saving
- Laravel Version: #.#.#
- Voyager Version: 0.11.#
- PHP Version:
- Database Driver & Version:

### Description:
Hi!

I've integrated Fullcalendar(fullcalendar.io) with Laravel Voyager , but faced this problem.
To save date and time for calendar, I've chosen timestamp format and "timestamp" in bread . But when I create task, every first time timestamp doesn't save date and time. It saves null or current timestamp (If I choose CURRENT_TIMESTAMP in default value of row in database). When I edit, it saves new date and time without problem. So as I understand, the main problem with edit-add.blade.php or with Controller. 
In ./vendor/tcg/voyager/resources/views/tools/database/edit-add-bread.blade.php I've found some differences in adding/editing timestamp, but didn't find any problem with them. 
The only row which looks like causing problem in Controller

` case 'timestamp':
                if ($request->isMethod('PUT')) {
                    if (empty($request->input($row->field))) {
                     $content = null;
} else {
                        $content = gmdate('Y-m-d H:i:s', strtotime($request->input($row->field)));
                    }
                }
` 
I tried to comment 
`if (empty($request->input($row->field))) {
                     $content = null;
} else {`
or even change
`           $content = null;`
to 
`                        $content = gmdate('Y-m-d H:i:s', strtotime($request->input($row->field)));
`

but nothing changed... 

Can you please help me?

bencomeau:
Added support for `POST` method when getting the content type of `timestamp`. Prior to this pull request only `PUT` was supported.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -366,7 +366,7 @@ abstract class Controller extends BaseController
             /********** TIMESTAMP TYPE **********/
             case 'timestamp':
                 $content = $request->input($row->field);
-                if ($request->isMethod('PUT')) {
+                if (in_array($request->method(), ['PUT', 'POST'])) {
                     if (empty($request->input($row->field))) {
                         $content = null;
                     } else {


marktopper:
Thanks!

PR #1774, opened 2017-09-20
Rename 2017_11_04_000000_alter_post_nullable_fields_table.php to 2016…

mortenscheel:
…_11_04_000000_alter_post_nullable_fields_table.php

Timestamp was in the future.


fletch3555:
It should be April 11th, 2017, since that's when it was added.  Please change the date stamp to 2017_04_11 instead of 2016_11_04.

This will cause issues with those that have already run the migrations, so we have to be careful with these changes.

mortenscheel:
Good point. I've changed the filename.

Issue #1766, opened 2017-09-20
php artisan migrate:refresh returns error on fresh install


adddz:
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version:
- Database Driver & Version:

### Description:

After a fresh install of Laravel and Voyager, if I try to run `php artisan migrate:refresh` it returns error.

```
php artisan migrate:refresh --seed


  [Doctrine\DBAL\DBALException]
  Unknown database type enum requested, Doctrine\DBAL\Platforms\MySQL57Platform may not support it.
```

mortenscheel:
I ran into the same issue, but I managed to fix it by modifying a hack someone else made in [this comment](https://github.com/laravel/framework/issues/1186#issuecomment-118194746).

Adding this to my AppServiceProvider's boot method fixed it for me.

`Schema::getConnection()->getDoctrineSchemaManager()->getDatabasePlatform()->registerDoctrineTypeMapping('enum', 'blob');`

fletch3555:
@marktopper, is this worth adding to the ServiceProvider in https://github.com/larapack/doctrine-support? Not sure if you already have something there or if it needs modifying

marktopper:
I will get this added to the Doctrine Support service provider.

marktopper:
This have been fixed in `larapack/doctrine-support:v0.1.4`.
https://github.com/larapack/doctrine-support/commit/02c9fc73324d9111bb03bb8b087653afde8972e0

Please run `composer update larapack/doctrine-support`

BeetleJ:
Any word on this issue. Seems to still not be working correctly. Still getting "This type is not supported".

fletch3555:
Not sure what word you're looking for.  The issue was closed 22 days ago with a comment of "please run composer update ...."

BeetleJ:
I did the composer update larapack/doctrine-support but still get the error "generic.exception: Enum type is not supported", and I am unable to add any new columns to a table that uses ENUM.

PR #1775, opened 2017-09-20
Fix for #1766

mortenscheel:
Not sure I've added this in the right place, or if there are better ways to check the DB connection type, but it fixes an issue that occurs when rolling back the migrations containing enum types on MySQL connections. I have no clue if this could have any side effects, but it seems to work fine.


marktopper:
Thanks, but will close this and add this to our package for handling our ENUM support.
https://github.com/larapack/doctrine-support

Issue #1684, opened 2017-09-08
Permissions and menu items


jargoud:
- Laravel Version: 5.5
- Voyager Version: 1.0 dev
- PHP Version: 7.1.1
- Database Driver & Version: MySQL 5.6.12

### Description:

Some items in the sidebar menu are not hidden, as they should according to the permissions I defined in the Roles page.
I noticed this problem for the Tools dropdown (if empty, it is visible but empty) and the Settings and Hooks pages.

### Steps To Reproduce:

From a fresh installation, go to roles and edit permissions for any of them: uncheck _Browse Database_, _Browse Compass_, _Menus_, _Settings_ and _Admin/Browse Hooks_ permissions.
After signing in with a user having this edited role, the _Tools_ dropdown, _Settings_ and _Hooks_ items are still visible.

evrend:
- Laravel Version: 5.5
- Voyager Version: 1.0 dev
- PHP Version: 7.1.5
- Database Driver & Version: MySQL 5.7.17

I have same problem.

evrend:
@fletch3555 , @jeremyargoud Looks like I've fixed the problem. But the code that checks the permissions must refactor to a fuction in package. I dont know where can i put it. So i dont send pull request. 

vendor/tcg/voyager/resources/views/menu/admin_menu.blade.php
```
<ul class="nav navbar-nav">

@php
    if (Voyager::translatable($items)) {
        $items = $items->load('translations');
    }

@endphp

@foreach ($items as $item)

    @php
        $originalItem = $item;
        if (Voyager::translatable($item)) {
            $item = $item->translate($options->locale);
        }

        // TODO - still a bit ugly - can move some of this stuff off to a helper in the future.
        $listItemClass = [];
        $styles = null;
        $linkAttributes = null;

        if(url($item->link()) == url()->current())
        {
            array_push($listItemClass,'active');
        }

        //If have children, default false. we will check childs permissions!
        if(!$originalItem->children->isEmpty()) {
            $show_menu_item = false;
        } else {
            $show_menu_item = true;
        }

        // With Children Attributes
        if(!$originalItem->children->isEmpty())
        {
            foreach($originalItem->children as $children)
            {
                if(url($children->link()) == url()->current())
                {
                    array_push($listItemClass,'active');
                }

                // Childrens Permission Checker
                $self_prefix = str_replace('/', '\/', $options->user->prefix);
[...]

PR #1783, opened 2017-09-20
Fix for 'Permissions and menu items #1684'

evrend:
@fletch3555 for you to review.

### DIFF
--- a/resources/views/menu/admin_menu.blade.php
+++ b/resources/views/menu/admin_menu.blade.php
@@ -8,7 +8,7 @@
 @endphp
 
 @foreach ($items as $item)
-    
+
     @php
         $originalItem = $item;
         if (Voyager::translatable($item)) {
@@ -25,6 +25,13 @@
             array_push($listItemClass,'active');
         }
 
+        //If have children, default false. we will check childs permissions!
+        if(!$originalItem->children->isEmpty()) {
+            $show_menu_item = false;
+        } else {
+            $show_menu_item = true;
+        }
+
         // With Children Attributes
         if(!$originalItem->children->isEmpty())
         {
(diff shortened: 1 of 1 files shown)


resources/views/menu/admin_menu.blade.php
```diff
@@ -34,6 +41,36 @@
                 {
                     array_push($listItemClass,'active');
                 }
+
+                // Childrens Permission Checker
```
fletch3555, line 45: I would really prefer to avoid duplicating this entire section... Not sure how best to approach that at the moment though.  Open for suggestions.

resources/views/menu/admin_menu.blade.php
```diff
@@ -34,6 +41,36 @@
...
+                if ($slug != '') {
+                    // Get dataType using slug
+                    $dataType = $options->user->dataTypes->first(function ($value) use ($slug) {
+                        return $value->slug == $slug;
+                    });
+
+                    if ($dataType) {
+                        // Check if datatype permission exist
+                        $exist = $options->user->permissions->first(function ($value) use ($dataType) {
+                            return $value->key == 'browse_'.$dataType->name;
```
fletch3555, line 57: This is checking the permissions directly instead of going through the auth system... Old way of doing it, needs updating.  I'll tack that onto my to-do list

fletch3555:
I'm thinking this whole thing needs a rewrite... it's a mess right now (not what you did, but what you started from).  I'm going to hold off on this for now until I can decide how best to handle this.  Thanks though!

evrend:
I know, Code is mess right now. But it works. :) I'm waiting forward for fix with clean code. Thank you for your fast feedbacks.

fletch3555:
We got an alternate fix (rewrite) of this in last week.  Closing

Issue #1751, opened 2017-09-19
Multilanguage bug


vaggelis2018:
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7.1.8
- Database Driver & Version: MariaDB 10.2

### Description:
When you enable multilanguage it comes up a weird bug. If i select to edit a post for example, and select from english to go to greek fields are showing up the greek translations, but if i return back to english it still shows on all fields except rich text the greek values and if i click update english translations are getting greek values.

### Steps To Reproduce:
Just enable multilanguage and create a post. Then try to edit it and switch from on language to another and then backwards.

akazorg:
I confirm this bug on Posts, however I was able to edit a `Category`.

Also not working:
- Browse view doesn't switch language when click the selector.
- Sidebar sub-nav toggle not working when multilingual is enable.

marktopper:
This is fixed in #1802

PR #1802, opened 2017-09-21
Fix multilingual

akazorg:
Some issues related on #1751.

Will fix Multilingual on view Browse and Posts edit-add.


marktopper:
Thanks

PR #1811, opened 2017-09-21
Fix: Order menu children by 'order' attribute

mdelally:
Noticed that the order of menu children did not reflect the 'order' field being set by the menu builder. Not the most elegant fix, but it works to ensure that the children retrieved from the `menu()` helper are sorted properly, instead of by their primary key.

I will note that I tried to apply this 'query scope' to the `children()` recursive relationship on the MenuItem model, but it did not affect the render order of the menu items, so I moved it into the `display()` method on the Menu model.


marktopper:
Thanks!

PR #1813, opened 2017-09-21
Refactoring admin_menu

fletch3555:
* Admin menu now uses Laravel auth instead of checking permissions directly.
* Parent menus will hide if all children are hidden.
* Cleaned up code a bit

### DIFF
--- a/resources/views/menu/admin_menu.blade.php
+++ b/resources/views/menu/admin_menu.blade.php
@@ -4,75 +4,53 @@
     if (Voyager::translatable($items)) {
         $items = $items->load('translations');
     }
-
 @endphp
 
 @foreach ($items as $item)
-    
     @php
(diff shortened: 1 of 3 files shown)


marktopper:
Thanks @fletch3555, works like a charm

evrend:
Thank you @fletch3555, good job

reficul:
I don't know if this commit fix also what I found now, I report to you @fletch3555.

I try to put all items of the default voyager menu in a parent category named "Voyager". After that I notice that menu doesn't work on 2+ level of nested items. In my example I can't able to expand 'tools' because inside 'voyager' item.

fletch3555:
That's a known issue and I think there's an open issue for it.  Please try to find it and comment there, or open a new issue if you can't find one.

PR #1821, opened 2017-09-22
use absolute url when populating rows from table

s4muel:
i had project installed in a subdirectory (with the document root probably not set-up correctly) and everything worked out of the box but this "rows populating" because of 404s. using absolute URLs solves my problem.

### DIFF
--- a/resources/views/tools/database/edit-add-bread.blade.php
+++ b/resources/views/tools/database/edit-add-bread.blade.php
@@ -475,7 +475,7 @@
             var tbl = dropdown.data('table');
             var selected_value = dropdown.data('selected');
             if(tbl.length != 0){
-                $.get('{{ route('voyager.database.index', [], false) }}/' + tbl, function(data){
+                $.get('{{ route('voyager.database.index', [], true) }}/' + tbl, function(data){
                     $(dropdown).empty();
                     for (var option in data) {
                        $('<option/>', {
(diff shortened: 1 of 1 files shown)


fletch3555:
Absolute urls only add the protocol and domain to the generated url, so it shouldn't fix the problem you describe.

Can you share the url you use to access the site, the url(s) shown in the 404 error(s), and your APP_URL value (feel free to anonymize your domain if you wish)

fletch3555:
Sounds almost like your APP_URL was misconfigured.  I'd rather avoid using absolute URLs, since Laravel should be generating relative urls that function correctly.  I can wait for you to get back to your desktop to get the actual URLs

s4muel:
here you go:
APP_URL=http://localhost/project/public


random search for "route" in voyager project yields couple of usages, where abolute urls are generated
https://github.com/the-control-group/voyager/blob/dcfe7918c7023e795cd844edeebf48a4867ba911/resources/views/dashboard/navbar.blade.php#L13
https://github.com/the-control-group/voyager/blob/56794013fdda551bc882a9bd3ec2539be39a99eb/resources/views/menus/builder.blade.php#L44
https://github.com/the-control-group/voyager/blob/18ecff8cfa2f5adef5d57bab71fa9848d1e1ab12/resources/views/compass/includes/commands.blade.php#L12
...and more.

here is the route() helper definition (generates absolute urls by default):
https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/helpers.php#L778

the rest of the voyager app seems to be working OK, because the absolute urls are used all over it.

i made it working for me, so it is not a big deal, but is there anything i can help you with, or give you any more info?

Rct567:
I had the same issue, see https://github.com/the-control-group/voyager/issues/2038 (the problem is a relative url like "/" or "/admin" points to the domain and not the project folder)

PR #1828, opened 2017-09-22
Implementation of formfields_custom inside Relationship

ThomasBUEE:
Hi,

The new relationship in 1.0 is great but sometimes it's interesting to use custom form fields.
So I implemented this option inside BREAD relationships

Let me know what you think.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -2,7 +2,7 @@
 	
 	@if(class_exists($options->model))
 
-		@php $relationshipField = @$options->column @endphp
+		@php $relationshipField = (@$options->type=="belongsToMany" ? $row->field : @$options->column) @endphp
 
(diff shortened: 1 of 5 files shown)


ThomasBUEE:
Sorry, I forgot to create a new branch to submit a new PR. Let me know if you want a new branch for this issue.

The last commit, resolve issue to multiple belongsToMany relationships in the same BREAD.
The field that used the column property, so it's mostly the same name "id" for each relationship, I changed to relationship key which it's most appropriate.

fletch3555:
@ThomasBUEE, please confirm that all 4 commits showing in this PR are intended to be included in this.  If not, we would greatly prefer a separate branch to minimize the cross-talk.

Thanks!

ThomasBUEE:
I revert the last commit who's not concerne by this topic and created a new PR for this issue.

fletch3555:
Okay perfect. I'll close this one then.

PR #1829, opened 2017-09-22
Add hook permission

marktopper:


fletch3555:
This should probably be added to the Hooks package instead, so Voyager isn't as tightly coupled to hooks

marktopper:
The menu item is already here.
https://github.com/the-control-group/voyager/blob/master/publishable/database/seeds/MenuItemsTableSeeder.php#L197

fletch3555:
I'm going to close this in favor of consolidating that code into the Hooks package

marktopper:
I will fix something up.

akazorg:
@fletch3555 It's a total nonsense having Hooks documentation on Core, and not having Menu + Permission. This should be approved.

fletch3555:
@akazorg I agree it's total nonsense, but I disagree that Voyager needs to be the source of that permission.  I suggested @marktopper have the voyager-hooks package add the permission, which makes much more sense, since that's the package that integrates Hooks into Voyager

akazorg:
We need to fix this. Right now Hooks aren't working OOTB and we are shipping Voyager with it.
This PR fix it right away... I would had merge it and fix the problem.

Let's fix this guys. ^^

PR #1838, opened 2017-09-23
Wording correction

gausam:
Should read "will you" instead of "you will". Also added a missing apostrophe.


fletch3555:
As a fellow grammar-nazi, I applaud all grammatical fixes! 👍

PR #1854, opened 2017-09-24
Show little thumbnails for multiple_images field

marcelosantos89:
I also forced to only show 4 or it would grow to much


handiwijoyo:
Sorry, I make mistake when resolving the conflict and merging the `1.0` into your `master` branch. I will open another PR for this.

PR #1867, opened 2017-09-25
If the multiple_images upload is null and it has a current images kee…

Tarnaph:
If the multiple_images upload is null and it has a current images keep the current images


zolotykh:
When it will be merged?

Issue #1647, opened 2017-09-02
Double relationship to the same table issue


reficul:
- Laravel Version: 5.4
- Voyager Version: 11.14
- PHP Version: 5.6
- Database Driver & Version:

### Description:
when you have two fields with a relationship to the same table in browse view the slug show the first selection also in the second field. 

### Steps To Reproduce:
Es.
Table Groups (id, admin, moderator...)
admin -> hasOne(User::class, 'id');
moderator -> hasOne(User::class, 'id');

I select "Tom" for admin and "Bob" for moderator
In browse view I see "tom" in admin and in moderator field.
It works in edit view.

reficul:
![schermata 2017-09-09 alle 08 44 09](https://user-images.githubusercontent.com/1073883/30237839-281b7c0e-953b-11e7-8699-92e063814f0b.png)
![schermata 2017-09-09 alle 08 43 27](https://user-images.githubusercontent.com/1073883/30237841-284aa0c4-953b-11e7-9edf-3e8950c1dbe3.png)
![schermata 2017-09-09 alle 08 43 16](https://user-images.githubusercontent.com/1073883/30237840-28462cb0-953b-11e7-824d-78c1e755a3c7.png)

Screenshots example

fletch3555:
Fixed by #1875

Issue #1650, opened 2017-09-02
use different key rather than id for relationships


xcaptain:
- Laravel Version: 5.5.0
- Voyager Version: dev-master
- PHP Version: 7.1
- Database Driver & Version: mariadb 10

### Description:
when edit an item with more than one belongsToMany relationships, because every name is `id`, so every item uses the same value.

### Steps To Reproduce:
1. add a belongsToMany relationShip to a table
2. add another belongsToMany relationShip on it.
3. edit a row in this table, this 2 relationship get same value.

xcaptain:
sorry I will close this because duplicated with https://github.com/the-control-group/voyager/issues/1647

PR #1875, opened 2017-09-26
Fix belongsToMany formfields names

ThomasBUEE:
Fix issue when you have multiple belongsToMany relationship in the same BREAD, fields have same names "id[]"
So, I think the best way is to use the key of the relationship.


josiasmontag:
I verified that this fixes #1647 and #1650.

PR #1885, opened 2017-09-26
Fix many queries to storage for get files info

dmekhov:
Previously, for each file there were 3 requests. If we have 100 files in a directory, for example, it will be 300 requests to the server. If you use remote storage - this creates large delays.

Now one request receives the necessary information for all files in the directory.


fletch3555:
This looks good to me, but I haven't had a chance to test it yet.  @tnylea, you built the media manager.  Would you mind taking a look?

handiwijoyo (approved):
Tested and working as expected.

PR #1890, opened 2017-09-27
Use $model->getKeyName() instead of hard-coded "id" key.

madzimai:
Restore [flexible behavior](https://github.com/the-control-group/voyager/blob/10622dc41248ac45747376a873ce3519f92d6f0d/src/Http/Controllers/VoyagerBreadController.php#L48) from v0.11


fletch3555:
Good catch. Thanks!

PR #1891, opened 2017-09-27
Enable Route Model Binding with the slugs instead of IDs

JapSeyz:


fletch3555:
Thank you for this @JapSeyz.  You're welcome to implement this yourself in your project.  It would add a bit of unnecessary complexity for us to add to the Voyager core.

Issue #1842, opened 2017-09-23
No Allowance For `BelongsToMany` Pivot Table In Relationships Form Field


Funso-Popoola:
- Laravel Version: 5.4.*
- Voyager Version: 1.0.0
- PHP Version: 7.1
- Database Driver & Version:
MySql
### Description:
The failure to pass in the pivot table registered with the `BelongsToMany` relationship while trying to browse, read or edit a data type with the said `BelongsToMany` relationship. This causes laravel to assume a pivot table which might not be correct, hence, an error.

### Steps To Reproduce:
- Create a `BelongsToMany` relationship that requires a pivot table for one of your DataTypes
- Try viewing the index of the DataType to browse the list of items

Funso-Popoola:
I was able to fix this by overriding the `formfields/relationship.blade.php` view and changing the occurrences of this call

`$dataTypeContent->belongsToMany($options->model)`

under the `@elseif($options->type == 'belongsToMany')` check

to 

`$dataTypeContent->belongsToMany($options->model, $options->pivot_table)`

Funso-Popoola:
I suppose this could be fixed with a PR, right? Or am I missing something?

marktopper:
Solved in #1892

PR #1892, opened 2017-09-27
relationship pivot fix

WolandV:
I have problem with new relationships in 1.0. Then create relationship, I set pivot table, but that value ignored. Then creating new bread element, I have an error: `Table '#table1#_#table2#' doesn't exist `, but in settings there is table `#table2#_#table1#`. In 0.11.14 I could fix it easy:
return $this->belongsToMany(Project::class, '#table2#_#table1#');
in model class.
That solution for 1.0

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -139,7 +139,7 @@
 				<select class="form-control select2" name="{{ $relationshipField }}[]" multiple>
 					
 			            @php 
-			            	$selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model)->pluck($options->key)->all() : array();
+					$selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->pluck($options->key)->all() : array();
 			                $relationshipOptions = app($options->model)->all();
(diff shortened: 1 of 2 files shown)


josiasmontag:
Solves #1842.
Can we get this merged?

fletch3555:
I'm good with this.  Thanks @WolandV!

resources/views/formfields/relationship.blade.php
```diff
@@ -139,7 +139,7 @@
 				<select class="form-control select2" name="{{ $relationshipField }}[]" multiple>
 					
 			            @php 
-			            	$selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model)->pluck($options->key)->all() : array();
+					$selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->pluck($options->key)->all() : array();
```
ghost, line 142: The Best Way Is : 
$selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->pluck(**$options->table.'.'.$options->key**)->all() : array();

because if you have the same name of (id) in both of tables you will get a bug that column 'id' in field list is ambiguous

PR #1919, opened 2017-10-01
Add nullable belongsTo relationship support

alexjoffroy:
Currently if you set a belongsTo relationship on a BREAD, then you are forced to select a value in the dropdown.

This PR adds the ability do define a null value on a belongsTo relationship. Let's take an example with `categories`, saying a category can have a parent category (but not required):
- Create the belongsTo relationship with the UI
- On your data_row, in the `details` column, add `"null": "None"` (None or whatever you want) like this : 
```
{"null":"None","model":"TCG\\Voyager\\Models\\Category","table":"categories","type":"belongsTo","column":"parent_id","key":"id","label":"name","pivot_table":"categories","pivot":"0"}
```
- Then when you add/edit a category, you get this:
capture d ecran 2017-10-01 a 16 13 38
- If you select a parent category, you can remove it with the little cross:
capture d ecran 2017-10-01 a 16 13 25
- On the browse view, you get this:
capture d ecran 2017-10-01 a 16 13 02

I think the next step is to give the ability to edit this directly in the Voyager UI.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -1,5 +1,5 @@
 @if(isset($options->model) && isset($options->type))
-	
+
 	@if(class_exists($options->model))
 
 		@php $relationshipField = @$options->column @endphp
@@ -8,7 +8,7 @@
 
 			@if(isset($view) && ($view == 'browse' || $view == 'read'))
 
-				@php 
+				@php
 					$relationshipData = (isset($data)) ? $data : $dataTypeContent;
 					$model = app($options->model);
             		$query = $model::find($relationshipData->{$options->column});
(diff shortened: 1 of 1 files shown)


fletch3555:
Please avoid committing code-formatting/whitespace changes.  They only clutter the diff when viewing the PR, and StyleCI should be catching/fixing them anyway.

alexjoffroy:
Pretty sure my editor auto fixed the files following the rules in the project root (https://github.com/the-control-group/voyager/blob/master/.php_cs). But it's weird that it had to do it, if StyleCI already did it.

Anyway, you can view the diff without withespaces changes here:
https://github.com/the-control-group/voyager/pull/1919/files?w=1

OliverVin:
@alexjoffroy @fletch3555 Default value for relationship is urgency .  Sometime, selectdropdown no need choosing , so default value is needed .

OliverVin:
Also it works , but this is not a best way, set option should't change database directly !

alexjoffroy:
So, what about this PR ? Is there something more I can do to help ?

@OliverVin I'm not sure I got your points ?

handiwijoyo:
@fletch3555 is this still relevant? If not can we close this?

PR #1920, opened 2017-10-01
Trigger fallback when locale value is empty

fyhring:
The translation fallback is only triggered when the translation doesn't exists. With this the locale translation must exists and not be empty.

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -5147,11 +5147,6 @@ float: left;
   margin-left:10px;
 }
 
-.modal-open .select2-container{
-  z-index: 999999 !important;
-  width:100% !important;
-}
-
 .toggle-group .btn, .toggle-group .btn.btn-default{
   padding-bottom:0px;
   margin-top:0px;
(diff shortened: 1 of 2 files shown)


resources/assets/sass/app.scss
```diff
@@ -5147,11 +5147,6 @@ float: left;
   margin-left:10px;
 }
 
-.modal-open .select2-container{
```
fletch3555, line 5150: Why is this removed? Is it related to your change somehow?  If not, please revert this change

fletch3555: @fyhring ^

fyhring: Sorry. I made a mistake. I meant to make two diffent PR's. I'll try to revert.

fletch3555: No worries.  Always do work in branches.  PRs aren't a "snapshot" of the changes at that point, they're a complete list of commits that differ between the source and destination branches and are therefore updated automatically if new commits are pushed.

PR #1931, opened 2017-10-04
Change section to group

gizburdt:
In the admin area, setting groups are called groups everywhere, except in the example at the top. This PR will change `section` to `group`


fletch3555:
so... what's wrong with `section` that we need to change it to `group`...?

gizburdt:
There's nothing "wrong" with `section`, but it is called `group` everywhere else in the admin area (and database). So I thought, maybe change it to `group` to make it more clear and consistent for users.

fletch3555:
Okay, sounds good!

I just like to be challenging.  It gets people to defend their reasoning, which helps the rest of us understand better.

PR #1935, opened 2017-10-05
Code cleanup

ashish-k-kataria:


fletch3555:
These are unnecessary changes with no impact on anything else.  I prefer to do these types of changes in large sweeps through the codebase or as part of a refactor for a single feature.

Thank you though

PR #1937, opened 2017-10-05
Modify $content variable.

chrisifm:
Modify $content variable.


fletch3555:
You need to explain WHY you're making changes.  Either a full description or a link to an issue with a brief explanation of why your change fixes the problem would be appropriate.

PR #1955, opened 2017-10-07
Publishable views

jeffochoa:
Currently, there's not an easy way to make some tweaks to the admin views. Making those views publishable will make this so much easier :smile:

### DIFF
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -252,7 +252,7 @@ class VoyagerServiceProvider extends ServiceProvider
                 "{$publishablePath}/lang/" => base_path('resources/lang/'),
             ],
             'views' => [
-                 dirname(__DIR__).'/resources/views' => base_path('resources/views/voyager/')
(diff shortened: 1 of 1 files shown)


akazorg:
Hi @jeffochoa, I think you are able to change any view.
Please check docs: https://voyager.readme.io/docs/overriding-views, and let us know if that work for you.
Thank you!

jeffochoa:
Hi, @akazorg !
Yes, you can do that, but is kind of limited, don't you think? What would happen if you want to change any other view besides the "bread" views, like the menu, the main layout, etc...

akazorg:
You have a point. It's limited. I thought we were able to change that already.
We should merge this.
Thank you for the PR! :)

fletch3555:
No, it's not limited... you can change any view that way, not just the BREAD views. This functionality is part of Laravel, not specific to Voyager. (See subsection "Overriding package views" here: https://laravel.com/docs/5.5/packages#views)

PR #1966, opened 2017-10-09
change default user.php path.

ztzhu2012:
(app\User.php to app\Modules\User.php)

1.change app\User.php;
<code>namespace App\Models;</code>

2.change config\auth.php;
<code>'model' => App\Models\User::class,</code>

3.fix this branch.


marktopper:
The default class name for Laravel's User model is `App\User`.
https://github.com/laravel/laravel/blob/master/app/User.php

Issue #1947, opened 2017-10-05
Artisan command not working 


nanadjei:
After executing `php artisan down` form the compass `php artisan up` does not work.  Still get *Be right back*.

davodavodavo3:
@nanadjei2 you want to bring alive application from compass?
Or you run php artisan up from console and it doesnt remove down file from Laravel storage path?

nanadjei:
@davodavodavo3 I wanted to bring the app alive from compass but it did not work until I run `php artisan up`  from the terminal.

fletch3555:
That actually makes sense, since Voyager is part of the application you just took down... We'll need to find a way to allow it anyway, provide some kind of notification to users, or just remove those commands from Compass... @tnylea, any thoughts?

fletch3555:
Closing due to age and because this is behaving as expected.

PR #1967, opened 2017-10-10
added voyagercheckmaintenancemode middleware

allenjd3:
This is a potential solution to #1947

This would require some method of stubbing the Kernel when you install voyager. The kernel would need to register this checkmaintenancemode middleware rather than the default- something like-

    protected $middleware = [
        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
        \App\Http\Middleware\TrimStrings::class,
        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
        \App\Http\Middleware\TrustProxies::class,
    ];

    /**
     * The application's route middleware groups.
     *
     * @var array
     */
    protected $middlewareGroups = [
        'web' => [
            \TCG\Voyager\Http\Middleware\VoyagerCheckForMaintenanceMode::class,
            \App\Http\Middleware\EncryptCookies::class,
            \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
            \Illuminate\Session\Middleware\StartSession::class,
            // \Illuminate\Session\Middleware\AuthenticateSession::class,
            \Illuminate\View\Middleware\ShareErrorsFromSession::class,
            \App\Http\Middleware\VerifyCsrfToken::class,
            \Illuminate\Routing\Middleware\SubstituteBindings::class,
        ],

        'api' => [
            'throttle:60,1',
            'bindings',
        ],
    ];

notice how I removed the default checkForMaintenanceMode middleware from the middleware array and added the voyager one to the web group. (It can't be in the middleware group because I need access to the request object). 

Anyway, Is there a way to stub out the Kernel or would that have other potential side affects?

### DIFF
--- a/src/Http/Middleware/VoyagerCheckForMaintenanceMode.php
+++ b/src/Http/Middleware/VoyagerCheckForMaintenanceMode.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace TCG\Voyager\Http\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Route;
+use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Foundation\Http\Exceptions\MaintenanceModeException;
+
+class VoyagerCheckForMaintenanceMode
+{
+    
+    protected $whitelist = [
+        "voyager.dashboard",
+        "voyager.compass.index",
+        "voyager.compass.post"
+    ];
+    /**
+     * The application implementation.
+     *
+     * @var \Illuminate\Contracts\Foundation\Application
+     */
+    protected $app;
+
+    /**
+     * Create a new middleware instance.
+     *
+     * @param  \Illuminate\Contracts\Foundation\Application  $app
+     * @return void
+     */
+    public function __construct(Application $app)
+    {
+        $this->app = $app;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     *
+     * @throws \Symfony\Component\HttpKernel\Exception\HttpException
+     */
+    public function handle($request, Closure $next)
+    {
+
+        
+        if ($this->app->isDownForMaintenance()) {
+            
+            if(!in_array($request->route()->getName(), $this->whitelist)){
+                $data = json_decode(file_get_contents($this->app->storagePath().'/framework/down'), true);
+
+                throw new MaintenanceModeException($data['time'], $data['retry'], $data['message']);
+            }
+
+        }
+
+        return $next($request);
+    }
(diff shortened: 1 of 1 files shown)


allenjd3:
Looks like I didn't take out the Route Facade that I ended up not using. I can resubmit if necessary

fletch3555:
I'm not sure how I feel about this fix, though it would very likely work.  At the very least, you should probably be extending the laravel middleware instead of creating one outright.

allenjd3:
There are a couple ways it could be used- it could be an optional feature that you opt into by changing your own kernel, it could be a stub that you optionally install to your own middleware folder so you have access to the whitelist-

I could maybe change it to extend laravel's checkformaintenancemode middleware but I have a feeling I would be overriding everything anyway.

fletch3555:
If you extend the built in one, you won't need to do half of what you have.  You could do something along the lines of:

```
if (/* current route is in voyager */) {
    return $next($request);
}
else {
    return parent::handle($request, $next);
}
```

allenjd3:
Ah I see,

The only potential downside I see on that would be that you would be checking every request against the whitelist rather than just the ones that are down for maintenance. I'm not positive but it seems like that would be slower. I'll mess around with it some tonight to try and extend the laravel one.

allenjd3:
I extended the checkformaintenancemode middleware and cleaned some stuff up. I made it a middleware that runs after the request so you can just replace the default middleware with the one thats extended

    protected $middleware = [
        \TCG\Voyager\Http\Middleware\VoyagerCheckForMaintenanceMode::class,
        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
        \App\Http\Middleware\TrimStrings::class,
        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
    ];

I left the reference to app in there so it can run the check against the whitelist only on the requests that are in maintenance mode.

fletch3555:
You can't run it after the request.  The whole point of maintenance mode is to prevent requests to the application that will likely break during system maintenance (i.e. you have to take down a dependency and it'll just throw errors if the requests are allowed to proceed)

allenjd3:
You need access to the request object for it to work- might have to fix it a different way then

fletch3555:
You already have access to the $request object....

allenjd3:
It's null until you pass out of the default middleware group which is why I originally put it in the web group. I don't think laravel registers the request until after you get past that point

allenjd3:
https://stackoverflow.com/a/31868954

fletch3555:
I'm going to decline this for now.  I don't believe it needs to be in Voyager's Core.  Feel free to make it into a Hook.  I'm sure someone will find it useful.

Thank you.

PR #1984, opened 2017-10-12
Relationship Cleanup/Revision

akazorg:
Small revision.
Convert tabs to spaces, docs and small refactor.

### DIFF
--- a/src/Http/Controllers/Traits/BreadRelationshipParser.php
+++ b/src/Http/Controllers/Traits/BreadRelationshipParser.php
@@ -10,19 +10,26 @@ trait BreadRelationshipParser
 {
     protected $relation_field = [];
 
+    /**
+     * Remove Relationship Field
+     *
+     * @param  DataType $dataType
+     * @param  string   $bread_type BREAD type. Default 'browse'
+     */
     protected function removeRelationshipField(DataType $dataType, $bread_type = 'browse')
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/Traits/BreadRelationshipParser.php
```diff
@@ -10,19 +10,26 @@
...
+    /**
+     * Remove Relationship Field
+     *
+     * @param  DataType $dataType
+     * @param  string   $bread_type BREAD type. Default 'browse'
+     */
     protected function removeRelationshipField(DataType $dataType, $bread_type = 'browse')
     {
-        $forget_keys = [];
+        $_keys = [];  // Forget keys
```
fletch3555, line 21: Why all the field renaming? I'm perfectly content with the verbose names as they're actually meaningful.  There's a reason we don't name everything `a`, `b`, and `c`

akazorg: Some variable names are so long, harder to read. I renamed variables based on their scope, tried to abbreviate, but I guess it was radical in some cases..
Why don't we abbreviate obvious variables?
Tell what to do with this so we can close this. TY!

akazorg:
Closing. I had too much fun renaming files :D

PR #1997, opened 2017-10-13
Decouple Database and BREAD

akazorg:
Database has been decoupled from the BREAD system, and we have two sections now. 
We were mixing this subjects. 
This will allow us to enable/disable only one of the sections, giving more control to the app.

- Seeds for Menu/Permission added
- Languages Keys modified

### Don't forget to update your assets!
`$ php artisan vendor:publish --tag='lang' --provider='TCG\Voyager\VoyagerServiceProvider' --force`


marktopper:
Merging this into `develop`.
Great job @akazorg 🎉

PR #2004, opened 2017-10-13
Fix paging links  parameter error

dongyuhappy:


fletch3555:
Is there an open issue describing what this solves? I don't remember seeing one.

fletch3555:
This was resolved in #2246

PR #2007, opened 2017-10-14
Error when deleting image from multiple image field

darit:
The params fed into the modal were taking the table name as the slug. 
This fails to delete the image if the slug is different from the table name.

### DIFF
--- a/resources/views/tools/database/relationship-partial.blade.php
+++ b/resources/views/tools/database/relationship-partial.blade.php
@@ -41,8 +41,8 @@
                 <option value="belongsToMany" @if(isset($relationshipDetails->type) && $relationshipDetails->type == 'belongsToMany'){{ 'selected="selected"' }}@endif>Belongs To Many</option>
             </select>
             <select class="relationship_table select2" name="relationship_table_{{ $relationship['field'] }}">
-                @foreach($tables as $table)
-                    <option value="{{ $table }}" @if(isset($relationshipDetails->table) && $relationshipDetails->table == $table){{ 'selected="selected"' }}@endif>{{ ucfirst($table) }}</option>
+                @foreach($tables as $tbl)
+                    <option value="{{ $tbl }}" @if(isset($relationshipDetails->table) && $relationshipDetails->table == $tbl){{ 'selected="selected"' }}@endif>{{ ucfirst($tbl) }}</option>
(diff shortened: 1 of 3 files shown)


darit:
I'm in the zone... I'm very sorry if I should open a new PR for this small changes.

Changed the variable name used inside the relationships partial since it was colliding with the table name variable.
This caused to show the last table name from the select on line 45.

image

Added the multiple_images row type handling inside the Read since right now it displays the json.

image

resources/views/tools/database/relationship-partial.blade.php
```diff
@@ -41,8 +41,8 @@
                 <option value="belongsToMany" @if(isset($relationshipDetails->type) && $relationshipDetails->type == 'belongsToMany'){{ 'selected="selected"' }}@endif>Belongs To Many</option>
             </select>
             <select class="relationship_table select2" name="relationship_table_{{ $relationship['field'] }}">
-                @foreach($tables as $table)
-                    <option value="{{ $table }}" @if(isset($relationshipDetails->table) && $relationshipDetails->table == $table){{ 'selected="selected"' }}@endif>{{ ucfirst($table) }}</option>
+                @foreach($tables as $tbl)
```
fletch3555, line 44: Changing `$table` to `$tbl` doesn't seem necessary.  Was there a reason for this change?

darit: $table has the current table name which we are talking about on the relationship, if we assign again the $table inside the foreach, later when you echo the $table variable on line 52, you will have the name of the last table inside the  tables array instead of the name of the table we are using, resulting on this:

image

This same logic is already implemented on file relationship-new-modal.blade.php line 25.

fletch3555: ohhhh.... good call.  I was looking solely at the PR changes, so I didn't see that bigger picture.  Very good.  Merging soon

fletch3555:
I'm going to merge this since it's benign enough, but generally speaking, we prefer to avoid mixing concerns in the same PR.  You have what appears to be a few distinct changes all in the same PR.

Thanks!

darit:
Thanks! I'll do PRs from branches from now on to avoid mixing concerns. Voyager rocks, good work!

PR #1987, opened 2017-10-12
Add Events Support

akazorg:
(edit)
Adds Event Support to Voyager.
@marktopper is completing tests #1990. This can be used.

Currently is triggering events for the following, (more can be added later):

| Event         | Parameters      | On |
| ------------- |-------------| -----:|
| `BreadAdded` | $dataType, $data |  |
| `BreadUpdated` | $dataType, $data |  |
| `BreadDeleted` | $dataType, $data |  |
| `BreadChanged` | $dataType, $data, $changeType | Bread(Add/Update/Delete) |
| `BreadDataAdded` | $dataType, $data |  |
| `BreadDataUpdated` | $dataType, $data |  |
| `BreadDataDeleted` | $dataType, $data |  |
| `BreadDataChanged` | $dataType, $data, $changeType | BreadData(Add/Update/Delete) |
| `BreadImagesDeleted` | $data, $images |  |
| `FileDeleted` | $path |  |
| `TableAdded` | $table |  |
| `TableUpdated` | $name($array) |  |
| `TableDeleted` | $name(string) |  |
| `TableChanged` | $name(string) |  Table(Add/Update/Delete) |

## How to Use
Please follow the [Laravel documentation](https://laravel.com/docs/master/events#registering-event-subscribers). You can use it by creating your own Listeners, or by listening to a given event at the boot section of the `EventServiceProvider`.

## Example Listening to `BreadDataChanged`

On your `EventServiceProvider`, on the boot() method.
```
    /**
     * Register any events for your application.
     *
     * @return void
     */
    public function boot()
    {
        parent::boot();

        Event::listen('TCG\Voyager\Events\BreadDataChanged', function (...$args) {
            // Write some logic here
            // ------------------------------------
            // $args->dataType    ---> TCG\Voyager\Models\DataType
            // $args->data        ---> (Object) Data changed
            // $args->changeType  ---> (String) Added|Updated|Deleted
        });
    }
```

Happy coding!

PR #2013, opened 2017-10-14
Events Convert to Class

akazorg:
This will convert existing events to Classes, following the Events implemented on PR #1987.

### DIFF
--- a/routes/voyager.php
+++ b/routes/voyager.php
@@ -15,7 +15,7 @@ use TCG\Voyager\Models\DataType;
 */
 
 Route::group(['as' => 'voyager.'], function () {
-    event(new Routing($menu));
+    event(new Routing(app('router')));
 
     $namespacePrefix = '\\'.config('voyager.controllers.namespace').'\\';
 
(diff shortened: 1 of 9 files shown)


routes/voyager.php
```diff
@@ -13,15 +15,15 @@
 */
 
 Route::group(['as' => 'voyager.'], function () {
-    event('voyager.routing', app('router'));
+    event(new Routing());
```
fletch3555, line 18: Not sure what these events (in the routes file) are even for... @tnylea / @marktopper, thoughts?

akazorg: After a search on my local packages, I've found an hint with Hooks.

Hooks setup will be revised, me and @marktopper have a draft for Hooks also based on events, when we check that, maybe we can remove this event? 
Do you want to remove it Fletch? I don't know if is that heavy for a system firing events.

marktopper: Its used by the hook system and some other hooks from what I know. Maybe we can remove this in `develop`.

Issue #1921, opened 2017-10-01
FEATURE creating table creates a migration for it?


AbanoubNassem:
Feature request : creating a table , generate/create migration for it .

it's useful when you are developing offline , then moving to production , so simply, running migration on the host  will create the whole database tables needed.

fletch3555:
This is on our radar for a future release.  Building a migration is certainly not a small/simple task.  We're open for pull requests!

jeffochoa:
I did something like this in my current project. I may be able to create a pull request for this in the next days.

jeffochoa:
PR open :crossed_fingers: https://github.com/the-control-group/voyager/pull/2014

cyrildewit:
@fletch3555 can this issue be closed? @jeffochoa created a hook for this [jeffochoa/voyager-bread-generator](https://github.com/jeffochoa/voyager-bread-generator). Or do still want to add this to the core? 

Might be useful to re-read this pr: #2014

fletch3555:
That hook isn't the same as this.  This issue is for automatically creating a migration class that matches the table that was created.  That hook creates BREAD programmatically, but the table already exists.  The `--migration` flag just calls `artisan make:migration`, which only creates a stub file.

cyrildewit:
So when I create a table with Voyager, a migration file should be created with the corresponding columns. But how are we going to handle changes to the table that have been made after the initial `create_xxx_table` migration?

~I think a simple `Dump Migration` button or `php artisan voyager:table:dump`.. would be the easiest way. The developer can do with it what he wants. Like: comparing it with an old migration and copy paste the needed fields to create a migration that updates the table on the server.~

cyrildewit:
I just found out that the button already exists for this feature, so I will try to do an attempt in the next weeks/months.

Issue #1477, opened 2017-07-17
How will I create migrations for the admin tables at the same time when I use voyager admin to create tables (I mean the database tab)? 


adityabanerjee992:
- Laravel Version: 5.4.16
- Voyager Version: 0.11.9
- PHP Version: 7.1.2
- Database Driver & Version:

### Description:
Hi I have an issue I am creating the tables for the admin panel using the database tab in the voyager but I notice that migrations have not been created for the tables. Please assist me how will I create migrations at the same time when I create table using the voyager admin panel. 

Now the issue is when I store the project on git and some one takes the clone he does not get the migrations for the admin panel as it was not created. Please assist , thank you.

fletch3555:
Voyager does not currently support creating migrations.  It's a feature we're working on.  For now, you'll have to create the migrations manually.

PR #2014, opened 2017-10-14
Create BREAD (migration, seed and model) files using artisan

jeffochoa:
There is a common issue when we try to deploy local projects to a different environment. Currently, we need to export the database or so, in order to keep all the new BREADs structure across all the environments.

The only way to do that without having to create a database import each time is by creating the migrations, seeds, etc. for each bread.

This allows the developers to create new BREADs from the command line using Artisan.

## How to use:

1. create a new bread
`php artisan voyager:bread books`

You can also generate the model and migration files

`php artisan voyager:bread books --migration --model`

2. Configure the bread
This command will create a new `BooksBreadSeeder` file with the basic configuration for a new bread-seed, there you can add/edit all the bread fields. See [DataRowsTableSeeder](https://github.com/the-control-group/voyager/blob/master/publishable/database/seeds/DataRowsTableSeeder.php)

3. Once the seeder is done you need to run:
`php artisan db:seed --class=BooksBreadSeeder`

4. Optionally you need to re-generate the permissions from the command line
`php artisan db:seed --class=PermissionRoleTableSeeder`
You can also do this manually from the admin panel

5. Don't forget to run the new migration
`php artisan migrate`

## Related issues
[https://github.com/the-control-group/voyager/issues/1477](https://github.com/the-control-group/voyager/issues/1477)

[https://github.com/the-control-group/voyager/issues/1921](https://github.com/the-control-group/voyager/issues/1921)

### DIFF
--- a/src/Commands/BreadGenerator.php
+++ b/src/Commands/BreadGenerator.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace TCG\Voyager\Commands;
+
+use Illuminate\Support\Str;
+use Illuminate\Console\GeneratorCommand;
+
+class BreadGenerator extends GeneratorCommand
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'voyager:bread {name}
+                            {--migration : creates a new migration for this bread}
+                            {--model : creates a new model for this bread}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create a new Bread';
+
+    /**
+     * Execute the console command.
+     * @return mixed
+     */
+    public function handle()
+    {
+        $seederName = Str::studly(Str::plural($this->argument('name')).'BreadSeeder');
+        $this->info('Making BREAD');
+
+        parent::handle();
+
+        if ($this->option('migration')) {
+            $this->createMigration();
+        }
+
+        if ($this->option('model')) {
+            $this->createModel();
+        }
+
+        $this->info('You are almost ready: ');
+        $this->info("1. Once you finish the seed configuration, you need to run : php artisan db:seed --class={$seederName}");
+        $this->info('2. Optionally you may want to re-generate the permissions with: php artisan db:seed --class=PermissionRoleTableSeeder');
+    }
+
+    /**
+     * Build the class with the given name.
+     *
+     * @param  string  $name
+     * @return string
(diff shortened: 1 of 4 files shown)


fletch3555:
I'm not sure how I feel about this just yet.  We used to have the ability to create migrations, but got rid of it because laravel's default `make:migration` command doesn't do anything but create a shell file.  It looks like your seeder file is the same (only creating the default `id` and timestamp fields.

jeffochoa:
@fletch3555 Yes of course, but the main feature here is the bread `seeder` itself which allows you to perform all the needed tasks in order to register all the fields, menus, etc. on each new bread.

In fact, the `--migration` flag is totally optional

ShamarKellman:
I like this idea, I actually do this manually when working on a project, having a command to do it would be perfect.

akazorg:
This is interesting, but a delicate subject, as @fletch3555 mentioned.

What happen when we change a BREAD? it would generate seeders again?

jeffochoa:
Same thing as the users or posts breads I guess 😅

akazorg:
Any bread in fact.

It would be great to save all BREADs state, because of the deployment, seeds could be used.

Migrations are harder to handle. -.-

I like your script, maybe this is more suitable for a Hook. Do you know Hooks @jeffochoa?

jeffochoa:
Yes, I know about hooks  but if you look at the issues section,  this seems to be like a common struggle. So I think this would be a nice feature to have just right out of the box.😉

fletch3555:
I mean, just because it's a common request, doesn't mean it makes sense to add it to core.  That's why we added the hooks functionality, so you can add your own features and share it with everyone.

akazorg:
Well, whether we use it as a Hook or not, your code should be used xD

In my opinion creating Seeds for saving the state, we start entering, somehow, on the subject of a backup system, and this will open other issues into debate, that are not related to Voyager (core), and should be treated as a Hook.

I'm having a hard time to figure out what's best, but I guess this is more suitable for a Hook.

jeffochoa:
Here is the hook https://larapack.io/hooks/voyager-bread-generator 

But I still think this could be added to the voyager-core package :sweat_smile:

akazorg:
Great job! We can merge that hook on Voyager Core anytime!!
I'm looking forward to test that!

akazorg:
Closing since Hook has been created.

PR #2016, opened 2017-10-14
Feature | Autocreate Menu + Permission on Bread added

akazorg:
This will short the time when dealing with BREAD creation.

After you create a BREAD item, this will create the Menu and assign Permission to the admin, this way you can browse that page immediately without more actions.

A config parameter `autoadd_menu_permission` was added. Default is true.

Hope you like this feature.


src/Listeners/AddMenuPermission.php
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Will create Menu item and Permission after creating a BREAD item.
+     *
+     * @param BreadAdded $event
+     *
+     * @return void
+     */
+    public function handle(BreadAdded $bread)
+    {
+        $this->addPermission($bread);
```
fletch3555: Can we change this so it's 2 distinct listeners? One adds the menu, and one adds the permission?  It simplifies the code a bit by making each class just a bit more atomic.

fletch3555 (dismissed):

src/Listeners/AddMenuPermission.php
akazorg: Absolutely, great idea!

akazorg: I've committed the changes

PR #2032, opened 2017-10-16
FIX: Provide compatibility with model binding.

ghost:
Reproduce: setup model binding for an entity ( e.g. Bankaccount )

Voyager route's result in:
screen shot 2017-10-16 at 19 29 55

Laravel will automatically get the proper model instance.

Deleting, updating a model via Voyager will not work anymore. destroy and save methods do not exist on the findOrFail result (Collection).


ghost:
It seems that the show and edit methods need these changes aswell. Want me to revert this one and recreate a PR? @fletch3555

PR #2033, opened 2017-10-17
Fixing BelongsToMany functionality

tnylea:
The belongsToMany functionality was not working correctly because it was not saving the correct form field in the edit-add bread file.

This fixes that issue.


marktopper:
Seems like TravisCI fails with:
```
Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function relationLoaded() on null in /home/travis/build/the-control-group/voyager/src/Traits/VoyagerUser.php:53
```

tnylea:
@marktopper Seems to be passing now 👍 
Let me know when you get a chance to merge it in. Thanks!

I've also fixed the tinymce to include the bulleted and number lists, the new version of tinymce needed some additional functionality for these.

I've also fixed the file storage url download in this PR :)

marktopper:
Great, will do a test when I arrive home in less than an hour and merge in ;)

PR #2035, opened 2017-10-17
feat: Optional, action-specific validation rules. Closes #2034.

### LINKED ISSUE #2034: Action-specific validation rules
- Laravel Version: 5.5.14
- Voyager Version: 1.0.5 (dev-master)
- PHP Version: 7.1.9

### Description:
Currently, any defined validation rules will apply to all the major actions
(store and update).

This premise breaks when for example we have a Logo field which is required
when creating a new record, but is optional when updating the record (because
the initial image provided will still be on file). So with the current
mechanisms, the user will ALWAYS be asked by Voyager to re-upload the image
each time they want to update anything else.

The user profile editing in Voyager avoids this scenarios because of the
default avatar mechanism, but the same mechanism is not always an option.

### Steps To Reproduce:
- Create a new BREAD type with, say, an image field. In the validation, set the image to be required, in order that users will provide an image when creating the record.
- Create a new record, and the validation will kick in as it should that you must upload the image (which you then do)
- Go into edit mode for your newly created record, edit any other fields as appropriate, leave the image field as is (since you already uploaded), and press _Submit_
- Voyager will tell you the image is required

## Proposal
I propose a mechanism for allowing the universal validation rules we have at the moment, but also a way to define action-specific validation rules. In terms of validation rules, the above would have been easily solved if there was a way to "require" the image when creating and then just having plain "image" checking in the validation rule (without "require") when updating.

Briefly ran it past @marktopper [on Slack](https://laravel-voyager.slack.com/archives/C373UEX5Y/p1508166810000177?thread_ts=1508146444.000383&cid=C373UEX5Y), and have a preliminary nod from him.

Solution is rather simple (as you'll see in PR #2035), and the use case is quite common in any app, I'd imagine.

gausam:
_Closes #2034_.

Currently, any defined validation rules will apply to all the major actions
(store and update).

This premise breaks when for example we have a Logo field which is required
when creating a new record, but is optional when updating the record (because
the initial image provided will still be on file). So with the current
mechanisms, the user will ALWAYS be asked by Voyager to re-upload the image
each time they want to update anything else.

The user profile editing in Voyager escapes this scenario because of the
default avatar mechanism, but the same mechanism is not always an option for custom BREADs.

With this commit, users will be able to specify their universal validation as
before - there is no change to the current way things work.
But they will now also be able to specify action-specific validation rules.

Using the above scenario as an example, we can now do:

```
{
    "validation": {
        "store": {
            "rule": "required|image",
            "messages": {
                "required": "Please provide a logo",
                "image": "Please upload a valid image file",
            }
        },
        "update": {
            "rule": "image",
            "messages": {
                "image": "Please upload a valid image file",
            }
        }
    }
}
```


handiwijoyo:
This is a good proposal, but I think we need to more simple way to handle this rather than to write almost identical rules twice.

fletch3555:
Instead of `store` and `update` (while they're the default in laravel, not everyone uses that), what if we continue using the BREAD naming for consistency.  In your example above, this would change to `edit` and `add`

gausam:
@fletch3555 I agree with you. Will make the changes as requested.

fletch3555:
This will likely need to be merged into 1.x-dev, not 1.0.

Please note that we have a complete rewrite of the BREAD pages that's almost done and does include a solution to the problem you're solving here.  I can't say whether it'll make it into 1.1 or not, so I'm okay with leaving this open for now.

gausam:
That's great news. I solved my use case by way a fork, so if this is to be solved in a coming release then I'm not opposed to closing this.

fletch3555:
👍

PR #2036, opened 2017-10-17
Added ability to create custom permission per table

ratt:
Hi, 

Were busy implementing a workflow for certain roles that required some additional (custom) permissions. EG the ability to restrict publishing. As we work in a team these permissions get added from a seed

Usage example: Permission::generateCustomFor('post', 'publish')

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -22,6 +22,16 @@ class Permission extends Model
         self::firstOrCreate(['key' => 'delete_'.$table_name, 'table_name' => $table_name]);
     }
 
+    /**
+     * Generate custom permission for table
+     * @param $table_name
+     * @param $permission
+     */
+    public static function generateCustomFor($table_name, $permission)
+    {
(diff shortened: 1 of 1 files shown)


fletch3555:
Is there a reason you can't just create these manually in your migration?  I'm not sure our models need more one-off functions added, but I'm also not 100% opposed to it.

ratt:
Not at all, we just create seeds similar to how you do it to preload
permissions for our custom tables. So we simply followed the same example
you did but rather created a function. We have no issues if this is not
included and just thought it might be useful.

akazorg:
This feature is handy, we should merge this.

fletch3555:
I'm not opposed to it, but I don't think it's necessary.  It enables a one-liner for migrations, where 2 lines normally would've sufficed.

akazorg:
It's true, we don't need this, or we would be bloating Voyager with many handy methods.

Thanks for sharing @ratt.

PR #2042, opened 2017-10-17
add option for random uploaded image name

ripples-alive:
It's good for voyager to preserve name for uploaded image. But I think we should have an option to allow the usage of random image name. I don't want the original image name to be leaked to viewers  as it may contain sensitive information.
So I generate this pull request to add an option for random image name. As for the option name, I follow the name convention in d12a66a86f838340bae75cf8e063ecead380c912 by @david-mk which have been removed in 58478d81eee049ffeaa61085fa540ce1fce9ea6b by @tnylea .

BTW, I found that options in voyager use both snake case (e.g. pivot_table) and camel case (e.g. forceUpdate), it seems really strange.


ripples-alive:
Add maybe it's better if we can extract the file name generation into a separated function. I see other code like `multi_images` and `files` still use the random name now.

OliverVin:
who can merge into master and release , it is very important . @david-mk  @tnylea

fletch3555:
When did the other change get merged in...? THAT one should be controlled by the config, not what you've done.  What you added back in is what we've always had, so the other one is technically a breaking change... 

If you could invert the config so it enables the original filename instead of the random filename, I'd be glad to get this merged in.

OliverVin:
@fletch3555  you mean enable random filename by default . Now default original filename, if want random filename need set preserveFileUploadName option false.

ripples-alive:
@fletch3555 Sorry, I don't understand what you means by saying the other change, d12a66a or 58478d8?

> What you added back in is what we've always had

I just want to use random image name. If voyager could do this, it's ok, and please tell me how.

And about inverting the config, I can do this. But I'd like to make sure you want to use random filename as default option. It would be a breaking change as voyager use original filename now. Or it's just my misunderstanding?

fletch3555:
Yes, what I'm asking for is to have the original functionality (random filename) be the default.  You can then set the config value (named whatever seems appropriate, not limited to what we have here) to make it use the original uploaded filename instead.

The breaking change already occurred.  I'm trying to revert that.

I hope that clears up the request.

ripples-alive:
Got it, I'll do it right now.

fletch3555:
Perfect, thanks!

PR #2044, opened 2017-10-18
Fix/provide model binding compatibility

ghost:
Follow up: #2032 .

Added model-binding compatibility for the show + edit methods aswell.


fletch3555:
Good catch

PR #2064, opened 2017-10-20
Translate user admin profile menu

cesaramirez:
translate user profile admin menu

**Before (lang/en)**
screen shot 2017-10-20 at 14 56 45

**After (lang/es)**
screen shot 2017-10-20 at 14 57 06

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -186,8 +186,6 @@ return [
 
     'show_dev_tips' => true, // Show development tip "How To Use:" in Menu and Settings
 
-    'show_message_analytics' => true, // Show analytics message in dashboard
-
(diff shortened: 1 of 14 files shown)


fletch3555 (changes requested):
Please try to avoid having multiple unrelated changes in a single PR.  I'd love to get the translation fix in, but I don't really think the analytics change is necessary.  I can't merge it since they're both in the same PR.  Also, the config change needs to be pushed to a future version as it's new functionality that may break if users don't have the new config entry.  Please separate them and I'll be glad to review it again.

cesaramirez:
Thanks for your comments, I fixed

cesaramirez:
Commit revert where analytics logic was modified.

PR #2075, opened 2017-10-24
Fixed modal routing path formatter

### LINKED ISSUE #1949: Modal wrong action routing
https://github.com/the-control-group/voyager/blob/0b11c4c09d494ca72497ec0fc2e66dd21fa457d6/resources/views/menus/browse.blade.php#L111

Should be:
```
$('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.index') }}/' + id;
```

Because if not, it's appending the id of the menu constantly.

Steps:
1) Click delete in any menu
2) Click cancel on the modal.
3) Click delete in the same menu or any other.

Result, sometime like that: http://voyager.dev/admin/menus/3/1/3

sidis405:
Added logic to correctly format the menu path.

Code is coherent with the one on the browse.blade.php of Bread.

Closes #1949

### DIFF
--- a/resources/views/menus/browse.blade.php
+++ b/resources/views/menus/browse.blade.php
@@ -106,9 +106,14 @@
         });
 
         $('td').on('click', '.delete', function (e) {
-            id = $(e.target).data('id');
 
-            $('#delete_form')[0].action += '/' + id;
+            var form = $('#delete_form')[0];
+
+            var deleteFormAction = form.action;
+
+            form.action = deleteFormAction.match(/\/[0-9]+$/)
+                ? deleteFormAction.replace(/([0-9]+$)/, $(this).data('id'))
+                : deleteFormAction + '/' + $(this).data('id');
 
             $('#delete_modal').modal('show');
         });


resources/views/menus/browse.blade.php
```diff
@@ -106,9 +106,14 @@
...
 
         $('td').on('click', '.delete', function (e) {
-            id = $(e.target).data('id');
 
-            $('#delete_form')[0].action += '/' + id;
+            var form = $('#delete_form')[0];
+
+            var deleteFormAction = form.action;
+
+            form.action = deleteFormAction.match(/\/[0-9]+$/)
```
fletch3555, line 114: I'm not sure how I feel about this solution.  It locks us into mandating numerical ID values.

lucabartoli: I didn't check on this issue, but... Can't be anything except `/`?
`/\/[^\/]+$/`

fletch3555: That regex is essentially matching `/*` where `*` is 1 or more characters that cannot contain a `/`.  So it will match `/a`, `/---`, and `/abc123!@#`, but not `/`, `//`, `/diu/`, or `/abc/123/!@#`

PR #2077, opened 2017-10-24
Showing display name of bread record on validation error

### LINKED ISSUE #2005: The "required" validation's error message uses the real field name instead of the display name
- Laravel Version: 5.5.0
- Voyager Version: 1.0.5
- PHP Version: 7.1.8
- Database Driver & Version: MySQL 5.7.19

### Description:

I gave a field a display name and marked it as required. I would expect the error message to say "The [display name] field is required." but it actually says "The [real field name] field is required.".

### Steps To Reproduce:

This can be reproduced using Voyager's built-in "pages".

1. Go to /admin/database/pages/bread/edit
1. Change the `excerpt` field's display name to "snippet"
1. Set the `excerpt` field's JSON to the following:
    ```
    {
        "validation": {
            "rule": "required"
        }
    }
    ```
1. Submit the form
1. Go to /admin/pages/create
1. Fill out the "title" field and submit the form without filling out the "snippet" field
1. Observe that the error message for the "snippet" field is "The excerpt field is required.". Notice that it's referring to the field by its real name instead of the display name.

It appears that if you want to mark a field with a display name as required, you have to also specify the error message, like this:

```
{
    "validation": {
        "rule": "required",
        "messages": {
            "required": "The snippet field is required."
        }
    }
}
```

In other words, you have to have the display name in two places instead of just one.

sidis405:
A fix for this matter. Showing display name instead of field on validation error

Closes #2005


marktopper:
Thanks 🎉

PR #2086, opened 2017-10-26
More accurate italian translation

lucabartoli:
- cancel was wrong: Correct translation is "Annulla" and not "Cancella" which means delete.
- Unified syntax for "Cancella", "Elimina", "Rimuovi" which basically have the same meaning
- symlink_failed_title fixed


lucabartoli:
I recreated the pull request because I mixed different commits, sorry guys

PR #2110, opened 2017-10-29
Remove unnecessary else

yzi-afk:


fletch3555:
While you're technically correct that the else is unnecessary, this PR doesn't really add any value.  The existing code is just as readable, if not more so.

PR #2127, opened 2017-10-31
Fix the required logic

ChenAlon:
The previous change just removed the required attribute which is not the full solution.
The problem is that in the edit route, when you have a file and you try to make changes, you have to reselect the file again to save changes.


fletch3555:
You're 100% correct.  Sorry I missed this on the last one.  Good eye, thanks!

PR #2141, opened 2017-11-02
update belongsto relation save and update

raulsi:
The Relationship field using the name as something like 'documents_belongsto_products_relationship' instead of foreign key 'product_id' so on create and save it's not able to save the field. also not able to view the relationship to already set.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -22,13 +22,13 @@
 
 			@else
 			
-				<select class="form-control select2" name="{{ $relationshipField }}">
+				<select class="form-control select2" name="{{ $options->column }}">
 					@php 
 						$model = app($options->model);
 	            		$query = $model::all();
 	            	@endphp
 					@foreach($query as $relationshipData)
-						<option value="{{ $relationshipData->{$options->key} }}" @if($dataTypeContent->{$relationshipField} == $relationshipData->{$options->key}){{ 'selected="selected"' }}@endif>{{ $relationshipData->{$options->label} }}</option>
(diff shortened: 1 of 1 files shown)


ripkens (approved):
Works for me

DrudgeRajen:
I think this is the correct solution .

DrudgeRajen (approved):
Works for me as well.

baboot:
work for create and update . But when entering on a edit form page, values from reletions (in stored models) not setted in select2.

RavivCohen-BWEBI:
Had the same problem,
the input name got transformed to 
for example: rewards_belongsto_campaign_relationship 
instead of campaign_id (which is the table column name). 
Debugged it until resolved with a very simple solution of clearing the voyager views cache
just by playing around with
vendor/tcg/voyager/resources/views/formfields/relationship.blade.php

Still do not know what caused the input name initial change but it worked.

DrudgeRajen:
this issue persist in 1.0.10

coire1:
This issue persists in 1.0.11 also.
Reverting #2121 and keep this one #2141 solve the issue for me.

mihaitaranu:
screen shot 2018-01-14 at 16 58 05
checking all in author_id worked for me

szsori:
I can confirm that the solution presented by @mihaitaranu resolves this in 1.0.13, while other fixes did not (I didn't try rolling back the version).  It seems like the proper workaround is to check all of the boxes for the field used in the relationship, but a proper fix would include relationship fields in case the user doesn't make this manual change.

banh-gao:
Reverting #2121 kind of works but is not a solution as it introduces a regression that was fixed in #2121. 
Here the problem relies in the field column of the data_rows table.
The field name for belongsTo relationship has to be the name of the field to set for the model.

For example for my field `shipments.member_id` that sets a belongs to relationship with `members.id`, the `datarows.field` for the belongsto relationship should be just: `member_id`, but it is wrongly set to `shipment_belongsto_member_relationship`.

Also #2144 seems to report the same problem.

PR #2142, opened 2017-11-02
Fix language selector

viktormaar:
When multilanguage are enabled in the app, in the language switcher the selected language's button has the same color as the selectable one(s). This quick-fix solves this.


fletch3555:
I'm not sure this is the right solution.  Where did you get the colors for your CSS changes?

PR #2143, opened 2017-11-02
Fix hooks permission

viktormaar:
The permission for hooks is missing after a clean install, and need to add manually. This quick-fix solves this.


fletch3555:
Closing due to separation of concerns.  This permission needs to be added through the VoyagerHooks package, not Voyager

PR #2155, opened 2017-11-04
Relationship table selection overwrites $table

emptynick:
The table select overwrites the $table which is used on line 55 in a belongsto relationship.
Instead of the selected table it will show the last table from the foreach loop


fletch3555:
Thanks for this!

PR #2156, opened 2017-11-04
Show Images in browse page for input type "mutlple images"

reindeeruz:


emptynick:
I dont think thats the way to go.
I have a few hundred entries with 10-30 images each.
This would make the table break completely.

A carousel would be useful. Plus it would be nice for bread-browse and edit also.
Or take the first image and make it clickable to a lightbox gallery.

fletch3555:
I agree. I don't think we should be showing multiple images on browse.  Just too congested.

PR #2161, opened 2017-11-05
Fixed typo in translation key

Widcket:
On line 99, replaced `logginign` with `loggingin`. The translated string is now showing up correctly on the UI.


fletch3555:
This should be changed in master, not develop, since it's already fixed in other languages.

PR #2169, opened 2017-11-06
Browse page Rows display based on "read" access Policy.

dattz:
On BREAD **Browse** page, Currently using the policies on a BREAD, the "read" access for an item only toggles the "View" action button of that item.
That means, if a policy returns false for an item, the item is still displayed on the browse page - only its "View" button is hidden.

If the user doesn't have access to "read" the item, it makes it sense to just not show it on the Browse page also.
To do this, I have moved the `@can('read', ...)` directive to cover the whole `<tr>`.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -77,109 +77,109 @@
                                 </thead>
                                 <tbody>
                                     @foreach($dataTypeContent as $data)
-                                    <tr>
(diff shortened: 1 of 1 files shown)


dattz:
Oh! This is affecting the Server Side Pagination. Help?

dattz:
Not able to find a workaround. Closing this.

PR #2171, opened 2017-11-06
Restricted checkbox to only those who have delete access

dattz:
The checkbox currently appears on items on the browse page. Even the user who doesn't have "delete" access, sees the checkbox. The checkbox serves no purpose for that user.
Hence, this PR limits the checkbox and only shows it to the users that have the "delete" access.


dattz:
Bad PR. Closing. Please delete.

PR #2172, opened 2017-11-06
Restrict Browse Page checkboxes to "delete" access

dattz:
The browse page shows checkboxes for every item. They are used to select multiple items and "Bulk Delete" them.
If a user doesn't have a "delete" access on that BREAD, the checkboxes serve no real purpose.

Until they start serving some purpose, the users with no "delete" access shouldn't see them.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -53,7 +53,9 @@
                             <table id="dataTable" class="table table-hover">
                                 <thead>
                                     <tr>
-                                        <th></th>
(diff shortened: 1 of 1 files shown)


fletch3555:
Please confirm you've checked the other views (i.e. posts, pages, users, etc.) to make sure this applies to them as well, since some of them override the BREAD views.

dattz:
Yes, I checked. No other view overrides this _bread/browse.blade.php_

fletch3555:
Perfect, thanks.

fletch3555:
👍

PR #2174, opened 2017-11-06
Language file typo etc.

dattz:
Fixed singluar version of word "entries" (not entrie but its entry).
The "Bulk delete" button has a small "d", whereas the "Add New" button has both capital letters. Changing it to "Bulk Delete".


fletch3555:
I'm okay with it in this instance because the changes are so small and perfectly reasonable, but please be careful when submitting multiple unrelated changes in the same PR, since we won't be able to accept/deny one without the other.

Issue #69, opened 2016-11-02
[Feature Request] Separate Connections for Voyager and main tables


vyuldashev:
It would be nice if we could specify different connections to Voyager and main tables. By "main" I mean tables that are in main project. I prefer using admin panel as a separate project. So we could specify another connection in `database.php`.

adriaanzon:
This would also make it easy to add a prefix to the table names (#64)

vyuldashev:
@adriaanzon

Yeah, but may be I want Voyager be in one database and my main project in another. Even separate MySQL instances.

afsardo:
But Voyager is an admin panel to manage your application if you have separate databases are you managing a different app? It needs access to your app data.

vyuldashev:
@afsardo No I manage my own application. And I use micro-service architecture. So API, Web and other services are separated.

afsardo:
So do you need Voyager for what? :X Sorry having a hard time understanding it.

vyuldashev:
@afsardo Administration Interface for current database

afsardo:
Hum what you mean is using Voyager with access to your app database, but all things db related voyager needs to work are stored in a diff db? I see..

I think that is a "whole" new product, one i am actually working on, generating a Backend for API.

Let's see the Maintainers opinion on it.

vyuldashev:
@afsardo I think it would not be so hard. I can spend time and create pull request for this. It's just separating the entities. Voyager looks like Blog system with Admin interface right now, but not like Admin Interface.

afsardo:
Hum maybe you are right ;) Hope you do it! That would be an awesome addition and really powerful.

Any help or "lack of ideas" shoot me a message we can try and get a solution.

vyuldashev:
@afsardo Already started to work on. I also suggest to use PSR-2.

afsardo:
Completely agree with you ;) In Sublime i have an auto format part of the build process, need to install it in Visual Code too.

vyuldashev:
@afsardo what do you think about this:

```
 // This is for Voyager
 'mysql' => [
     'driver' => 'mysql',
     'host' => env('DB_HOST', 'localhost'),
     'port' => env('DB_PORT', '3306'),
     'database' => env('DB_DATABASE', 'forge'),
     'username' => env('DB_USERNAME', 'forge'),
     'password' => env('DB_PASSWORD', ''),
     'charset' => 'utf8',
     'collation' => 'utf8_unicode_ci',
[...]

PR #2175, opened 2017-11-06
Separate database connections for voyager and other tables to BREAD

fbuccioni:
Hi, I'm trying to use Voyager as admin console of a database and I saw the issue #69 trying to use the same feature.

I made a different base class of `Model` and `User` that gets the `$connection` property from `voyager.database.connection` config.

Steps to use it:

1. Configure both db, voyager and main in `config/database.php`
2. Set voyager db connection as default in `config/database.php`
3. Install voyager `artisan voyager:install` 
4. Set `voyager.database.connection` variable as voyager db connection in `config/voyager.php`
5. Set main db connection as default in `config/database.php`

### DIFF
--- a/src/Models/User.php
+++ b/src/Models/User.php
@@ -3,7 +3,7 @@
 namespace TCG\Voyager\Models;
 
 use Carbon\Carbon;
-use Illuminate\Foundation\Auth\User as Authenticatable;
+use TCG\Voyager\Auth\User as Authenticatable;
 use TCG\Voyager\Contracts\User as UserContract;
 use TCG\Voyager\Traits\VoyagerUser;
 
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -270,7 +270,15 @@ class VoyagerServiceProvider extends ServiceProvider
         // otherwise it will throw an error because no database
         // connection has been made yet.
         try {
-            if (Schema::hasTable('data_types')) {
+            $voyagerDBConn = config('voyager.database.connection');
+
+            if (empty($voyagerDBConn)) {
+                $hasDataTypeTable = Schema::hasTable('data_types');
+            } else {
+                $hasDataTypeTable = Schema::connection($voyagerDBConn)->hasTable('data_types');
+            }
+
+            if ($hasDataTypeTable) {
                 $dataType = VoyagerFacade::model('DataType');
                 $dataTypes = $dataType->get();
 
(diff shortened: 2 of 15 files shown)


src/Traits/Authenticatable.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace TCG\Voyager\Traits;
+
+use Illuminate\Foundation\Auth\User;
+
+class Authenticatable extends User
```
fletch3555: This is a class, not a trait, and shouldn't be a trait.  Please change the namespace.

src/VoyagerServiceProvider.php
```diff
@@ -270,22 +269,20 @@ public function registerGates()
         // otherwise it will throw an error because no database
         // connection has been made yet.
         try {
-            if (Schema::hasTable('data_types')) {
```
fletch3555, line 273: What's the reasoning behind removing this check?

src/Models/User.php
```diff
@@ -3,8 +3,8 @@
 namespace TCG\Voyager\Models;
 
 use Carbon\Carbon;
-use Illuminate\Foundation\Auth\User as Authenticatable;
 use TCG\Voyager\Contracts\User as UserContract;
+use TCG\Voyager\Traits\Authenticatable;
```
fletch3555: Same comment, Authenticatable is a class, not a trait.

fletch3555 (changes requested):
I don't really agree with these changes.  Models will already use the configured "default" connection.  If you need this to be different, just change `$this->connection` on your model class.

Please explain your use-case a bit more and I may reconsider

src/VoyagerServiceProvider.php
fbuccioni: Just in the debug stage I remove that if, now are ok

src/Traits/Authenticatable.php
fbuccioni: Done

src/Models/User.php
fbuccioni: Done

falcacibar:
My use case is because I have a db with several tables in another location and I'd like to not use that database to store the voyager tables, keeping it organized.

I need to make a mantainer and admin interface for that database so I use voyager because it fits right in the profile I want to use.

When I use separated connection in models I cannot use the database tool and I don't have the tool to create or edit new BREADs.

fbuccioni:
My use case is because I have a db with several tables in another location and I'd like to not use that database to store the voyager tables, keeping it organized.

I need to make a mantainer and admin interface for that database so I use voyager because it fits right in the profile I want to use.

When I use separated connection in models I cannot use the database tool and I don't have the tool to create or edit new BREADs.

PR #2186, opened 2017-11-07
Fix language picker

emptynick:
This will fix the language picker color being the same.
The active color is a 15% darkened version of the original color.

You will need to run
`php artisan vendor:publish --tag=voyager_assets --force`


fletch3555:
I'm inclined to push this to v1.1 as well due to the need to republish assets.  Bugfix changes shouldn't require the user to do anything as part of the upgrade.  Thoughts?

fletch3555:
I'll make the exception this time since it's so minor... Just need to resolve the merge conflict (pull it down, recompile assets, commit the result) @emptynick

fletch3555:
Same thing here @emptynick, you'll need to fix the merge conflict.

PR #2187, opened 2017-11-07
Translate Checkbox BREAD

cesaramirez:
Translate checkboxs

### DIFF
--- a/resources/views/tools/database/edit-add-bread.blade.php
+++ b/resources/views/tools/database/edit-add-bread.blade.php
@@ -153,7 +153,7 @@
 
                     <div class="panel panel-primary panel-bordered">
                         <div class="panel-heading">
-                            <h3 class="panel-title panel-icon"><i class="voyager-window-list"></i> {{ __('voyager.database.edit_rows', ['table' => $table]) }}</h3>
+                            <h3 class="panel-title panel-icon"><i class="voyager-window-list"></i> {{ __('voyager.database.edit_rows', ['table' => $table]) }}:</h3>
                             <div class="panel-actions">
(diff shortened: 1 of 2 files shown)


resources/views/tools/database/edit-add-bread.blade.php
```diff
@@ -153,7 +153,7 @@
 
                     <div class="panel panel-primary panel-bordered">
                         <div class="panel-heading">
-                            <h3 class="panel-title panel-icon"><i class="voyager-window-list"></i> {{ __('voyager.database.edit_rows', ['table' => $table]) }}:</h3>
+                            <h3 class="panel-title panel-icon"><i class="voyager-window-list"></i> {{ __('voyager.database.edit_rows', ['table' => $table]) }}</h3>
```
fletch3555: I really don't care much either way, but for my own curiosity, any particular reason for removing the colon?

heycesar: Hello, in the translation in Spanish it is duplicated.
But checking the other translations only happens, it would be better to change it only in the translation.
screen shot 2017-11-08 at 15 24 10

fletch3555: Agreed, I'd remove it from the translation(s) and add it back on this line.

heycesar: Change done!

fletch3555:
Perfect, thank you.  One note, please try to avoid mixing concerns in your PRs.  If you need to update translations, then just update the translations.  Don't also change backend code, views, etc.

I'm okay with it in this case, since removing the colon a very innocuous change.  Just try to avoid doing so in future PRs.

Thanks!

PR #2190, opened 2017-11-08
 Use custom attributes to improve default validation messages. Closes #2179

### LINKED ISSUE #2179: Validation fails when display name and field name of a BREAD field don't match
- Laravel Version: 5.5.*
- Voyager Version: 1.0.7
- PHP Version: 7.1.*
- Database Driver & Version: N/A

### Description:

On editing a BREAD resource, when using different field display name and field name, validation rules don't apply and get errors on fields where the conditions are met.

### Steps To Reproduce:

1. Fresh Voyager Install
2. Create a minimum table using Database Administration, with 'id' and 'sometext' fields.
3. Add BREAD to the resource
4. On the display label for 'sometext' field, set something different from 'sometext', like 'texto' and save
5. Create a new resource and set a value for the field 'sometext' labelled 'texto'
6. Press submit. No error will be shown, and there will be javascript errors because can't find the field with selector "[name=texto]" (which actually doesn't exist, as the name attribute for the field is 'somefield')

### Related issues: 

#1963 Validation scroll to first error not work

gausam:
This PR reverses changes made in [PR 2077](https://github.com/the-control-group/voyager/pull/2077), and uses [a simpler approach](https://github.com/the-control-group/voyager/issues/2179#issuecomment-342701108) to achieve the same effect.

Caveat is it branches off the branch of another [PR of mine](https://github.com/the-control-group/voyager/pull/2035), which Mark & Eric were yet to review. I was trying out the fix on my other PR's branch thinking it wouldn't work right off the bat (seemed to simple), and it did.


src/Http/Controllers/Controller.php
```diff
@@ -116,32 +116,44 @@ public function insertUpdateData($request, $slug, $rows, $data)
...
 
-    public function validateBread($request, $data)
+    public function validateBread($request, $data, $action = null)
     {
         $rules = [];
         $messages = [];
+        $customAttributes = [];
 
         foreach ($data as $row) {
+            if ($row->display_name) $customAttributes[$row->field] = $row->display_name;
```
blitux, line 126: Are we enforcing a coding standard? The way the if is expressed is not allowed on PSR-2. It MUST have the brackets. But that if we're enforcing that standard. StyleCI is complaining about the code style.

src/Http/Controllers/Controller.php
```diff
@@ -116,32 +116,44 @@ public function insertUpdateData($request, $slug, $rows, $data)
...
+
             $options = json_decode($row->details);
 
             if (isset($options->validation)) {
-                if (isset($options->validation->rule)) {
-                    if (!is_array($options->validation->rule)) {
-                        $rules[$row->display_name] = explode('|', $options->validation->rule);
+                $validationOptions = $options->validation;
+
+                // if store or update-specific validation is defined, use that instead
```
blitux, line 133: On the docs there are no mention of different rules per action (update or store). I think if this PR tries to fix an issue, it should do exactly that. A feature should be done in a different PR.

src/Http/Controllers/Controller.php
```diff
@@ -116,32 +116,44 @@ public function insertUpdateData($request, $slug, $rows, $data)
...
+                if ($action === 'store' && isset($options->validation->store)) {
+                    $validationOptions = $options->validation->store;
+                } elseif ($action === 'update' && isset($options->validation->update)) {
+                    $validationOptions = $options->validation->update;
+                }
+
+                if (isset($validationOptions->rule)) {
```
blitux, line 140: Try to avoid nested IFs. If you're inside a loop, you can call "continue" to jump to the next loop. This is for better readability and simplicity in the code.

src/Http/Controllers/VoyagerBreadController.php
```diff
@@ -225,7 +225,7 @@ public function update(Request $request, $id)
         $this->authorize('edit', $data);
 
         // Validate fields with ajax
-        $val = $this->validateBread($request->all(), $dataType->editRows);
+        $val = $this->validateBread($request->all(), $dataType->editRows, __FUNCTION__);
```
blitux, line 228: Same as the comment on the validateBread method. The validation per action is not stated on the docs. A PR of a bugfix should not contain new features.

blitux (commented):
Hi to all. I just give my 2 cents here. Take the comments as a recomendation.

src/Http/Controllers/Controller.php
gausam: Don't worry, StyleCI sends in PRs auto for those :)

gausam:
@blitux Defn agree, I said the same thing on Slack. Like I said, this was a quick doodle (on the wrong branch) that worked out well.

As far as code style changes to Voyager's code, the team welcomes those. In my PRs I just conform to whatever code layout approach is in place at that point to reduce friction and unintended side effects which is what got us to this issue in the first place. I prefer a refactor to be its own PR, unless of course the refactor is what makes the PR relevant. So if you see a part that you want to re-do, then by all means do, with a pinky promise you're sure we won't wind up here again :smile: 

I see you pushed a PR. Let me close this one in favour of yours, and have changes we need on your side of things. Let's take the discussion there.

gausam:
Closing in favour of https://github.com/the-control-group/voyager/pull/2191

src/Http/Controllers/Controller.php
fletch3555: @samtheson is correct.  We like having StyleCI be happy when the PR is opened, but if it's not, StyleCI will automatically open a PR to fix it once the PR gets merged.

PR #2196, opened 2017-11-08
Improved datepicker selection

JapSeyz:
Changed datepicker to: https://nehakadam.github.io/DateTimePicker/

It's responsive, and allows for user-input. Which the current one doesn't as far as I can see, and that's a hassle since the current one defaults to 1970, and going from 1970 to preset time takes a while, when you can only go forwards one year at a time. I may be wrong here though, I just couldn't get it to go forwards more than a year at a time.

It's also quite stylish.


emptynick:
But you dont make use of it in the form fields right?
The documentation says you need to use it like
`<input type="text" data-field="datetime" readonly>`
Voyagers date-input looks like
`<input type="date" data-name="X"  name="X">`

resources/assets/js/app.js
```diff
@@ -29,6 +29,7 @@ require('./load-remote.js');
 require('cropperjs');
 
 $(document).ready(function(){
+    $("#dtBox").DateTimePicker();
```
fletch3555, line 32: Probably doesn't matter, but why change the position of the initialization call? Also, I'm not familiar with the package you're suggesting.  I'm guessing it uses a single element to hold the modal/popup instead of the individual fields, which is why you added the <div> below and are using `#dtBox` instead of `.datePicker`

JapSeyz: Correct, it uses a single element to hold the element.

It looks like this: 
screenshot from 2017-11-09 02 29 45


The new location is an oversight, as I just needed to make sure it was in document.ready for my tests, and forgot to move it once I had concluded my tests.

emptynick:
Actually you can use the current date, just click the circle in the upper right corner.
To jump through years, just click `December 2017` and scroll down.
Theres an infinite selection.

datepicker

PR #2198, opened 2017-11-09
Use custom attributes INSTEAD to improve default validation messages

gausam:
Like @ratt pointed out in issue #2179, the mechanism in PR #2077 needs some reworking, which this PR does.

It uses `customAttributes` to achieve the same effect. I also reversed changes made to `blade` files and `app.js` in PR #2077.


emptynick:
I did not yet start doing QA, but I see you missed pushing the `resources/assets/sass/app.scss`

PR #2223, opened 2017-11-14
Fix permission check for adding an item

ratt:
Currently, the permission checks for edit when it should be checking for add for a new item


fletch3555:
Thanks!

PR #2229, opened 2017-11-15
add scrollwheel

mtk3d:


fletch3555:
so... which is it? #2228 or #2229 ?

mtk3d:
Both of them, but you can remove this pull requests, and then i will make it better in single commit.

fletch3555:
Okay

PR #2232, opened 2017-11-15
fix select_multiple

mbdwey:
Hello,
I know that select_multiple without relation are not official option but it's available in the codebase 
please check this fix
select_multiple need array on browse but this template try to decode is a like if it's json so this change unify browse and edit,create


fletch3555:
If you're going to remove code for a PR, please just remove it, don't leave it commented out.

With that said, that particular line was added by a recent PR to fix other issues.  I'm not saying it was the right solution (there still appear to be issues), but it did fix the issue it was intended for.  Simply removing that line is not the correct solution here.

mbdwey:
I find that other places should be checked , I will resubmit another pull request

fletch3555:
You don't need to submit a new one.  Additional commits to this branch will automatically show up here.

fletch3555:
Closing due to no response.  We can reopen this once the requested changes have been addressed

Issue #2225, opened 2017-11-14
default avatar is gone


caveoIT:
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version:7.1
- Database Driver & Version:PostgreSQL 9.3.5

### Description:
Thanks very much for your efforts, and sorry if there is too much load on you to fix all that bugs, But
when i delete any user with default avatar from Voyager UI, Default avatar deleted too, So every time i must upload the default avatar again.

### Steps To Reproduce:
1- Create new user.
2- Upload a new avatar and remove the old one.
3- Change the new avatar name to default.
4- Go to users page and delete any user with default avatar.
5- Go to the media section -> users

francescomugnai:
Same problem here

UPDATE: I found the problem, I hadn’t changed the APP_URL on the .env file

dfk7677:
I confirm that if a user with the default avatar is deleted, the image _users/default.png_ is deleted. It seems there is no check in the `cleanup` function of the `VoyagerBreadController`class for shared images (and files in general).

dfk7677:
I am not so familiar with the code to know if other files can be shared, but if this is not the case, a simple check in `deleteBreadImages` function for _users/default.png_, which should never be deleted, would solve this issue.

Instead of (line 426 in VoyagerBreadController.php):
```php
$this->deleteFileIfExists($data->{$row->field});
```
Insert check:
```php
if($data->{$row->field} != "users/default.png") {
    $this->deleteFileIfExists($data->{$row->field});
}
```

fletch3555:
I'm okay with that as a stop-gap solution.  Other files can be shared, but not through the file/image upload fields, which is how they're being deleted, so this should be good.  We'll be doing a rewrite in a future release for how Voyager manages media (files and images) that get uploaded to it, so keep an eye open for that.

Would you like to open the PR for that change, @dfk7677?  If not, we can wait for someone else to do so (I don't have the time at the moment, but may later tonight)

emptynick:
This was fixed in #2233

omer-jan:
@millerone  thanks i changed the   APP_URL on the .env file and it work for me

PR #2233, opened 2017-11-15
Fix for issue #2225

dfk7677:
Added a check for file _users/default.png_ (default avatar image) when deleting image files, as this file is not supposed to be deleted.

This is a fix for [#2225](https://github.com/the-control-group/voyager/issues/2225).


fletch3555:
Please use feature branches when committing changes to open a PR.  You've now introduced additional changes into this PR (bc2b179).  Please revert these changes as they're unrelated and I can't accept the PR with them

dfk7677:
I am sorry about this. Changes are reverted now.

fletch3555:
Very good, thank you.

emptynick:
I think the filename should not be hardcoded, as it is a config var.
Just use:
`if ($data->{$row->field} != config('voyager.user.default_avatar')) {`

src/Http/Controllers/VoyagerBreadController.php
```diff
@@ -423,7 +423,9 @@ protected function cleanup($dataType, $data)
     public function deleteBreadImages($data, $rows)
     {
         foreach ($rows as $row) {
-            $this->deleteFileIfExists($data->{$row->field});
+            if ($data->{$row->field} != 'users/default.png') {
```
fletch3555: As @emptynick suggested, `'users/default.png'` should not be hardcoded, and should be replaced with the config var.

fletch3555 (changes requested):

dfk7677:
Thanks @emptynick . Change done @fletch3555 .

fletch3555 (approved):

fletch3555:
Perfect, thanks! @emptynick, have you given this a look from a QA perspective?  If all good, I can merge it

emptynick:
Yes I did, its working fine.
Just trying to think about any cases where this would get a prolem - but I cant think of any.

fletch3555:
Nor can I, thanks!

PR #2244, opened 2017-11-16
Fixing voyager tests

tnylea:
Not sure why this was working before but I believe this is going to fix the tests now 👍 
@fletch3555 let me know if you can merge this in if it passes so that way we can get the green `passing` icon back :) Thanks.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -26,7 +26,7 @@
     <php>
         <env name="APP_ENV" value="testing"/>
         <env name="APP_DEBUG" value="true"/>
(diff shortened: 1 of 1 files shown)


fletch3555:
I also have no idea how that was working before, since APP_URL is supposed to be the protocol + FQDN + root path... The default value from laravel is `http://localhost`, which is probably what should be used here, not `/`.

Mind making that one change? Then I'll get it merged in right away! Thanks!

tnylea:
@fletch3555 Good call, just changed that. Let me know if you can get it merged in 👍 
Appreciate it.

PR #2245, opened 2017-11-16
added rtl css & modified master login blade for support

farzinft:


src/Helpers/helpers.php
```diff
@@ -20,3 +20,10 @@ function voyager_asset($path, $secure = null)
         return asset(config('voyager.assets_path').'/'.$path, $secure);
     }
 }
+
+if (!function_exists('isRtl')) {
+    function isRtl()
+    {
+        return app()->getLocale() == 'fa' || app()->getLocale() == 'ar';
```
fletch3555: I'm not sure I like having the language codes hard-coded here.  There are more languages than just Farsi and Arabic that are rtl. (https://meta.wikimedia.org/wiki/Template:List_of_language_names_ordered_by_code)

Can we move this to a config setting or something? I'm open to suggestions on this.

publishable/assets/css/app.rtl.css
```diff
@@ -0,0 +1,29 @@
+@charset "UTF-8";/*!
```
fletch3555, line 1: How is this file generated? Or did you just copy app.css and rename it?  If the latter, you can't do that since it won't get any future changes

fletch3555 (changes requested):

src/Helpers/helpers.php
farzinft: you're right , i'll do it

publishable/assets/css/app.rtl.css
farzinft: i used a npm package named rtlcss 
https://www.npmjs.com/package/rtlcss

fletch3555: Right, but that's a one-time rewrite, right? I don't see you adding it as a dependency anywhere.

We'll need to make sure that file gets generated from mix before I can accept this.

src/Helpers/helpers.php
fletch3555: Perfect, thanks!

publishable/assets/css/app.rtl.css
farzinft: it's a npm package and installed globally  so i cant use as dependency , so your suggestion is  changing on available app.scss file ?

fletch3555: I'm saying that this needs to be automated in some way, so that the rtl file can be generated any time someone updates the normal file (via compiling the raw scss)

farzinft: right, ok i'll do it , thanks!

hamog:
Thank you. It is great.

fletch3555:
I cannot accept this until the app.rtl.css file is generated in an automated way.

FaridAghili:
npm install -g rtlcss
rtlcss app.css

And app.rtl.css is generated automatically

fletch3555:
We're open to PRs to make this generate automatically.

PR #2290, opened 2017-11-29
Feature User Multi Roles

### LINKED ISSUE #371: [Feature Request] User Multi Roles
In some circumstances multi roles per user are really missing.

### LINKED ISSUE #1150: Roles: Editor cannot edit his own profile?
- Laravel Version: 5.4.19
- Voyager Version: 0.11
- PHP Version: 7.1.3
- Database Driver & Version: Mysql

Description:

I think this is not really an issue, but something to be aware of.

I created an editor role that does not have access to edit user profile.

However, I think it should be allowed to edit its own profile.

If you have time, please check on this.

Best Regards

### LINKED ISSUE #2192: Warning in Roles: the user can become an admin
- Laravel Version: 5.5.20
- Voyager Version: 1.0
- PHP Version: 7.1.7
- Database Driver & Version: MySQL 5.7x64

### Description:
If the user **is not 'admin' and do not have access** to 'roles', but he can 'browse_admin', then **he can change his role to 'admin'** from 'Profile'

2017-11-08-role-1v
2017-11-08-role-4v

### Steps To Reproduce:
 Dashboard -> navbar -> dropdown -> profile -> Edit my profile -> role

akazorg:
Here follows my propose for User Multi Roles.

Note: A user, cannot, ever, change his own Role.

I have kept the field `users.role_id`, and it's used for the default User Role. We can't remove this field, or it would break all existing Voyager installations. 

The ideal solution is using a single table `users_roles`, my suggestion is to update this after we revise Voyager with the upgrade system, that will allow us to upgrade existing installations without any issues.

**Users Multi Roles / Permissions DB**
image

Please test it and send feedback.

Closes #371
Closes #1150
Closes #1726
Closes #2192

### DIFF
--- a/publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
+++ b/publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
@@ -14,7 +14,7 @@ class AddUserRoleRelationship extends Migration
     public function up()
     {
         Schema::table('users', function (Blueprint $table) {
-            $table->unsignedInteger('role_id')->nullable(false)->change();
+            $table->unsignedInteger('role_id')->change();
             $table->foreign('role_id')->references('id')->on('roles');
         });
     }
@@ -27,6 +27,7 @@ class AddUserRoleRelationship extends Migration
     public function down()
     {
         Schema::table('users', function (Blueprint $table) {
+            $table->integer('role_id')->change();
             $table->dropForeign(['role_id']);
         });
     }
--- a/publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
+++ b/publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddUserRoleRelationship extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->unsignedInteger('role_id')->nullable(false)->change();
+            $table->foreign('role_id')->references('id')->on('roles');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropForeign(['role_id']);
+        });
+    }
+}
--- a/publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
+++ b/publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
@@ -1,0 +1,39 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class DropPermissionGroupsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('permissions', function (Blueprint $table) {
+            $table->dropColumn('permission_group_id');
+        });
+
+        Schema::dropIfExists('permission_groups');
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('permissions', function (Blueprint $table) {
+            $table->integer('permission_group_id')->unsigned()->nullable()->default(null);
+        });
+
+        Schema::create('permission_groups', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('name')->unique();
+        });
+    }
+}
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -9,19 +9,14 @@ class Role extends Model
 {
     protected $guarded = [];
 
-    public function usersDefault()
-    {
-        return $this->hasMany(Voyager::modelClass('User'))->select('*');
-    }
-
-    public function usersAlternative()
-    {
-        return $this->belongsToMany(Voyager::modelClass('User'), 'user_roles')->select('users.*');
-    }
-
     public function users()
     {
-        return $this->usersDefault()->union($this->usersAlternative()->toBase());
+        return $this->hasMany(Voyager::modelClass('User'))->select('*')->getQuery()
+                    ->union(
+                        $this->belongsToMany(Voyager::modelClass('User'), 'user_roles')
+                             ->select(Voyager::model('User')->getTable().'.*')
+                             ->getQuery()
+                    );
     }
 
     public function permissions()
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -9,10 +9,19 @@ class Role extends Model
 {
     protected $guarded = [];
 
+    public function usersDefault()
+    {
+        return $this->hasMany(Voyager::modelClass('User'))->select('*');
+    }
+
+    public function usersAlternative()
+    {
+        return $this->belongsToMany(Voyager::modelClass('User'), 'user_roles')->select('users.*');
+    }
+
     public function users()
     {
-        return $this->hasMany(Voyager::modelClass('User'))->get()
-                    ->merge($this->belongsToMany(Voyager::modelClass('User'), 'user_roles')->get());
+        return $this->usersDefault()->union($this->usersAlternative()->toBase());
     }
 
     public function permissions()
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -11,12 +11,11 @@ class Role extends Model
 
     public function users()
     {
-        return $this->hasMany(Voyager::modelClass('User'))->select('*')->getQuery()
-                    ->union(
-                        $this->belongsToMany(Voyager::modelClass('User'), 'user_roles')
-                             ->select(Voyager::model('User')->getTable().'.*')
-                             ->getQuery()
-                    );
+        $userModel = Voyager::modelClass('User');
+
+        return $this->belongsToMany($userModel, 'user_roles')
+                    ->select(app($userModel)->getTable().'.*')
+                    ->union($this->hasMany($userModel))->getQuery();
     }
 
     public function permissions()
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -11,7 +11,8 @@ class Role extends Model
 
     public function users()
     {
-        return $this->belongsToMany(Voyager::modelClass('User'), 'user_roles');
+        return $this->hasMany(Voyager::modelClass('User'))->get()
+                    ->merge($this->belongsToMany(Voyager::modelClass('User'), 'user_roles')->get());
     }
 
     public function permissions()
--- a/src/Traits/VoyagerUser.php
+++ b/src/Traits/VoyagerUser.php
@@ -60,7 +60,7 @@ trait VoyagerUser
     }
 
     /**
-     * Set default User Role
+     * Set default User Role.
      *
      * @param string $name The role name to associate.
      */
--- a/src/Traits/VoyagerUser.php
+++ b/src/Traits/VoyagerUser.php
@@ -11,8 +11,6 @@ use TCG\Voyager\Models\Role;
  */
 trait VoyagerUser
 {
-    protected $cache_permissions = [];
-
     /**
      * Return default User Role.
      */
@@ -78,9 +76,13 @@ trait VoyagerUser
 
     public function hasPermission($name)
     {
-        $this->loadPermissionsCache();
+        $this->loadPermissionsRelations();
 
-        return in_array($name, $this->cache_permissions);
+        $_permissions = $this->roles_all()
+                              ->pluck('permissions')->flatten()
+                              ->pluck('key')->unique()->toArray();
+
+        return in_array($name, $_permissions);
     }
 
     public function hasPermissionOrFail($name)
@@ -122,17 +124,6 @@ trait VoyagerUser
 
         if (!$this->relationLoaded('roles.permissions')) {
             $this->load('roles.permissions');
-        }
-    }
-
-    private function loadPermissionsCache()
-    {
-        $this->loadPermissionsRelations();
-
-        if (empty($this->cache_permissions)) {
-            $this->cache_permissions = $this->roles_all()
-                                            ->pluck('permissions')->flatten()
-                                            ->pluck('key')->unique()->toArray();
         }
     }
 }
--- a/src/Traits/VoyagerUser.php
+++ b/src/Traits/VoyagerUser.php
@@ -11,27 +11,59 @@ use TCG\Voyager\Models\Role;
  */
 trait VoyagerUser
 {
+    protected $cache_permissions = [];
+
+    /**
+     * Return default User Role.
+     */
     public function role()
     {
         return $this->belongsTo(Voyager::modelClass('Role'));
     }
 
     /**
+     * Return alternative User Roles.
+     */
+    public function roles()
+    {
+        return $this->belongsToMany(Voyager::modelClass('Role'), 'user_roles');
+    }
+
+    /**
+     * Return all User Roles, merging the default and alternative roles.
+     */
+    public function roles_all()
+    {
+        $this->loadRolesRelations();
+
+        return collect([$this->role])->merge($this->roles);
+    }
+
+    /**
      * Check if User has a Role(s) associated.
      *
-     * @param string|array $name The role to check.
+     * @param string|array $name The role(s) to check.
      *
      * @return bool
      */
     public function hasRole($name)
     {
-        if (!$this->relationLoaded('role')) {
-            $this->load('role');
+        $roles = $this->roles_all()->pluck('name')->toArray();
+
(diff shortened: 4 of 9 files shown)


vFire:
AWESOME!

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->unsignedInteger('role_id')->nullable(false)->change();
+            $table->foreign('role_id')->references('id')->on('roles');
```
fletch3555: We didn't already have a foreign key here...? huh...

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->unsignedInteger('role_id')->nullable(false)->change();
```
fletch3555: Do we set a default value anywhere? It's possible that someone could have users currently with a null value for `role_id`, which would cause this migration to break.

publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('permissions', function (Blueprint $table) {
+            $table->dropColumn('permission_group_id');
```
fletch3555, line 17: Were we using this field/table at all? I don't remember it getting implemented.  If not, is this change related to the primary purpose of this PR, or is it an unrelated change? If the latter, it should be moved to it's own PR

src/Models/Role.php
```diff
@@ -11,7 +11,8 @@ class Role extends Model
 
     public function users()
     {
-        return $this->belongsToMany(Voyager::modelClass('User'), 'user_roles');
+        return $this->hasMany(Voyager::modelClass('User'))->get()
```
fletch3555: I understand why you did it, but this method no longer returns a `Builder` object, so it can't be filtered.  This may not be a problem in this case, but it goes against laravel convention.  i.e. you won't be able to call `$role->users()->where(.....)->get()`

src/Traits/VoyagerUser.php
```diff
@@ -46,15 +78,9 @@ public function setRole($name)
 
     public function hasPermission($name)
     {
-        if (!$this->relationLoaded('role')) {
-            $this->load('role');
-        }
+        $this->loadPermissionsCache();
```
fletch3555: While I completely understand the reason behind this, is it absolutely necessary for this PR? Can it be implemented on it's own then modified for this PR? "Caching Permissions" seems 100% unrelated to "Multiple Roles"

fletch3555 (changes requested):

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
akazorg: I see, I had that doubt too. We should skip this, then?

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
akazorg: We did not, but if we allow nullables, we should remove this foreign key?

publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
akazorg: I think Permissions group was added for handling User Multi Roles, but was never implemented.
With this PR, I don't see the use of Group permissions, so I removed it.

src/Models/Role.php
akazorg: Sorry, you are absolutely right, that ->get() should be removed.

src/Traits/VoyagerUser.php
akazorg: You're right, I also dislike that approach. Maybe we can skip this cache variable.
Will check that.

Thanks @fletch3555 for your feedback! :)

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
fletch3555: Yes, I would remove the change to remove nullability on this field.

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
fletch3555: MySQL (at least) allows null values in a foreign key field (so long as it's not a UNIQUE index): https://stackoverflow.com/questions/7573590/can-a-foreign-key-be-null-and-or-duplicate.

I think the foreign key is fine

publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
fletch3555: Fair enough.  Works for me.

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
akazorg: fixed

publishable/database/migrations/2017_11_26_013050_add_user_role_relationship.php
akazorg: fixed

src/Models/Role.php
akazorg: @fletch3555 I have no idea how to merge the 2 relations and keep the `Builder` object, is that even possible? Please, help!!! :D

If this is impossible, we can use it this way, and fix it when Voyager Upgrade System is released?

fletch3555: That's why I started with "I understand why..." haha

It's has to be possible (I'm thinking unions).  Try this: https://laracasts.com/discuss/channels/eloquent/combining-multiple-hasmany-relationships-into-one-model-relation-query-builder

akazorg: @fletch3555 done, with unions, please take a look.
I had to add `->select('*')` and `->select('users.*')`, but we need to improve this... any idea?
for instance, we can't refer to `users` table name directly.

nickbryan:
Awesome, looking forward to this one!

src/Models/Role.php
fletch3555: I would've just put both `belongsTo...()` calls inline connected with the `union()` instead of other methods.  Also, instead of `->select('*')`, you should be able to do `->getQuery()`([ref](https://laravel.com/api/5.5/Illuminate/Database/Eloquent/Relations/BelongsToMany.html#method_getQuery)), which will return the `Builder` object you need for `->union()`

In short, I picture it something along the lines of this:
```
public function users()
{
return $this->hasMany(Voyager::modelClass('User'))->getQuery()->union($this->belongsToMany(Voyager::modelClass('User'), 'user_roles')->getQuery());
}
```

fletch3555 (approved):

fletch3555:
Needs a bit of QA/testing love, but code looks good to me.  Only question is whether there are any breaking changes that should push this back to v1.1 instead of v1.0.8

nickbryan:
Im pretty new to this project but from looking at the code I can't see any breaking changes introduced. I can check it out over the weekend and do some QA/testing if it helps?

akazorg:
Thanks @nickbryan, that would be awesome.

nickbryan:
Hi @akazorg ,

I am just looking at this now. I have built with the php artisan voyager:install --with-dummy command for the dummy data.

Couple things I have noticed so far:

It would seem the translations are missing in the publishable directory so the default and additional roles are appearing as voyager.profile.role_default and voyager.generic.roles_additional. Not sure if you guys get translations done after the PR is merged or if it should be a part of this? Also, should they be under the same translation namespace ie voyager.profle or voyager.generic?

screen shot 2017-1 2-01 at 13 45 27

Another thing is that when I go to edit the admin@admin.com user I cannot change the roles for that user, is this because its the my user profile (who i am logged in as)? It works for the test user that I created (seen above) but not for the admin user that I am logged in as. Is this behaviour intended?

fletch3555:
Translations should be included with the PR, not done after.

Also, we should probably identify "admin" users (by the admin role maybe...?) And allow them to bypass the restriction.  At the same time, it's fairly commonplace for users to not be able to downgrade their own role to prevent getting locked out of the system entirely.  Thoughts @akazorg?

nickbryan:
Yeah, I would agree with the users not being able to edit there own roles but for administration I feel it is needed. This could just be defaulted for the admin role but maybe we could add an option somewhere that allows a user to edit there own roles, maybe that could be on the user table and set to true for the administrator by default? I presume that would need to be reflected when someone does not use the dummy data and adds a user through the artisan command? If an admin user is added in this way they would not be able to change roles for themselves.

akazorg:
Hello! Nice catch, I missed moving translations into the Voyager folder xD Will do that.

I think a User cannot change his role, or append additional roles to their account, **ever**, or we would be opening a security issue here. 
Users with permission `edit_users` are able to change other users roles. I believe this is the cleanest way of doing it. 
What you think?

[edit]
If admins have all permissions, they will not require additional roles.

akazorg:
I've updated the missing translation file and view.

I was hoping to use this on my next project, but had to delay it :/ anyway, I think this is good for production, please let me know if you want to change something.

I'll be quite busy till the end of next week.

Have fun guys :D

nickbryan:
From my testing everything seems to be good for production also. If we are settled on users not being able to change their own roles then I think its good to go. It does make sense to approach it from this angle with the security issues that may arise. Thanks for the work @akazorg the quicker we can get this into a release the better as we also want to use this in our next project in January.

fletch3555:
@akazorg, what happens if you don't run the migrations? Will this break?  I'm thinking of users running a simple upgrade without realizing they need to do that extra step.

emptynick:
I'm having problems with Select2 (I guess).
If you select user as the default role, you cant select admin again.
But I think this is a problem with Select2 z-index. The admin option is directly above the lower Select2.
Maybe you could put them next to each other as `col-md-6` for now until theres a proper fix.

akazorg:
@nickbryan thank you for testing and review, very useful!

@fletch3555 What if Users run the migrations? Voyager will crash immediately reporting the missing tables, so it might be easy to track.

Alternatively, what do you suggest for the upgrade script? 

While I love the idea of the `Voyager Upgrade System`, and personally have been doing some efforts for implement this feature, using it, would change this PR entirely, because it would make no more sense on using both relationships, since we are able to run the upgrade script and move `users.role_id` to the pivot table `users_roles.role_id`, along with removing the field `users.role_id` and relationship with `roles` table.

I'm not sure if @marktopper is working on the `Voyager Upgrade System`.

@emptynick I'm afraid you are facing another bug? Is not happening with me.
I left a harmless bug there, a User can have the same role defined twice (Default Role and Additional Roles).
Solution: when the Default Role changes, we should disable (and deselect) the Default Role at the <select> Additional Roles.

Thank you for your feedback!

emptynick:
Its definitely a different bug, and I'm looking into it.
e4485134f9

akazorg:
I see!!! Nice catch @emptynick, thanks for the image :) tricky bug!

I think we should handle this with the `z-index`, using `col-md-6` wont work on mobile because fields get stacked.

fletch3555:
My question was more whether we can really get this in as a non-breaking change.  It's not completely harmless, as you've admitted above, since it will break if a user hasn't run the migration.  The issue is that most users won't think to republish assets and rerun migrations after just a bugfix update.  I'm thinking more and more that this needs to go to `develop` for a 1.1 release.  Shouldn't be too far off if you're worried about timing

akazorg:
You're right @fletch3555, this should go to `dev`, it would break. 
Anyway, people will be able to use it trough `dev` branch.
Thanks for your work on this.

mdk999:
I'm trying to use this feature in a project but unable to get it working/installed. The comments above mentioned it would be in the 'dev' branch but when I installed (via composer) 'dev-master' the changes were not there. I then downloaded from here 
https://github.com/the-control-group/voyager/tree/b7f2723a6d83c49bb2d2587909b73a74d4e81edc
and then ran artisan voyager:install which seemed to install but again none of the changes. I ran artisan migrate:refresh then got
SQLSTATE[HY000]: General error: 1832 Cannot change column 'role_id': used in a foreign key constraint 'us  
  ers_role_id_foreign' (SQL: ALTER TABLE users CHANGE role_id role_id INT DEFAULT NULL) 
I ran artisan migrate:reset and got
SQLSTATE[23000]: Integrity constraint violation: 1217 Cannot delete or update a parent row: a foreign key  
   constraint fails (SQL: drop table `roles`) 

Any help would be appreciated - thanks

akazorg:
@mdk999, try `artisan vendor:publish` to update migrations.

publishable/database/migrations/2017_11_26_014010_drop_permission_groups_table.php
whyayala: @akazorg @fletch3555 I could see another use case for `permission_group_id`. If permissions were grouped they could be mass assigned or removed from a role, while also being able to add and remove individual permissions. AWS works similarly, where you can mass assign permissions to a user, or select from groups of permissions. That was you can give a user full access to let's say S3, without having to pick each individual S3 permission. This is what I had imagined when i discovered that column had been added. Are you sure you want/need to remove this?

fletch3555: @whyayala, that's what roles are. Additional grouping for the sole purpose of assignment is just added complexity

whyayala: @fletch3555 I understand that, but if you have a lot of users, roles, and permissions it could become encumbering for an admin to manage permission to role assignments on a granular level. It's a minor convenience and I agree that it adds more complexity, but I felt obligated to mention it.

fletch3555: I completely understand and appreciate that obligation, but for now at least, I don't see the benefit outweighing the cost of complexity, so we're going to hold off on that.  If the argument gets made later, we'll consider adding it back as part of that implementation.  For now, it does nothing.

Thank you.

johnss:
I would suggest to add 1 more table to store grouping of roles,
the name would be role_attributes, role_types, role_groups , etc.

`An example of ABAC would be allowing only users who are type=employees and have department=HR to access the HR/Payroll system and only during business hours within the same timezone as the company.`

By that quote above we need `type=employee` & `departement=HR` to access HR/Payroll, if role attribute features added, it will make it very similar to ABAC, this probably will make Voyager a PHP script with the most complex permission control system.


To know more about ABAC
http://blog.identityautomation.com/rbac-vs-abac-access-control-models-iam-explained

fletch3555:
@johnss I think ABAC is much too complex for this project.  I seriously doubt (at this point at least), that anyone is trying to use Voyager for a site that is large-enough scale to require such fine-grained controls.  It would only make the smaller implementations that much more cumbersome to set up.  Perhaps we'll refactor this in a later version.  Maybe we'll refactor the auth system into something that allows Hooks to tie into it, allowing anyone to build whatever level of control they need.  In either case, it won't be going in now.  Thank you for the suggestion though.

akazorg:
Hi @johnss, for this example:
>An example of ABAC would be allowing only users who are type=employees and have department=HR to access the HR/Payroll system and only during business hours within the same timezone as the company.

I believe we can do the following?
- create `role` `employees` and assign to User
- create `permissions` for `access the HR/Payroll`
- assign `permissions` to role `employees`

(the logic of business hours, etc.. should be coded separately)

Creating Roles Groups would be too much, the system seems to works perfect this way.
Thanks for your feedback!

fletch3555:
I'm going to merge this in for 1.1.  @akazorg, would you mind working on a draft upgrade guide, at least for steps related to this change?

akazorg:
Sure, will do @fletch3555. 
IDK the progress on the `Upgrade Script`, but I guess this draft is always required.

fletch3555:
@akazorg.  At the very least, it should be in the docs.

johnss:
Sorry for late reply

@akazorg yes that would work, but it's still an RBAC, if you have so many roles (hundred / thousand), managing role will be painful and role grouping will be a much needed features, that's what ABAC is trying to solve.

And I don't quite agree with @fletch3555, ABAC is flexible, it can act as simple ACL it can check for permission without first assigning a role, while in RBAC a permission need to be assigned to make it work, in ABAC grouping permission into roles is optional and grouping of roles into role type/group is also optional

I agree with your other approach to implement ABAC, given the rare use cases and high complexity, probably its better to use hooks / events to modify listing of roles/permission and letting users implement their own ABAC.

jampack:
Can someone explain whats the purpose of "users" table relation with "roles" table where user gets foreign key role_id with a zero to many relationship.
I dont understand the purpose of this relation: 
capture

emptynick:
This is mentioned in the first post, just above the diagram.

tommiekn:
Hi,

How can i check what roles the user is assigned to ?

Auth::user()->relations['roles'][0]->attributes['name'] is the first one but can't i search for roles ?

Thanks

Issue #2287, opened 2017-11-28
Cover Image Upload Field Accepts Non-Image Filetypes


ux-engineer:
- Laravel Version: 5.5.21
- Voyager Version: 1.0.7
- PHP Version: 7.1.9

### Description:

Voyager default Pages datatype's image field (perhaps others) accepts PDF file as chosen by the file picker. But upload ends with an error page:

```Unsupported image type. GD driver is only able to decode JPG, PNG, GIF or WebP files.```

### Suggestion:

Image file input element should have [_accept_ attribute](https://stackoverflow.com/questions/3828554/how-to-allow-input-type-file-to-accept-only-image-files) defined for restricting the user to being able to choose only image files.

fletch3555:
Agreed.  We're open for pull requests

PR #2299, opened 2017-12-03
show only image files while uploading cover images

pseudoanime:
This should fix issue #2287

### DIFF
--- a/publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
+++ b/publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
@@ -13,7 +13,7 @@ class AddVoyagerUserFields extends Migration
             if (!Schema::hasColumn('users', 'avatar')) {
                 $table->string('avatar')->nullable()->after('email')->default('users/default.png');
             }
-            $table->integer('role_id')->nullable()->after('id');
+            $table->unsignedInteger('role_id')->nullable()->after('id');
         });
     }
 
--- a/publishable/database/migrations/2016_01_01_000000_create_pages_table.php
+++ b/publishable/database/migrations/2016_01_01_000000_create_pages_table.php
@@ -8,15 +8,13 @@ class CreatePagesTable extends Migration
 {
     /**
      * Run the migrations.
-     *
-     * @return void
      */
     public function up()
     {
         // Create table for storing roles
         Schema::create('pages', function (Blueprint $table) {
             $table->increments('id');
-            $table->integer('author_id');
+            $table->unsignedInteger('author_id');
             $table->string('title');
             $table->text('excerpt')->nullable();
             $table->text('body')->nullable();
@@ -31,8 +29,6 @@ class CreatePagesTable extends Migration
 
     /**
      * Reverse the migrations.
-     *
-     * @return void
      */
     public function down()
     {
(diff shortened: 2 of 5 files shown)


emptynick:
You didnt change the bread/edit-add.blade.php file and you implemented the image-type again.
Theres already an if for it.
This wont change the issue you linked.

publishable/database/migrations/2016_01_01_000000_add_voyager_user_fields.php
```diff
@@ -13,7 +13,7 @@ public function up()
             if (!Schema::hasColumn('users', 'avatar')) {
                 $table->string('avatar')->nullable()->after('email')->default('users/default.png');
             }
-            $table->integer('role_id')->nullable()->after('id');
+            $table->unsignedInteger('role_id')->nullable()->after('id');
```
fletch3555, line 16: We shouldn't be changing existing migrations (except for major version updates) because it won't run for users with existing installs.

publishable/database/migrations/2016_01_01_000000_create_pages_table.php
```diff
@@ -8,15 +8,13 @@ class CreatePagesTable extends Migration
...
-     *
-     * @return void
      */
     public function up()
     {
         // Create table for storing roles
         Schema::create('pages', function (Blueprint $table) {
             $table->increments('id');
-            $table->integer('author_id');
+            $table->unsignedInteger('author_id');
```
fletch3555, line 17: Again

publishable/database/migrations/2016_01_01_000000_create_posts_table.php
```diff
@@ -7,16 +7,14 @@ class CreatePostsTable extends Migration
...
-     * @return void
      */
     public function up()
     {
         // Create table for storing roles
         Schema::create('posts', function (Blueprint $table) {
             $table->increments('id');
-            $table->integer('author_id');
-            $table->integer('category_id')->nullable();
+            $table->unsignedInteger('author_id');
```
fletch3555, line 16: Again

publishable/database/migrations/2016_01_01_000000_create_posts_table.php
```diff
@@ -7,16 +7,14 @@ class CreatePostsTable extends Migration
 {
     /**
      * Run the migrations.
-     *
-     * @return void
```
fletch3555, line 11: Why remove this?

publishable/database/migrations/2016_05_19_173453_create_menu_table.php
```diff
@@ -26,7 +24,7 @@ public function up()
             $table->string('target')->default('_self');
             $table->string('icon_class')->nullable();
             $table->string('color')->nullable();
-            $table->integer('parent_id')->nullable();
+            $table->unsignedInteger('parent_id')->nullable();
```
fletch3555, line 27: Again

resources/views/settings/index.blade.php
```diff
@@ -284,12 +284,16 @@
                                         @elseif($setting->type == "file" && isset( $setting->value ))
                                             <div class="fileType">{{ $setting->value }}</div>
                                         @endif
+                                        @if($setting->type == "image")
+                                            <input type="file" name="{{ $setting->key }}" accept="image/*" />
```
fletch3555, line 288: If you're going to do this, the if should only be around the difference (`accept` attribute), not duplicating the entire input tag.

fletch3555 (commented):
I'm not certain it's the right solution.  I agree with @emptynick that this won't fix that specific issue since you didn't edit the correct file (bread/edit-add.blade.php)

PR #2301, opened 2017-12-03
Feature/routes prefix from config

maxkrasnov:
- transfered admin route to Voyager::routes() and remove from install file
- now can change full path for admin from config file
- breadcrumb fix admin prefix ex 'admin/yetpath/yetpath'

PR #2303, opened 2017-12-03
Fix rich text require

maxkrasnov:
remove 'require' attribute from rich_text_box.blade.php, because:

> Chrome wants to focus on a control that is required but still empty so that it can pop up the message 'Please fill out this field'. However, if the control is hidden at the point that Chrome wants to pop up the message, that is at the time of form submission, Chrome can't focus on the control because it is hidden, therefore the form won't submit.
> 
> So, to get around the problem, when a control is hidden by javascript, we also must remove the 'required' attribute from that control.


fletch3555:
Same comment as #2301

PR #2306, opened 2017-12-04
Feature / router setting only from config file

maxkrasnov:
- router setting only from config file
- remove from install command add ['prefix'=> 'admin'] in routes/web.php

### DIFF
--- a/routes/voyager.php
+++ b/routes/voyager.php
@@ -14,7 +14,7 @@ use TCG\Voyager\Models\DataType;
 |
 */
 
-Route::group(['as' => 'voyager.'], function () {
+Route::group(['as' => 'voyager.', 'prefix' => config('voyager.prefix')], function () {
     event(new Routing());
 
     $namespacePrefix = '\\'.config('voyager.controllers.namespace').'\\';
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -122,8 +122,7 @@ class InstallCommand extends Command
             base_path('routes/web.php'),
             "\n\nVoyager::routes();\n\n"
         );
-
-
+        
         \Route::group(['prefix' => 'admin'], function () {
             \Voyager::routes();
         });
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -101,7 +101,7 @@ class InstallCommand extends Command
         if (false === strpos($routes_contents, 'Voyager::routes()')) {
             $filesystem->append(
                 base_path('routes/web.php'),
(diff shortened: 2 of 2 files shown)


emptynick:
This **WILL** break existing installations.
`routes/voyager.php` will be replaced,
but the install-command will not remove the prefix.
Then you have doubled-prefixes /admin/admin

fletch3555:
Agreed.  I can't let this in `master`.  We'll have to come up with a way to resolve this going forward.  I think we can get it into 1.1 with an upgrade guide stating to remove the group in your `routes/web.php`

emptynick:
This will be fine if there was an UpgradeCommand.
But dont make it harder than it is.
```
$old_route = "\n\nRoute::group(['prefix' => 'admin'], function () {\n    Voyager::routes();\n});\n";
$new_route = "\n\nVoyager::routes();\n\n";
$filesystem->put(base_path('routes/web.php'), str_replace($old_route, '', $routes_contents));
$filesystem->append(base_path('routes/web.php'), $new_route);
```

Written in here - not tested

maxkrasnov:
@emptynick, @fletch3555 check solution:
installer check exist `Voyager::routes()` and remove them from routes/web.php, after we check construction prefix "admin" and if empty, remove it. And after again append `Voyager::routes()`.

Sorry my English :)

emptynick:
You dont run the InstallCommand (`php artisan voyager:install`) if you just update from `1.0.*` to `1.0.*`
Just try setting-up a fresh voyager installation, then check-out your branch `"tcg-voyager": "dev-master#b5099ad"` and try accessing voyager ;)

maxkrasnov:
@emptynick if i'm have few constructions in group for prefix "admin"? we will duplicate the code `Voyager::routes()`

emptynick:
I dont understand what you mean.
The only way to fix it to get it into v1.0 is to check if theres already an admin prefix in `routes/voyager.php` and if yes - dont use `'prefix' => config('voyager.prefix')`.

fletch3555:
I'm going to close this.  The `voyager.prefix` config entry has been removed going forward.  The only thing Voyager should care about is it's current path, and all URLs should be based on that.  That path is defined by the "prefix" given in the `Route::group()`, which I would like to keep in `routes/web.php`

maxkrasnov:
Are you sure this is the right solution? in two places the paths are stored, why not store in one configuration file on param prefix?

emptynick:
Because you have your routes defined in `routes/x.php`.
This should be the only place where routes are imo.

fletch3555:
Because we're refactoring-away those two places.  The route file will be the only instance.

Issue #2288, opened 2017-11-29
Can't save new multilingual page


def-media:
- Laravel Version: 5.5.21
- Voyager Version: 1.0.7
- PHP Version: 7.0
- Database Driver & Version: MySQL 10.1.24-MariaDB-cll-lve

### Description:
Can't save new multilingual page.
The Developer Tools console shows error: "An invalid form control with name='body' is not focusable."
There is a hidden textarea element (name="body", id="richtextbody") with the "required" attribute.
When I press "Save" this textarea element does not pass validation and the form cannot be submitted.
![Voyager bug](https://www.cyclolika.com/img/voyager_cant_save_new_page.jpg)
This error does not occur when editing existing multilingual page, only when saving a new multilingual page.

### Steps To Reproduce:
Enable multilanguage.
Create new multilingual page (fill out the form).
Pres the "Save" button.

luckosracerfish:
"An invalid form control with name='body' is not focusable." still is shown in latest master branch.

Also, why is meta description and keywords required? I would like to decide for myself about this stuff especially that keywords are not used much anymore.

baciualexandru:
Try to change in BREAD, where you have Input body...change with something else like body_new. Hope it helps.

def-media:
Thank you for your answer, but it seems like this problem is solved.
- Laravel Framework 5.5.28
- Voyager v1.0.13
- Linux Ubuntu 16.04
- Google Chrome Version 64.0.3282.119 (Official Build) (64-bit)

In this environment, I can't reproduce this problem.
Somewhere I read that it was a bug in the Google Chrome browser.

Issue #1998, opened 2017-10-13
Rich Text & Translatable, field cannot be saved


bedoz:
- Laravel Version: 5.5.14
- Voyager Version: 1.0.5
- PHP Version: 7.0.24
- Database Driver & Version: MariaDB 5.5.53

### Description:
Hello Everybody,
i tried to search on closed issue without luck.
I found a bug on system.
When i use rich text on BREAD and i have Translatable Trait on model the field cannot be saved.
If i deactivate Translation everything is ok.
The same way if i use a text or textarea field and Translations.

This is a schematic version for easy comprehension:
Rich Text & Translatable = no save field
Not Rich Text & Translatable = save field
Rich Text & Not Translatable = save field

Can anybody find a way to solve this?
thank you very much
Alex

bedoz:
No one have the same issue?

Triztan:
I just tested it with the --with-dummy flag at installation of voyager... It works pretty nice(maybe missing an alert or something when switching between languages, so we know if it was changed correctly)
Maybe you are missing to change at config/voyager.php the multilingual => enabled key to true

bedoz:
Hi Triztan,
this is my config/voyager.php section:
`'multilingual' => [
        'enabled' => true,
        'default' => 'it',
        'locales' => [
            'it',
            'en',
        ],
    ],`
on my models i put `use TCG\Voyager\Traits\Translatable;` and `use Translatable;` into class.
I also set `protected $translatable = ['testo', 'orari'];` to set what fields need to be translated.

I forgot something?

Thank you very much
Alex

bedoz:
Nobody?

vaggelis2018:
I think the problem is here in src/Helpers/helperTranslations.php:
The stripHTMLTags comes in as true and it strips the html. With this way the values are saves in database but there is no proper html to be shown. This happens when multilingual is enabled.
```
function get_field_translations($model, $field, $rowType = '', $stripHtmlTags = false)
    {
        $_out = $model->getTranslationsOf($field);

        if ($stripHtmlTags && $rowType == 'rich_text_box') {
            foreach ($_out as $language => $value) {
                $_out[$language] = strip_tags($_out[$language]);
            }
        }

        return json_encode($_out);
    }
```

AndCymru:
as a comment to this: if you look into the database, all the tags are there, tinymce just won't display them.

bedoz:
In my case, into the database these fields are empty

[...]

PR #2308, opened 2017-12-04
Fix / rich text box require

maxkrasnov:
#2288, #1998
remove 'require' attribute from rich_text_box.blade.php, because:

> Chrome wants to focus on a control that is required but still empty so that it can pop up the message 'Please fill out this field'. However, if the control is hidden at the point that Chrome wants to pop up the message, that is at the time of form submission, Chrome can't focus on the control because it is hidden, therefore the form won't submit.
> 
> So, to get around the problem, when a control is hidden by javascript, we also must remove the 'required' attribute from that control.

### DIFF
--- a/resources/views/formfields/rich_text_box.blade.php
+++ b/resources/views/formfields/rich_text_box.blade.php
@@ -1,4 +1,4 @@
-<textarea @if($row->required == 1) required @endif class="form-control richTextBox" data-name="{{ $row->display_name }}"  name="{{ $row->field }}" id="richtext{{ $row->field }}">
+<textarea @if($row->required == 1) @endif class="form-control richTextBox" data-name="{{ $row->display_name }}"  name="{{ $row->field }}" id="richtext{{ $row->field }}">
     @if(isset($dataTypeContent->{$row->field}))
         {{ old($row->field, $dataTypeContent->{$row->field}) }}
     @else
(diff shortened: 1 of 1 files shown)


resources/views/formfields/rich_text_box.blade.php
```diff
@@ -1,4 +1,4 @@
-<textarea @if($row->required == 1) required @endif class="form-control richTextBox" data-name="{{ $row->display_name }}"  name="{{ $row->field }}" id="richtext{{ $row->field }}">
+<textarea @if($row->required == 1) @endif class="form-control richTextBox" data-name="{{ $row->display_name }}"  name="{{ $row->field }}" id="richtext{{ $row->field }}">
```
fletch3555: If you're going to remove `required`, you need to remove the empty `@if() ... @endif`` around it.

fletch3555 (changes requested):

fletch3555:
I'm not too familiar with the i18n functionality, but I don't believe this is the correct solution.  @akazorg, do you have any thoughts/suggestions?

resources/views/formfields/rich_text_box.blade.php
maxkrasnov: ouhc, in a hurry missed

akazorg:
@fletch3555 I wasn't aware of that Chrome bug. 
If this solution works, then looks great :)
Thanks for the PR @maxkrasnov!

fletch3555:
@akazorg, how do the i18n fields work? Does it keep the normal EN field on the page but hidden?  If so, this would fix it.  If not, this just removes the ability to make the field required.  That's why I'm not sure this was the best fix.

maxkrasnov:
@fletch3555 , @akazorg maybe add custom check require on js for this? agree?

akazorg:
@maxkrasnov sorry but I don't have an answer, need to test this.

@fletch3555 the way I18N works is using an hidden input for each translatable field. This field holds a JSON object containing all available translations for each input.
When the user select another language, the JS will update all hidden fields with the previous language on screen.
Once User submits the form, the JS will again update hidden fields, and then submits the form.

yourPixel:
@maxkrasnov Thank you very much !!!

PR #2313, opened 2017-12-06
Updated default data type for timestamp column

sanampatel:
Updated default data type for timestamp column from timestamp to datetime, which resolve the error!


fletch3555:
> which resolve the error!

What error? Is there an open issue? We need more detail than this.

sanampatel:
Just changed to datatype to datetime from timestamp. Cause timestamp data type is not available, which gives error while adding timestamp in database creating form!

image

fletch3555:
You need to either fully document the issue you're reporting (and supposedly fixing), or link to an issue that does.  We can't blindly accept changes without knowing why they're made.

fletch3555:
Closing.  We can reopen this once my previous question is answered.

PR #2320, opened 2017-12-08
Use assertArrayHasKey and assertNotEquals

carusogabriel:


fletch3555:
This change doesn't really do anything, but it does look cleaner.  Thanks!

Issue #2317, opened 2017-12-07
Improving support for special chars


xotik:
Hello everyone,

Today I stumbled against the following (I'm not using version 1.0 yet, it might have changed by now):

In the browse.blade.php view of BREAD, when the field is of type 'text' or 'textarea', text gets trimmed if larger than 200 chars, right?

Then I started seeing some fields, not all, appearing blank in the table. After some debugging, I found out that when the max char in strlen() is a special char, substr() can't handle it very nice.

Luckily, there's an easy fix: just use mb_substr() instead.

Is it a good idea to update all Voyager code to use mb_substr() instead of substr()? I´ll leave it up to you :)

Cheers!

emptynick:
I see your point.
`strlen` does not count chars, but bytes.
Definitely worth considering!

xotik:
Glad I could help!

fletch3555:
Open to pull requests.

PR #2330, opened 2017-12-10
Fix for issue #2317

dfk7677:
I just replaced function `substr` with `mb_substr` in [browse.blade.php](https://github.com/the-control-group/voyager/blob/master/resources/views/bread/browse.blade.php) as proposed.

Function `substr` is also used in [relationship.blade.php](https://github.com/the-control-group/voyager/blob/master/resources/views/formfields/relationship.blade.php), but I am not sure that this is an issue.

Fixes [#2317](https://github.com/the-control-group/voyager/issues/2317)


emptynick:
Only using `mb_substr` does not make sense if you dont also use `mb_strlen`.
Otherwise it still counts wrong. Even it is cutted correctly.

fletch3555:
@dfk7677 I had to clean up your branch a bit.  You may have issues pulling as the history was changed.  In this case, just delete your local branch and pull it fresh.

PR #2336, opened 2017-12-11
Moved factory definitions to config.

gluttons:
Moved factory definitions to config.
Updated old static references to models to use new factory methods.


gluttons:
These changes makes it possible to override stuff in a more consistent way, as far as I've tested the changes does not create any issues. It mostly abstracts stuff to config and makes use of the factory methods where it wasn't before.

publishable/config/voyager.php
```diff
@@ -1,6 +1,46 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Policies config
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify policies for certain models
+    |
+    */
+
+    'policies' => [
```
fletch3555: These two policies were defined in `VoyagerServiceProvider` because they don't have BREAD to set the policy on (see `VoyagerServiceProvider::registerGates()`).  I think moving this to a config, while normally s good idea, will lead to confusion with Voyager users.  Instead of making the change through the UI, they'll do it through config.  This may be a better practice for most apps, but is contrary to the purpose of Voyager.

src/Helpers/helpers.php
```diff
@@ -10,7 +10,7 @@ function setting($key, $default = null)
 if (!function_exists('menu')) {
     function menu($menuName, $type = null, array $options = [])
     {
-        return TCG\Voyager\Models\Menu::display($menuName, $type, $options);
+        return TCG\Voyager\Facades\Voyager::model('Menu')::display($menuName, $type, $options);
```
fletch3555, line 13: Thank you for this change.  This is what it should've been anyway.

src/Models/MenuItem.php
```diff
@@ -22,13 +23,13 @@ class MenuItem extends Model
 
     public function children()
     {
-        return $this->hasMany('TCG\Voyager\Models\MenuItem', 'parent_id')
+        return $this->hasMany(Voyager::modelClass('MenuItem'), 'parent_id')
```
fletch3555, line 26: Again, thank you for this change. This is what it should've been anyway.

publishable/config/voyager.php
```diff
@@ -1,6 +1,46 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Factory config
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify what classes should be returned when using e.g Voyager::model('Category')
+    |
+    */
+
+    'factory' => [
```
fletch3555, line 14: Where did the "factory" name come from? I'm not opposed to it, just curious

publishable/config/voyager.php
```diff
@@ -1,6 +1,46 @@
...
+        'models' => [
```
fletch3555, line 15: Similar comment to the policies config below this.  This was originally done to encourage use of Voyager's as a UI to manage your app, however in this case, there is no UI for the models.  I'm okay with this change, but I think we should at least deprecate the `Voyager::useModel()` method in favor of this config-based approach.

fletch3555 (changes requested):
Due to the changes to the config, this would break existing sites upon updating.  I'm going to bump it to `develop` (v1.1 pre-release).  Also, please see my comments below.

fletch3555:
I have no idea why all those other commits are showing up... I just checked the actual git branch and it looks fine to me.

publishable/config/voyager.php
gluttons: I had a hard time naming this config because of a conflict in the config-file, where the "models"-key is used to change the default namespace of BREAD. So I had to think twice about what it actually was, and factory came to mind.

publishable/config/voyager.php
gluttons: I must admit I'm new to the voyager environment, the reason behind me wanting to move these to config was because I'm working on a 'multi-site' sort of application where admins can only see part of the collections, but still have read, edit and delete privileges.

Voyager worked really well except for some edge cases, like these hard-coded values within the class. Thats why I thought it good to extract them to config.

I can solve this by overriding Voyager.php in my own app and override the facade as well, if you don't think this is suited for the overall purpose of voyager.

publishable/config/voyager.php
gluttons: The reason behind this was as noted above for overridability, if i don't use the set class for these policies it would not work as expected. 

I could fix this setting by being more dynamic in the provider instead, which might be a better Idea actually. to avoid confusion but still keep it dynamic. __I'll fix it ASAP!__

publishable/config/voyager.php
fletch3555: I don't think you need to override the whole class just to change some of these values.  Voyager provides a `useModel()` method on the facade, which lets you override these "hard-coded" model classes.  Policies can be overridden by (re)defining a Policy in your app's AuthServiceProvider, so long as that provider runs after VoyagerServiceProvider (Laravel 5.5's autodiscovery mechanism might make this more difficult, but it's possible)

gluttons: Oh! i completely missed that setting! How do you want me to go forward with this PR then? remove the config or deprecate useModel?

gluttons:
I updated the PR and removed policies from the config file, I also deprecated the useModels method as requested.

gluttons:
I will split this into 2 PR's aswell. one containing the bug-fixes with the factory models and one with the change to configs.

Issue #2292, opened 2017-11-29
Accessing a bread-table creates 10x the row-count of queries


JapSeyz:
I have 300 users in my users-table, when trying to access /users

It creates over 3000 queries, 2974 of which are duplicates.
The databtable should fetch new rows on demand, but it seems it loads every single one of them as soon as I try to access the page.

![screenshot from 2017-11-29 21 46 02](https://user-images.githubusercontent.com/2234034/33398442-41ca52ac-d54f-11e7-950c-1324ed2355ee.png)

JapSeyz:
The query actually took about 2 minutes, not two seconds as Debugbar seems to think

JapSeyz:
Even though I have translations turned off, it seems at least 80% of the calls is this single query, ofc with every key being different:

```sql
select * from `ltm_translations` where (`locale` = 'da' and `group` = 'voyager' and `key` = 'generic.bulk_delete') limit 1
```

JapSeyz:
commenting out

```php
if (($isModelTranslatable = is_bread_translatable($model))) {
    $dataTypeContent->load('translations');
}
```

In the BreadController didn't help. I'm investigation further

JapSeyz:
I managed to tune it down to only 4x the amount of row-count as queries.

The 1600 of the calls was Laravel Translation Manager, trying to reach the generic.error translation in the database instead of the fallback language. **Not Voyagers fault**.

![screenshot from 2017-11-29 22 09 48](https://user-images.githubusercontent.com/2234034/33399279-57370aba-d552-11e7-9772-3d5e8b813a48.png)

![screenshot from 2017-11-29 22 10 07](https://user-images.githubusercontent.com/2234034/33399286-5b658792-d552-11e7-83fd-3c93df0ca236.png)

![screenshot from 2017-11-29 22 09 40](https://user-images.githubusercontent.com/2234034/33399297-669bb4e2-d552-11e7-9f45-3946f558575b.png)

![screenshot from 2017-11-29 22 09 23](https://user-images.githubusercontent.com/2234034/33399308-71dba2a4-d552-11e7-92dc-7f7d2d5d866f.png)

JapSeyz:
Still 1.1k queries that's not unique

JapSeyz:
Albeit, not pretty, adding this in basePolicy takes it down to 330 queries:

```php
if (! isset(self::$datatypes[get_class($model)])) {
            $dataType = Voyager::model('DataType');
[...]

PR #2337, opened 2017-12-11
Optimised javascript datatables

JapSeyz:
To see an explanation, checkout #2292


fletch3555:
I don't see any reason why we can't get this into master for the next "bugfix" release, but it looks like you created your branch from `develop`.  Would you mind cherry-picking your changes into a new branch created from `master`?  We need to avoid prematurely introducing breaking changes that are currently in `develop`

PR #2339, opened 2017-12-11
Moved optimisations to master

JapSeyz:
Moved from develop to master-branch by request of @fletch3555

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -22,7 +22,7 @@ class Role extends Model
 
     public static function getRelationship($id)
     {
-        if (! isset(self::$relationships[$id])) {
+        if (!isset(self::$relationships[$id])) {
             self::$relationships[$id] = self::find($id);
         }
 
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -8,6 +8,7 @@ use TCG\Voyager\Facades\Voyager;
 class Role extends Model
 {
     protected $guarded = [];
+    protected static $relationships = [];
 
     public function users()
     {
@@ -17,5 +18,14 @@ class Role extends Model
     public function permissions()
     {
         return $this->belongsToMany(Voyager::modelClass('Permission'));
+    }
+
+    public static function getRelationship(int $id)
+    {
+        if (! isset(self::$relationships[$id])) {
+            self::$relationships[$id] = self::find($id);
+        }
+
+        return self::$relationships[$id];
     }
 }
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -20,7 +20,7 @@ class Role extends Model
         return $this->belongsToMany(Voyager::modelClass('Permission'));
     }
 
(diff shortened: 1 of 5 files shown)


fletch3555:
Not sure why Travis is unhappy.  Did you run phpunit locally? (just run `composer install` from project root and then `./vendor/bin/phpunit`).  If not, I'll try to do so later today.

JapSeyz:
I'll do that real quick. It's setup to use my phpcs settings.

JapSeyz:
@fletch3555 I am getting the same error, with an incorrect whitelist. I don't know why that's happening but the first PR on the dev branch was successful.

Could it be the "add more tests" pr that you guys merged?

fletch3555:
No, because that PR wasn't merged into the 1.0 stream.

The "no whitelist configured" Error line can be ignored for now.  That exists on successful runs as well.  The issue here is the "no output received in 10 minutes" line, which is new.  Do you get that when running it locally as well?

JapSeyz:
@fletch3555 It just runs indefinitely without any output.

fletch3555:
Okay, please merge the `1.0` branch into your own so we know the ONLY difference is your one commit.  Then run it again.  If it continues to hang, change to the 1.0 branch and try again.  If it doesn't hang there, it's somehow related to your commit.  If it does, it's not your commit, and we'll have to try something else to track it down.

JapSeyz:
@fletch3555 Should be fixed now. I had forgotten about the return statement before the ternary operator in VoyagerAdminMiddleware.php

fletch3555:
I had to modify your branch.  You merged your branch into 1.0 instead of 1.0 into your branch.  You may need to delete your branch locally and check it out fresh again if any future changes need to be made on it.

Once TravisCI is happy, we'll get someone on QA-reviewing this.  Thanks!

JapSeyz:
@fletch3555 Of course, sorry for the mess-up. I am a bit swallowed up by work at the moment.
Glad you were able to fix it quickly.

fletch3555:
No worries, so am I haha.

JapSeyz:
@fletch3555 I'll let you know if I find any more bottlenecks, but giving the app a once-over with Debugbar should show you where the bottlenecks are.

src/Models/Role.php
```diff
@@ -18,4 +19,13 @@ public function permissions()
     {
         return $this->belongsToMany(Voyager::modelClass('Permission'));
     }
+
+    public static function getRelationship($id)
```
fletch3555, line 23: This was only added in the `Role` model, but relationships can be used in any model... This is breaking Voyager: #2431

JapSeyz: This should be added to the base voyager model, or as a trait.

fletch3555: Agreed: #2432

JapSeyz: I have also created one in #2433

PR #2340, opened 2017-12-12
Only publish necessary files on install.

gluttons:
* Loading migrations and translations from provider.
* Namespacing translations.
* Updated install command for 5.5

This PR could be split in two parts, but I thought it was related. I wanted to load translations via service provider instead of publishing them, for sanity. And with that Translations came as a natural candidate for the same cause.

I extracted the language files to multiple files and it uses voyager as a namespace instead, I only did this for the /en/ language for now, until it is reviewed. If the proposed changes are accepted I will fix the remaining languages aswelll.


gluttons:
actual changes are in:
src/Commands/InstallCommand.php
src/VoyagerServiceProvider.php

rest is translations

gluttons:
I will split this PR into 2 and resubmit them

PR #2341, opened 2017-12-12
Fix typo in settings blade

xtrees:
fix typo in settings blade


fletch3555:
That's an amusing typo.  Good catch, thanks!

PR #2343, opened 2017-12-12
Loading migrations by default, instead of publishing, on voyager::install

gluttons:
Loading migrations from provider.
Updated install command for 5.4

### DIFF
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -69,7 +69,11 @@ class InstallCommand extends Command
         $this->call('hook:setup');
 
         $this->info('Publishing the Voyager assets, database, language, and config files');
-        $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class]);
+
+        //Publish only relevant resources on install
+        $tags = ['voyager_assets', 'seeds', 'demo_content', 'config', 'lang'];
+
+        $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class, '--tag' => $tags]);
         $this->call('vendor:publish', ['--provider' => ImageServiceProviderLaravel5::class]);
 
         $this->info('Migrating the database tables into your application');
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -95,14 +95,11 @@ class VoyagerServiceProvider extends ServiceProvider
 
         $this->loadViewsFrom(__DIR__.'/../resources/views', 'voyager');
 
-        if (app()->version() >= 5.4) {
-            $router->aliasMiddleware('admin.user', VoyagerAdminMiddleware::class);
+        $router->aliasMiddleware('admin.user', VoyagerAdminMiddleware::class);
+        $this->loadMigrationsFrom(realpath(__DIR__.'/../publishable/database/migrations'));
 
-            if (config('app.env') == 'testing') {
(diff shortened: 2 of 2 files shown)


src/Commands/InstallCommand.php
```diff
@@ -69,7 +70,19 @@ public function handle(Filesystem $filesystem)
         $this->call('hook:setup');
 
         $this->info('Publishing the Voyager assets, database, language, and config files');
-        $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class]);
+
+        //Publish only relevant resources on install
+        $tags = ['voyager_assets', 'seeds', 'demo_content', 'config', 'migrations', 'lang'];
+
+        if (app()->version() >= 5.4) {
+            $tags = array_diff($tags, ['migrations']);
```
fletch3555: This is some REALLY weird logic... you define the full array, then remove an entry... then add it back.  This needs to be changed

src/Commands/InstallCommand.php
```diff
@@ -69,7 +70,19 @@ public function handle(Filesystem $filesystem)
         $this->call('hook:setup');
 
         $this->info('Publishing the Voyager assets, database, language, and config files');
-        $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class]);
+
+        //Publish only relevant resources on install
+        $tags = ['voyager_assets', 'seeds', 'demo_content', 'config', 'migrations', 'lang'];
+
+        if (app()->version() >= 5.4) {
```
fletch3555: This check is unnecessary.  Voyager currently only supports 5.4+

src/VoyagerServiceProvider.php
```diff
@@ -97,6 +97,7 @@ public function boot(Router $router, Dispatcher $event)
 
         if (app()->version() >= 5.4) {
             $router->aliasMiddleware('admin.user', VoyagerAdminMiddleware::class);
+            $this->loadMigrationsFrom(realpath(__DIR__.'/../publishable/database/migrations'));
```
fletch3555: This may cause errors due to the migrations running twice if someone has already published their migrations.  Also, there's no need to keep it inside the 5.4+ check

fletch3555 (changes requested):

gluttons:
I've removed the 5.4 checks, fixed the wierd logic that caused, and removed the abillity to publish migrations as well.

src/VoyagerServiceProvider.php
```diff
@@ -95,14 +95,11 @@ public function boot(Router $router, Dispatcher $event)
 
         $this->loadViewsFrom(__DIR__.'/../resources/views', 'voyager');
 
-        if (app()->version() >= 5.4) {
-            $router->aliasMiddleware('admin.user', VoyagerAdminMiddleware::class);
+        $router->aliasMiddleware('admin.user', VoyagerAdminMiddleware::class);
+        $this->loadMigrationsFrom(realpath(__DIR__.'/../publishable/database/migrations'));
```
fletch3555, line 99: If migrations are no longer publishable, we should probably move them out of that directory.  We can handle that in a subsequent PR.  No need to change it now, just wanted it documented.

fletch3555 (approved):

fletch3555:
Expecting merge conflicts between this and #2344, depending which gets merged first.

PR #2345, opened 2017-12-12
Moved factory model definitions to config for easy overrides

gluttons:


gluttons:
Closing this because of issues with models, will have new PR coming up to address these issues first

PR #2350, opened 2017-12-12
Use model factory in voyager facade

gluttons:
Found some more places where we where accessing the models directly


fletch3555:
👍 Thanks!

akazorg:
Great job, we really need it.

PR #2362, opened 2017-12-14
Fix direct usage of Voyager model class

### LINKED ISSUE #2354: rename users table to admin_users
- Laravel Version: 5.4.36
- Voyager Version: ^1.0
- PHP Version: 5.6

### Description:

I am using Voyager as a backend for web service. I want admin users in `admin_users` table, for our primary users, I want to use `users` . 

I am installing Voyager with dummy data. I have updated seeds, migrations and `App\Users`  to use `admin_users` table.  but it's not working properly. 

The only way it works when I add  `protected $table =  'admin_users';` to `TCG\Voyager\Models\User`. 

is there any better way to extend `TCG\Voyager\Models\User` model to use `admin_users` table.

Error:

I get this error when I access `admin/database` or `admin/media`
Error

`AppServiceProvider.php`

    use TCG\Voyager\Facades\Voyager;

    {
        public function boot()
        {
            Voyager::useModel('User', \App\User::class);
        }
    }

`User.php`

    namespace App;

    use Illuminate\Notifications\Notifiable;
    use Illuminate\Foundation\Auth\User as Authenticatable;

    class User extends \TCG\Voyager\Models\User
    {
        use Notifiable;
        protected $table = 'admin_users';
    }

fletch3555:
Fixes #2354

### DIFF
--- a/src/Voyager.php
+++ b/src/Voyager.php
@@ -319,7 +319,7 @@ class Voyager
         }
 
         if (!isset($this->users[$id])) {
-            $this->users[$id] = User::find($id);
+            $this->users[$id] = $this->model('User')->find($id);
         }
 
         return $this->users[$id];


gluttons:
@fletch3555 this fix is also in PR #2350, which also addresses some other direct model usages. The only difference being `$this->model()` instead of `self::model` which I used.

fletch3555:
Thank you for pointing that out @gluttons.  Closing this in favor of #2350

Issue #2369, opened 2017-12-17
Laravel Timezone Ignored by BREADs


dattz:
- Laravel Version: 5.4.36
- Voyager Version: 1.0.7
- PHP Version: 7.0.22
- Database Driver & Version: MySQL 5.7.20

### Description:
Bread's timestamp fields are converted to GMT even when laravel's timezone is set to something else.

### Steps To Reproduce:
1. Change Laravel Timezone from GMT to anything else (config/app.php)
2. Create a new BREAD with a timestamp field.
3. Add an entry to that bread. Note the value of the timestamp field you add.
4. The date is converted to GMT and stored in the database ignoring the Laravel's timezone settings.

### Possible Cause:
As also mentioned by @emptynick in the Slack's help Channel, the cause of this exists in the following files that use the [gmdate](http://php.net/manual/en/function.gmdate.php) method:
1. In the timestamp formfield view
2. and in the Controller.php

marktopper:
This has been fixed in `v1.x-dev` which will be released as `v1.1` sometime soon.

PR #2370, opened 2017-12-17
Changed timestamp files to use Carbon instead of gmdate method

dattz:
Possible fix for the Issue #2369 
Replaced usage of gmdate method with Carbon that will follow Laravel's timezone settings.


handiwijoyo:
@dattz Thanks man!

defaultpage:
Thanks! Very helpful for me! 💯

defaultpage:
As temporary solution we can create custom view with @dattz view changes and custom controller with next code:
```php
<?php

namespace App\Http\Controllers;

use App\Traits\FixBreadTimestampTrait;
use TCG\Voyager\Http\Controllers\VoyagerBreadController;

class FixedBreadTimestampController extends VoyagerBreadController
{
   use FixBreadTimestampTrait;
}
```
FixBreadTimestampTrait:
```php
<?php

namespace App\Traits;

use Illuminate\Http\Request;
use Carbon\Carbon;

trait FixBreadTimestampTrait
{
    public function getContentBasedOnType(Request $request, $slug, $row)
    {
        $content = null;
        switch ($row->type) {
            case 'timestamp':
                $content = $request->input($row->field);
                if (in_array($request->method(), ['PUT', 'POST'])) {
                    if (empty($request->input($row->field))) {
                        $content = null;
                    } else {
                        $content = Carbon::parse($content);
                    }
                }
                break;
            default:
                $content = parent::getContentBasedOnType($request, $slug, $row);
        }

        return $content;
    }
}
```

PR #2377, opened 2017-12-19
Fixes issue with erasing fields data

sergbbb:
When field is emty it is not saved, and leave last data


seilem:
Same things but with relationship

https://github.com/the-control-group/voyager/pull/2378

emptynick:
Which fields dont get saved when the value is empty?
I just tried it on Categories and its working.

Edit:
From L5.5 `has()` and `exists()` is the same. [See here](https://github.com/laravel/framework/blob/79f34c859046c3ddecede41377ab2a75e208548b/src/Illuminate/Http/Concerns/InteractsWithInput.php#L62-L92)
L5.5 upwards should fix this problem.

PR #2378, opened 2017-12-19
Relationship is exception now too.

seilem:
There was a bug when you couldn't delete the last element of relationship type.


reficul:
This PR fix an issue for me. Select2 and BelongToMany not save value on DB without this PR.

PR #2382, opened 2017-12-21
Fix relation error with Permission and role

xtrees:
change relation from hasMany to  belongsToMany

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -10,7 +10,7 @@ class Permission extends Model
 
     public function roles()
     {
-        return $this->hasMany(Role::class);
+        return $this->belongsToMany(Role::class);
     }
 
     public static function generateFor($table_name)


fletch3555:
I feel like this was done intentionally, but I can't conceive of any possible reason why...  Did you have a problem that this fixed?

xtrees:
I want get roles related to one permission.  They connected through middle table permission-role ，hasMany does not work should be belongs to Many

fletch3555:
I agree that it is in fact (currently) a many-to-many relationship, which requires the use of `->belongsToMany()`, but I was just wondering if you ran into an issue, or if you were just looking at the code and thought it looked wrong.

xtrees:
For example .I want list all of my roles who has the ability to manage posts so i can group users in those roles together. No need to check one by one

xtrees:
When I do this.  Call the roles() it shows error so I only create My permission model override the roles function

PR #2393, opened 2017-12-25
Add integrity constraint violation handling to BREAD Controller

vnilov:
Handle an error to make it more user friendly.


fletch3555:
I agree it should gracefully handle the error, but you should also have appropriate validation in place so it never gets to that point

PR #2400, opened 2017-12-27
Patch #2398 Move settings from VoyagerDummyDatabaseSeeder to VoyagerDatabaseSeeder

### LINKED ISSUE #2398: Add default settings when without using dummy flag
When install voyager without dummy flag, the settings are empty. It's hard to figure out what key need to add for something like web title and description. Better we leave the default settings included.

image

```The title is empty:```

image

I will submit PR for this.

handiwijoyo:
Fixes #2398


fletch3555:
👍

PR #2401, opened 2017-12-27
Adds the Persian language and prefixes field with its table name

### LINKED ISSUE #2392: Define many-to-many relationships
I have three tables:
`posts`: default voyager posts table
`post_tag`: the Post-Tag pivot table 
`tags`: tags table which has only id, name, and timestamp fields

You know each `Post` may have many tags and each `Tag` may have many posts, I wanna add a multi-select field into the create-post view, so the user can select multiple tags for a post. (we already can select a category in the create-post view but can't select multiple categories at a time for a post)

Is this possible to implement this using voyager relationship tool?

P9Q:
Fixes #2392


P9Q:
Both Arabic and Persian are RTL languages, I simply translated English locale file into Persian, But for a better translation, I need RTL version of Voyager, I will improve the translation as soon as you release RTL support.

fletch3555:
I can't merge this in as-is... Please limit your PRs to one "fix" each.  In this case, adding the Persion language should be one, and the table prefix should be another.

Issue #2205, opened 2017-11-10
Validation problem


AhmadOf:
Laravel Version: 5.5.20
Voyager Version: 1.0.7
PHP Version: 7.1.10
Database Driver & Version: MySQL 5.7.20

### Description:
Validation not working when trying to add a new record of a type which have some validation rules.

### Steps To Reproduce:
1. Add validation rule to any BREAD like this:
```
{
    "validation": {
        "rule": "required",
        "messages": {
            "required": "This :attribute field is a must."
        }
    }
}
```
2. trying to add a record from this type.
3. No thing happen if the field is empty so no error message and can't add the record.
4. Checking JS console give the error:
Cannot read property 'top' of undefined at Array.<anonymous> (app.js:24) ...., actually this line is related to displaying errors as I think and trying to fix the error manually I've changed the code:
```
$.each(e.errors,function(t,n){0===Object.keys(e.errors).indexOf(t)&&$("html, body").animate({scrollTop:$("[name='"+t+"']").parent().offset().top-$("nav.navbar").height()+"px"},"fast"),$("[name='"+t+"']").parent().addClass("has-error"),$("[name='"+t+"']").parent().append("<span class='help-block' style='color:#f96868'>"+n+"</span>")})},error:function(){$(n).unbind("submit").submit()}})})})},,,function(e,t,n){var i,r;"function"==typeof Symbol&&Symbol.iterator;
```

to this:

```
$.each(e.errors,function(t,n){0===Object.keys(e.errors).indexOf(t)&&$("html, body").animate({scrollTop:$("[data-name='"+t+"']").parent().offset().top-$("nav.navbar").height()+"px"},"fast"),$("[data-name='"+t+"']").parent().addClass("has-error"),$("[data-name='"+t+"']").parent().append("<span class='help-block' style='color:#f96868'>"+n+"</span>")})},error:function(){$(n).unbind("submit").submit()}})})})},,,function(e,t,n){var i,r;"function"==typeof Symbol&&Symbol.iterator;
```

and the problem solved partially but I get another problem that the error message shown and still shown despite of the field is not empty.

vitormicillo:
I have the same problem

erayusta:
I have the same problem extented bread

rttolipas:
Same problem I encountered here. Any solution?

caveoIT:
Same problem after migrate the database with self generated tables or columns, So i think the problem is that voyager didn't accept migration from outside Voyager UI.

[...]

PR #2414, opened 2017-12-31
Validation problem #2205

mrkindy:
Validation problem #2205


emptynick:
This was also fixed in #2660

PR #2422, opened 2018-01-03
Add resizable traits

handiwijoyo:
This traits will allow us to use thumbnail helper to generate image url when using Image additional fields. See: https://voyager.readme.io/docs/additional-field-options#section-image

Example of use:
```

class Post extends Model
{
    use Resizable;
}
```

Use it on view (usually on your site frontend):
```
@foreach($posts as $post)
    {{$post->thumbnail('small')}}
@endforeach
```

Or you can specify the optional image field name (attribute), default to `image`
```
@foreach($posts as $post)
    {{$post->thumbnail('small', 'photo')}}
@endforeach
```

### DIFF
--- a/src/Models/Post.php
+++ b/src/Models/Post.php
@@ -6,11 +6,13 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Auth;
 use TCG\Voyager\Facades\Voyager;
+use TCG\Voyager\Traits\Resizable;
 use TCG\Voyager\Traits\Translatable;
 
 class Post extends Model
 {
     use Translatable;
+    use Resizable;
 
     protected $translatable = ['title', 'seo_title', 'excerpt', 'body', 'slug', 'meta_description', 'meta_keywords'];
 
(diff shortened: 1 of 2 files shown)


fletch3555:
@handiwijoyo, would you mind updating the docs for this?

handiwijoyo:
@fletch3555 I think we need to add new `Helpers` section in the docs, that will document all of available helpers function

fletch3555:
I'm inclined to agree.  I'll create the page later today if you'd like to suggest edits for it.  Not sure if I can give you access to Readme.io, but if so, I'll do that instead

def-media:
Please consider replacing "rtrim" function.

I had the next situation:
$name = rtrim('posts/January2018/OLJEeEfdQi4mI7Cgbhae.jpeg', '.jpeg')
which returns 
"posts/January2018/OLJEeEfdQi4mI7Cgbha"
(missing the last "e" character in the string).

It seems like "rtrim" also removes the last character of the file name if that character is present in the file exstension.

I suggest replacing: 
$name = rtrim($image, '.'.$ext);
with 
$name = \Illuminate\Support\Str::replaceLast('.'.$ext, '', $image); .

It works in my case.

PR #2425, opened 2018-01-04
mysql spatial functions are standardized to (lon, lat)

danielwashbrook:
Mysql Spatial functions use x,y or long,lat so to be consistent with other libraries the coordinates fields should store values in the same manner.
https://dev.mysql.com/doc/refman/8.0/en/gis-mysql-specific-functions.html#function_point


fletch3555:
Can you provide references where it says this is the standard order? The link you provided just says `'x','y'`

danielwashbrook:
Ah sorry, so an x,y of the world as x runs left to right and longitude runs left to right. Since POINT can be used for many things, it isn't specific, but here as other functions based on POINT that specify long,lat
https://dev.mysql.com/doc/refman/5.7/en/spatial-geohash-functions.html

In the GeoJSON spec it states
"Point coordinates are in x, y order (easting, northing for projected coordinates, longitude, latitude for geographic coordinates)"
http://geojson.org/geojson-spec.html#id2

Can't believe I'm going to reference stack overflow but...
https://stackoverflow.com/questions/5756232/moving-lat-lon-text-columns-into-a-point-type-column/5757054#5757054
https://dba.stackexchange.com/questions/182136/does-mysql-store-point-datatypes-as-lat-lng-or-lng-lat/182626

fletch3555:
I had a feeling you were equating x/y to long/lat due to the apparent directions on a standard graph/plot.  I just wanted to make sure that wasn't your only basis for this change.  The other references (specifically geojson and mysql.. SO is far from a "standardized spec") are exactly what I was hoping you would provide.  So thanks for that!

I'll merge this in, but I have to bump it to the v1.1 stream since there are devs already using these points, and this would certainly break their apps.  I don't have a set date for v1.1 yet, but it should be soon (within the next month or two probably)

fletch3555 (approved):

PR #2426, opened 2018-01-04
Caching setting values

JapSeyz:


handiwijoyo:
@JapSeyz Can you provide the details why do we need this?

JapSeyz:
@handiwijoyo Caching the setting to avoid hitting the DB on every page-load for the setting.

Same thing could be done for menu.

I'll admit, it's not the biggest boost, but for API rutes that needs a single setting value, and can be called a few hundred times a minute at a peak hour, calling the DB for the setting-value every time is redundant if it can be stored in Redis.

It doesn't give you any noticable difference on normal view-based routes, but it can really add up on API routes.

handiwijoyo:
@JapSeyz Ok. The idea looks fine for me. But I think we need to add events for settings controller. Then we can put the cache inside the listener and have a config option to enable and disable the cache.

handiwijoyo:
After take a look into the Voyager::setting() https://github.com/the-control-group/voyager/blob/1.0/src/Voyager.php#L161

image

It's already have different cache method. We might need to change this too.

JapSeyz:
@handiwijoyo That cache is only request-wide. Once the request is over, the caching is removed.

I'll look into implementing the global cache there too, and I'll update the PR during the weekend.

I'll also look into the events in the weekend.

JapSeyz:
@handiwijoyo I have added a listener and moved the code to the Voyager.php file.

If you agree with what's done, I'll fix the StyleCI issues

src/Events/SettingUpdated.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
```
fletch3555: Why is this added here? We don't have it elsewhere in the project.  It may be a good idea, I don't know, but I'm a bigger fan of consistency.  Please remove it from this and other files.

JapSeyz: I will, it's one of the things from my php Cs fixer.

Once one of you approve that the actual code is alright I'll give it a once over and remove any styleCI issues as well as these. I think I also wrote it in the comment above.

src/Voyager.php
```diff
@@ -160,10 +160,27 @@ public function formFields()
 
     public function setting($key, $default = null)
     {
+        $globalCache = config('voyager.settings.cache', false);
+
+        if ($globalCache && Cache::tags('settings')->has($key)) {
+            return Cache::tags('settings')->get($key);
+        }
+
         if ($this->setting_cache === null) {
```
fletch3555, line 218: Was it intentional to leave the per-request/local cache in place?  I feel like this would be much simpler if it was simply replaced with `Cache::` calls like you've already done.

fletch3555:
@JapSeyz, sorry for the delay in review here.  Looks good.  Only question is stated above, why the multiple layers of cache?

src/Voyager.php
JapSeyz: It's been a while, but if I recall correctly, it was to disallow users to use the caching function, if they for some reason wanted to do that. It can be stripped out without issues

src/Listeners/ClearCachedSettingValue.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Listeners;
```
fletch3555: TravisCI is failing due to this line.  Please change it to the correct namespace (`TCG\Voyager\Listeners`)

src/Helpers/helpers.php
```diff
@@ -1,20 +1,22 @@
 <?php
 
-if (!function_exists('setting')) {
+declare(strict_types=1);
+
+if (!\function_exists('setting')) {
```
fletch3555: None of the changes in this file are necessary for this PR.  Please revert them.

src/Events/SettingUpdated.php
fletch3555: Sorry, I know this has been a while, but if you don't mind cleaning up the phpcs changes, I'm good with it.

I'm not too worried about StyleCI, since that will fix itself.

Thanks!

fletch3555 (changes requested):

fletch3555 (approved):

Issue #2431, opened 2018-01-04
Belongs To relationship.blade.php exception


vaidotasposkus:
- Laravel Version: 5.5.28
- Voyager Version: 1.0.0
- PHP Version: 7
- Database Driver & Version:

### Description:

Something wrong with my application. When I try to in list i get exception Call to undefined method Illuminate\Database\Query\Builder::getRelationship() (View: C:\Work\shooting_calendar\vendor\tcg\voyager\resources\views\formfields\relationship.blade.php) (View: X:\xx\xx\vendor\tcg\voyager\resources\views\formfields\relationship.blade.php). And don't know why it doesn;t work. Does someone have the same proble?

aruberutochan:
I have the same problem. 

- Laravel Version: 5.5.28
- Voyager Version: 1.0.0
- PHP Version: 7.0
- Database Mariadb

I realised that changing the file /vendor/tcg/voyager/resources/views/formfields/relationship.blade.php to the previous version in the repository It goes well, basically you have to change the line 14:
`
$query = $model::getRelationship($relationshipData->{$options->column});`
by 
`$query = $model::find($relationshipData->{$options->column});`

I am new on voyager and laravel and I don't know what consequences involves that change so i don't want to make a pull request by now.

Maybe if you don't want to modify the code in the tcg/voyager vendor folder, you could temporarily edit your user model adding this to the App\User
```

class User extends \TCG\Voyager\Models\User

{

    // ... More code here

    protected static $relationships = [];

    public static function getRelationship($id)

    {

        if (!isset(self::$relationships[$id])) {

            self::$relationships[$id] = self::find($id);

        }

        return self::$relationships[$id];

    }

    // ... More code here

}

```
It works for me but i am only learning and not making anything in a production app, I hope it will be fixed soon in the master branch

joshuadegier:
@aruberutochan until the fix is merged (I still get the error on a fresh install I did today), you can also create a template under /resources/views/voyager/formfields/relationship.blade.php to override the template that comes with Voyager by default (in stead of changing files in de vendor directory).

leonardyhuang:
I had the same issue, and my workaround is to use this trait below in the models with relationship:

```php
<?php

namespace App\Traits;

trait HasVoyagerRelationship
{
    public static function getRelationship($id)
    {
        return self::find($id);
    }
}
```

vitormicillo:
hi guys, how to solve this problem?

5alvatore:
[...]

PR #2433, opened 2018-01-04
Bugfix #2431 BelongsTo exception

JapSeyz:
This uses the old FIND method as a fallback, and exposes a trait that users can add to their own models

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -11,7 +11,12 @@
 				@php 
 					$relationshipData = (isset($data)) ? $data : $dataTypeContent;
 					$model = app($options->model);
-					$query = $model::getRelationship($relationshipData->{$options->column});
+
+                    if(method_exists($model, 'getRelationship')){
+                        $query = $model::getRelationship($relationshipData->{$options->column});
+                    } else {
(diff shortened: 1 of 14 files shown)


fletch3555:
Duplicate of #2432.  Not sure I like adding the fallback that way, but it _may_ need to be done to be more user-friendly.  You also only add this for `Role` and `User` models.  Mine added it for the rest as well.

JapSeyz:
Yeah, that fallback seems to be the best thing for now.

I can add it for the rest too. I just don't think they have relationships?

JapSeyz:
Added the trait to the rest of the models

tnylea:
Thanks for this. I was actually going to add a PR for this as well.

@JapSeyz Can you fix the StyleCI recommendations and I can merge this in.

Thanks.

JapSeyz:
@tnylea Should be good to go

fletch3555:
@tnylea, my preference is leaning toward my PR (see link above).  All it needs is the fallback, which I'm not really sure I like (it's more user-friendly, but somewhat hides the caching benefit. I think I'd rather force users to add the trait rather than just hide the error).  Also, this one has the trait added to a few models unnecessarily (e.g. DataRow and DataType)

PR #2436, opened 2018-01-05
Fix #2392 many-to-many relationships

### LINKED ISSUE #2392: Define many-to-many relationships

P9Q:
The StyleCI analysis failure is not because of my edit at all, I double checked the MySQL syntax, fortuantely 
`$options->table.'.'.$options->key` will be translated into:
```sql 
select `tags`.`id` from
```
Not:
```sql 
select `tags.id` from
```

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -139,7 +139,7 @@
 				<select class="form-control select2" name="{{ $relationshipField }}[]" multiple>
 					
 			            @php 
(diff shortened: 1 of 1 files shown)


marktopper (approved):
Thanks for the fix

PR #2437, opened 2018-01-05
MenuItemPolicy accesses key instead of name

niels-numbers:
MenuItemPolicy accesses now the key column instead of the name column of permissions table. The latter produced an error:

image

### DIFF
--- a/src/Policies/MenuItemPolicy.php
+++ b/src/Policies/MenuItemPolicy.php
@@ -37,7 +37,7 @@ class MenuItemPolicy extends BasePolicy
         }
 
         // If permission doesn't exist, we can't check it!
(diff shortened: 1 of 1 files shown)


fletch3555:
Can you change it to `->whereKey('browse_'.$slug)` instead of `->where('key', '=', 'browse_'.$slug)`? I think it's more succinct/

niels-numbers:
Your right that is much better!

fletch3555:
Welp.... Turns out `whereKey` is an actual method on the models (referencing `$primaryKey`), so it's causing all MenuItems to display (since there are obviously no permissions with an ID of "browse_[x]").  100% my fault.  Fixing right now.

PR #2452, opened 2018-01-10
Fix #2347 Patch slugify to only aplha numeric string

### LINKED ISSUE #2347: Slugify allows non URL characters.
- Laravel Version: 5.5.25
- Voyager Version: 1.0.9
- PHP Version:7.1

### Description:
When entering a title in a post or page, the slugify function allows non URL characters. Things like ' or ! which can break the URL. Trying to add the "remove" option in the BREAD for the post creates invalid JSON error and won't save.

This is sort of a big deal, I don't want to have to override all the CRUD pages and I don't want users to have to create their own slugs manually.

### Steps To Reproduce:
Create a post, type isn't! into the title.

### LINKED ISSUE #2493: Slug has trailing hyphen when title ends in hyphen
- Laravel Version: 5.5.x
- Voyager Version: 1.0.x
- PHP Version: 7.x
- Database Driver & Version: Not relevant

### Description:
When using slugify to turn a title into a slug, if the title has punctuation at the end they slug will have a trailing `-`

#### Example: 
**Title:** *How much do you know about cryptocurrency?*
**Slug:** *how-much-do-you-know-about-cryptocurrency-*

### Steps To Reproduce:
See above.

handiwijoyo:
This will remove unwanted character like *, !, #, etc

Fixes #2347 
Fixes #2493


resources/assets/js/slugify.js
```diff
@@ -103,7 +103,8 @@
                     .replace(/^\s+|\s+$/g, '')      // Trim
                     .replace(/[^-\u0600-۾\w\d\$\*\(\)\'\!\_]/g, _sep)   // Remove invalid chars
```
fletch3555: '*' and '!' are exlicitly allowed here, just a couple lines up from your changes.  Wouldn't it be better to just remove them from there instead?

handiwijoyo: I don't exactly understand this line, can we remove this line? It's been there before.

fletch3555: The line converts all characters that are NOT a hyphen, between the range of [unicode(0x0600)](http://www.fileformat.info/info/unicode/char/0600/index.htm) and `۾` (unicode(0x0645)), a [word character](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) (`A-Z`, `a-z`, `0-9`, `_`), a digit (`0-9`, technically redundant due to `\w`), `$`, `*`, `(` or `)`, `'`, `!`, or `_` (again, redundant due to `\w`), and converts them to `_sep`

resources/assets/js/slugify.js
```diff
@@ -101,10 +101,9 @@
 
                 str = _slug
                     .replace(/^\s+|\s+$/g, '')      // Trim
-                    .replace(/[^-\u0600-۾\w\d\$\*\(\)\'\!\_]/g, _sep)   // Remove invalid chars
+                    .replace(/[^-\u0600-۾\w]/g, _sep)   // Remove invalid chars
```
handiwijoyo, line 104: Is this ok now? @fletch3555

fletch3555: At quick glance, it seems like it would meet the requirements.  We'll have to do a bit of testing

emptynick:
Some thoughts from my side:
In line 104 you only have an alpha-numeric-string + unwanted chars, after replacing the char-maps, 
why dont we just remove *everything* except `a-z0-9` and spaces?
Line 106 uses hardcoded dashes, we could use something like: 
`.replace(new RegExp('\\'+_sep+'\\'+_sep+'+', 'g'), _sep)`
And in the end, we should remove all seperators from the beginning and end of the string:  
`/(^\-+|\-+$/g`

fletch3555:
The char-maps don't remove symbols, whitespace, etc.  It only converts accented and non-latin characters to latin equivalents (e.g. greek letter `alpha` becomes `a`).  The subsequent regex is there to remove symbols and other invalid chars.

But I agree, line 106 should be updated, and we should definitely be trimming off separators from beginning and end

emptynick:
So after replacing known special-chars to latins, there are only `a-z`, `0-9`, spaces, and unwanted chars.   
So we could just remove everything else, except aplhanumerics and white-spaces    
Instead of
`.replace(/[^-\u0600-۾\w]/g, _sep)   // Remove invalid chars`
it could be  
```
//Replace non-latin with latins
.replace(/[^a-z\s0-9]/g, ' ')
.replace(/\s+/g, _sep)
.replace(/^\-+|\-+$/g, '');
```

fletch3555:
Okay, I see what you're saying.  You still need the "remove duplicate separators" one, but you can use that instead of the "remove whitespace" one, since that can be handled by removing `\s` from the one above it

emptynick:
This should be fine:
```
.replace(/[^a-z0-9]/g, _sep) //Replace everything except alphanumerics with _sep
.replace(new RegExp('\\'+_sep+'\\'+_sep+'+', 'g'), _sep) //Remove multiple _sep with a single _sep
.replace(new RegExp('^\\'+_sep+'+|\\'+_sep+'+$', 'g'), ''); //Remove _sep from start and end of string
```

fletch3555:
Agreed @emptynick.  @handiwijoyo, would you like to update your PR to match @emptynick's comment, or just open a new PR?

akazorg:
Hey guys, you can't remove that line or we lose Arab support.

```
.replace(/[^-\u0600-۾\w]/g, _sep) // Remove invalid chars
```

fletch3555:
Are those characters valid for URLs?

akazorg:
They do. I added that for fixing a bug, at the time.

emptynick:
Closing in favor of #2830

PR #2454, opened 2018-01-10
Refactores the nasty switch statement. 

morloderex:
@marktopper


morloderex:
A rebase might have been wrong here..

fletch3555:
Fixed your branch.  You had extra commits since `1.x-dev` was out of date with `1.0`.

morloderex:
Hehe how should i know that.. :P

fletch3555:
Not saying you should have, nor that it was your fault.  Just informing you of the change, that's all.

emptynick:
Is there a reason that this is not implemented in the Formfield-class itself?

src/Http/Controllers/ContextTypes/BaseType.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace TCG\Voyager\Http\Controllers\ContentTypes;
```
fletch3555: You have a typo in all of your new classes.  You used the namespace `ContentTypes`, but named the folder `ContextTypes`, so the classes aren't being found.  This is causing tests to fail.  Please fix it.

fletch3555 (changes requested):

morloderex:
whoops. Sorry @fletch3555 Fixed.

@emptynick This could also be done in that way, but i was trying to keep it the way it was as much as possible. But you could agrue that it should go in there.

src/Http/Controllers/ContextTypes/BaseType.php
morloderex: This has been fixed

marktopper (approved):
Great refactor

marktopper:
NOTE: If I remember right, then StyleCI fixes are automatically fixed upon merge.

fletch3555 (approved):

fletch3555:
Correct @marktopper.  I'm hesitant to merge this at the moment, because I know this whole area of the app is fairly fragile and needs a major refactor (in the works, I promise).  But I'll merge it in.  I don't see this change causing any issues unless there's a significant copy/paste error we didn't see.

PR #2455, opened 2018-01-10
 Address .login-sidebar properly for CSS overrule

sebastianmulders:
Currently, when the primary color in the config is changed, it is not reflected on the border on the .login-sidebar div. The reason for this is that it isn't properly address and thus can't be overruled. This PR proposes addressing the sidebar so that it will overrule the default setting and color.


fletch3555:
You've simply made the selector MORE specific.  The existing selector is perfectly fine for applying those styles, but may be too generic and be overridden by a more specific selector elsewhere.  I don't have a testbench environment accessible at the moment, so I can't verify this.

In either case, I doubt the `body` tag portion of the selector is necessary.  I think it's ugly and provides no benefit.

sebastianmulders:
Fair enough, maybe I just used the wrong words but, indeed, that's exactly what I did and intended to do. I have tested this, and with this MORE specific selector, the border-color now correctly reflects the value set in the config. Obviously, before, it did not.

As for the `body` part, it might be too much. It's just that, in the current situation on line 25 there already is a similar selector (`body.login .form-group-default.focused`). I tried to keep it consistent. I can do another PR without the `body` if that's what is preferred.

fredbradley (approved):
I don't see anything wrong with this. I was about to submit a pull request for the very same thing.

emptynick (approved):
Works as expected

fletch3555 (approved):

PR #2459, opened 2018-01-10
Feature #1515 Save and create new record

### LINKED ISSUE #1515: Add a button to save content and create a new one.
- Laravel Version: 5.4.28
- Voyager Version: 0.11.14
- PHP Version: 7.1
- Database Driver & Version: MySQL 5.5

### Description:
First of all, I would like to thank you all for this great work you did !

Here is my proposition: when creating/editing new BREAD content, it would be great to have 3 buttons to save : 

 - Save : save and go back to the browse list
 - Apply : save and stay on the content
 - Save and Add new : save and create another content

That would be a great time saver.

### Steps To Reproduce:

Add a new BREAD content, there is only a "Save" button which will show you the create content.

Have a nice day !

handiwijoyo:
This PR will add feature to have "Save & Create New" button on edit-add bread view.

image

Closes #1515


akazorg:
Hey, nice job. I would like to share some ideas.

#### Use config file
I think we should use the config to save the default action, or users will complain about.

#### 3 buttons
Let's make this with all variations?
- Save and return
- Save and stay
- Save and new

#### Dropdown margin
- can you remove the margin under the button? (I thought the default BS style was not including any margin)

Happy coding! :)

handiwijoyo:
I don’t think we need to add config for this and prevent config too many options. I will leave the default behavior to redirect user to browse page after save. 

I agree that we will need one more button “Save & Continue Edit”. 

I will add more commit then.

akazorg:
I think the config is important, because some people wan't to use `Save and Insert` as default. Without a config, they will require to extend and maintain views, and it doesn't sound right.

Victor-emil:
I was actually on my way to implement this on my own, but good to see that I am not the only one who wants additional buttons. In my case I need a "Save" and a "Save and close", which I guess is just the same as "Save" and "Save and continue". Just a different choice of words.

fletch3555:
I'm inclined to prefer just Save and Save&Close buttons.  Adding the config feels like it'll get bloated quickly (someone out there is going to want a save&do-a-thing&do-3-other-things&go-to-med-school&become-a-doctor&etc...).  Save&Close is the safest as it puts you back where you started, giving you just one more click to create another new entry.

akazorg:
Well, the first time I need the default behavior to be `Save and Create new`, I will definitely `cry` for this config parameter. :D

If you want to add it, here follows a draft:
```
    /*
    |--------------------------------------------------------------------------
    | Default Save Behaviour
    |--------------------------------------------------------------------------
    |
    | This option controls the default "Save" button that will be used when
    | submitting a BREAD form. By default, on submit, we redirect to the index,
    | but you may specify another action.
    |
    | Supported: "on_save_index", "on_save_create", "on_save_stay"
    |
    */

    'on_save_action' => 'on_save_index',
```

fletch3555:
The problem I see is people wanting different actions for different pages, which this doesn't support.  It's easier to just provide one option and take a firm stand on the issue.

handiwijoyo:
So the default "Save" when user add / edit the data, it will stay on Edit page. Then "Save & Close" will redirect them back to browse page.

image

akazorg:
@handiwijoyo will redirect to the index, like it's working currently. Don't change that, please.

akazorg:
I recognize the need of the config. If you search on issues about this subject, you will get some interesting reactions. In most cases, there's a valid point for changing the default button.

fletch3555 (dismissed):

marktopper:
Just fixed the merge conflicts by compiling the combined assets locally and committing that as a resolve.

Victor-emil:
What is the current status on this one?

sleepless:
@fletch3555 why was this closed? Is there a solution now?

PR #2461, opened 2018-01-10
Feature #1261 Add Disabled form field type

### LINKED ISSUE #1261: Add type "disabled" in field types
- Laravel Version: 5.4.x
- Voyager Version: 0.10.11
- PHP Version: 7
- Database Driver & Version: 5

### Description:
Would be great to have the field type as disable, for the cases we just want to show the value, but not allow to edit it. Or even to show it as text (not textfield)
Currentlty  we need to validate IF field->name =='something" then apply the disable property, or with Jquery and look for the field and apply the property

handiwijoyo:
Now you can have a disabled input type.

`On Bread Settings:`

image

`On Edit View:`

image

Closes #1261

### DIFF
--- a/resources/views/formfields/text.blade.php
+++ b/resources/views/formfields/text.blade.php
@@ -1,4 +1,5 @@
 <input @if($row->required == 1) required @endif type="text"  data-name="{{ $row->display_name }}"  class="form-control" name="{{ $row->field }}"
         placeholder="{{ isset($options->placeholder)? old($row->field, $options->placeholder): $row->display_name }}"
        {!! isBreadSlugAutoGenerator($options) !!}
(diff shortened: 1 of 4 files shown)


fletch3555:
So... is this a way of making a field not show up? or making it show up but appear disabled (grayed out)?  If it's the former, I don't see any use for this, since you can just uncheck the BREAD checkboxes.  If the latter, then wouldn't you also need to know whether you're displaying a disabled textbox, or a disabled checkbox, etc.?  This seems like a much better option for the JSON config field than a field type.

handiwijoyo:
@fletch3555 You are right. This should be a JSON option rather than a field type. I will submit another PR for this.

PR #2492, opened 2018-01-15
Fix user-role relationship

giorgioma:
The relationship User Role was broken, it was using a pivot table `user_roles` that did not exist.
I updated the code to not use the pivot table.

I took the chance to add Voyager::modelClass in the relationship inside `./src/Model/Permission.php`

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -3,8 +3,9 @@
 namespace TCG\Voyager\Models;
 
 use Illuminate\Database\Eloquent\Model;
-use TCG\Voyager\Traits\HasRelationships;
 use TCG\Voyager\Facades\Voyager;
+use TCG\Voyager\Traits\HasRelationships;
+
 
 class Permission extends Model
 {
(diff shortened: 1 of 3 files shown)


src/Traits/VoyagerUser.php
```diff
@@ -13,7 +13,7 @@
 {
     public function role()
     {
-        return $this->belongsTo(Voyager::modelClass('Role'));
```
fletch3555: We should leave this alone.  It functions identically since you changed it to the reciprocal relation, but it's cleaner before.

src/Models/Role.php
```diff
@@ -14,7 +14,7 @@ class Role extends Model
 
     public function users()
     {
-        return $this->belongsToMany(Voyager::modelClass('User'), 'user_roles');
+        return $this->belongsTo(Voyager::modelClass('User'), 'id', 'role_id');
```
fletch3555: This should be a `hasMany()`, then you won't need the `id` and `role_id` params

src/Models/Permission.php
```diff
@@ -13,7 +14,7 @@ class Permission extends Model
 
     public function roles()
     {
-        return $this->hasMany(Role::class);
+        return $this->hasMany(Voyager::modelClass('Role'));
```
fletch3555, line 18: good find, thank you.

fletch3555 (changes requested):

giorgioma:
Changed as requested, it does make more sense

fletch3555 (approved):

PR #2496, opened 2018-01-16
Cleanup hook

marktopper:
Now that `larapack/voyager-hooks:v1.0.5` adds the menu item and the permission, it is no longer needed here.


composer.json
```diff
@@ -21,7 +21,7 @@
         "larapack/doctrine-support": "~0.1.4",
         "arrilot/laravel-widgets": "^3.7",
         "league/flysystem": "~1.0.41",
-        "larapack/voyager-hooks": "~1.0.2"
+        "larapack/voyager-hooks": "~1.0.5"
```
marktopper, line 24: Version `1.0.5` will handle the menu item and permission seeding.

src/Commands/InstallCommand.php
```diff
@@ -116,6 +113,9 @@ public function handle(Filesystem $filesystem)
             $this->seed('VoyagerDummyDatabaseSeeder');
         }
 
+        $this->info('Setting up the hooks');
+        $this->call('hook:setup');
+
```
marktopper, line 118: Moved to execute the `hook:setup` command until after Voyager has run its own seeders.
Since if the `Admin` menu does not exists, the menu item `Hooks` will not be added.

fletch3555 (approved):

PR #2510, opened 2018-01-18
Do not respect global scopes in the bread

iSWORD:
I believe an admin panel should not obey global scopes. This commit will make the VoyagerBreadController discard the filters imposed by global scopes using "withoutGlobalScopes" and "newQueryWithoutScopes".


iSWORD:
bump.. ?

emptynick:
This is definitely a breaking change as some people may be using global scopes and end up having to rewrite everything to Policies.
At least for `1.0.x`

iSWORD:
Agreed. I should probably add it as an option in the BREAD builder instead.

emptynick:
Make it either a config var, or take a look at #2827
It introduces a new `details` column to Data-Types with which you _could_ implement a var `respect_global_scopes` or so

PR #2511, opened 2018-01-19
Fix Redirecting to intended URL after Login

josvar:
This is done by redirecting with guest method instead of redirect with flashed session data.


emptynick:
This is working perfectly fine!

emptynick (approved):

fletch3555 (approved):

PR #2429, opened 2018-01-04
Fix #2381 upsize & quality not applied

handiwijoyo:
fixes #2381
fixes #2250 
fixes #1328

Issue #2381, opened 2017-12-20
the upsize set true, there is not effect


bradapeter:
- Laravel Version: 5.2.26
- Voyager Version: 1.0
- PHP Version:
- Database Driver & Version:

### Description:
when I insert a picture in the post, although I set the the upsize with true. the small image's width less the 1000pixel. according to the guide. the uploaded the picture has no effect to upsize.

so what's going on?

emptynick:
Looking at the code there is no check if `upsize` is true or false.
This should be done [here](https://github.com/the-control-group/voyager/blob/1.0/src/Http/Controllers/Controller.php#L226)

PR #2519, opened 2018-01-20
Resolve images resize , quality and upsize .

SaadEddinHassania:
Resolve images resize , quality and upsize .

I used resize before, but it doesn't work , and the problem was in the condition . I resolve it and enable quality and upsize .

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -42,8 +42,8 @@ abstract class Controller extends BaseController
          * Prepare Translations and Transform data
          */
         $translations = is_bread_translatable($data)
-                        ? $data->prepareTranslations($request)
-                        : [];
+            ? $data->prepareTranslations($request)
+            : [];
 
         foreach ($rows as $row) {
             $options = json_decode($row->details);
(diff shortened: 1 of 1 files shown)


emptynick:
(Possible) duplicate of https://github.com/the-control-group/voyager/pull/2429

SaadEddinHassania:
but my edit for resizing too , this #2381 doesn't resolve resizing

SaadEddinHassania:
But there is different in resizing condition

src/Http/Controllers/Controller.php
```diff
@@ -469,6 +448,34 @@ private function is_animated_gif($filename)
         return $frames > 1;
     }
 
+    public function validateBread($request, $data)
```
fletch3555: Why did you move this method?  It does nothing except complicate the diff.  Please either provide a good reason or move it back.

fletch3555 (changes requested):

src/Http/Controllers/Controller.php
SaadEddinHassania: I did move it back , but in an amend commit by wrong

fletch3555:
#2381 is different, but definitely the same as #2538.  Closing this one as duplicate since the checks all pass on #2538

PR #2522, opened 2018-01-20
Edit Roles system to depend on role's order "The least is the strongest"

AsemAlalami:
Edit Roles system to depend on role's order "The least is the strongest", order 1 is strongest (for admin).
sorry for many pull request but I wanted to approved from "styleci".
thanks.


fletch3555:
FYI, PRs will update automatically if you push new changes to the same branch. There's no need to close the PR and open a new one just to add new commits.

src/Http/Controllers/VoyagerRoleController.php
```diff
@@ -7,15 +7,115 @@
...
+            $this->removeRelationshipField($dataType, 'browse');
+
+            if ($search->value && $search->key && $search->filter) {
+                $search_filter = ($search->filter == 'equals') ? '=' : 'LIKE';
+                $search_value = ($search->filter == 'equals') ? $search->value : '%'.$search->value.'%';
+                $query->where($search->key, $search_filter, $search_value);
+            }
+
+            // Browse roles that greater than his role only except admin( order = 1)
+            $user_role_order = Auth::user()->role->order;
```
marktopper, line 58: Please include `Illuminate\Support\Facades\Auth` in other to use `Auth`.

marktopper (dismissed):

AsemAlalami:
@marktopper 
I fixed the exceptions that can occur.
thanks.

marktopper:
Okay, just to go through the PR.
Then what this does is that it adds the option to set a hierarchy order for roles.
This will allow a user to show & edit roles that has a higher order than the user's lowest ordered role.

Am I right?

AsemAlalami:
Yes, you're right.
and the user will add users with role higher order than the user's lowest ordered role.

marktopper:
Any thoughts on this @fletch3555?

AsemAlalami:
note: if the user has **add role** permission, he can set permissions only that his owned it.

src/Http/Controllers/VoyagerRoleController.php
```diff
@@ -57,6 +160,9 @@ public function store(Request $request)
         }
 
         if (!$request->ajax()) {
+            if ($request->input('order') < Auth::user()->role->order) {
+                return back()->with(['errors' => collect(['Order must be grater than your order - your order is '.Auth::user()->role->order])]);
```
fletch3555: Couple things here....

1. You misspelled `greater`
2. Error messages need to be translated, and shouldn't be hard-coded, especially not in the controller.
3. I don't really like referring to it as "your order".  Makes it seem like you're purchasing something.

src/Policies/RolePolicy.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace TCG\Voyager\Policies;
+
+use TCG\Voyager\Contracts\User;
+
+class RolePolicy extends BasePolicy
+{
+    public function read(User $user, $model)
+    {
+        $role_greater = $user->role->order == 1 || $model->order > $user->role->order;
```
fletch3555, line 11: What's the guarantee that `admin` has order=1? I don't like magic numbers like this.

fletch3555 (changes requested):
I'm not sure how I feel about including this in the core... I completely understand that it's a desired feature for Voyager, but I'm not sure this is the best approach.  I think it makes a few too many assumptions, where we're trying to make Voyager be as non-intrusive as possible.

src/Http/Controllers/VoyagerRoleController.php
AsemAlalami: Thanks.
I solved this notes and change the term from "order" to "rank"

src/Policies/RolePolicy.php
AsemAlalami: @fletch3555 
In the fact I do that to `Developer` role.
I created many projects depends on `voyager` in every time I need `developer` role to prevent admin to edit in `BREAD` tables

fletch3555:
I'm leaning toward declining this.  A single field that determines the hierarchy of roles seems like a great way to get hacked (if they can get into the database or otherwise change that one value, they'll be able to get into everything).  Also, it doesn't ACTUALLY make them hierarchical, since permissions aren't inherited at all, something I expect many of our users would expect after seeing that they can change the "rank" of roles in this way.  I'm thinking a truly hierarchical (parent-child relationship with itself) method is required, complete with recursive permission checking.

Finally, this will most likely conflict with @akazorg's "multi role" stuff in the 1.1 stream... This doesn't handle the case where a user has multiple roles

PR #2525, opened 2018-01-21
Replaced rtrim() with str_replace_last() to get the name of the file

ghost:
The use of rtrim($image, '.'.$ext); would trim characters of the $image string before the '.' if those characters where contained in $ext.


ghost:
@fletch3555 I hope now I did it correctly
Edit: Well, for some reason, although it is the same change in code this time it failed the unit testing. I have probably done something wrong...

fletch3555:
No you're good.  I'm guessing Laravel (or some other dependency) changed something since it's only failing on builds using "recent" versions of dependencies (builds that don't have `--prefer-lowest`).  This is affecting multiple builds, not just your PR, so I know it's not something you did

fletch3555 (approved):

PR #2534, opened 2018-01-23
Allow to override the users table to log in [multi auth setup]

DxCat:
This PR will allow using other table instead of being tightly-coupled to the users table.
It would be very useful for multi-auth setup where the admins and the users are separated on different table.

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -1,6 +1,29 @@
 <?php
 
 return [
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Voyager auth config
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify voyager auth configs that you want to use to login
+    | to the voyager admin.
+    |
+    */
+
+    'auth' => [
+        /*
+         * Set the table name that you want to authenticate against
+         */
+        'table' => 'user',
+        
+        /*
+         * Set the guard you wish to use. Use null for default laravel guard
+         */
+        'guard' => 'null',
+    ],
+    
     /*
     |--------------------------------------------------------------------------
     | User config
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -21,7 +21,7 @@ return [
         /*
          * Set the guard you wish to use. Use null for default laravel guard
          */
-        'guard' => 'null',
+        'guard' => null,
     ],
 
     /*
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -1,7 +1,7 @@
 <?php
 
 return [
-    
+
     /*
     |--------------------------------------------------------------------------
     | Voyager auth config
@@ -17,13 +17,13 @@ return [
          * Set the table name that you want to authenticate against
          */
         'table' => 'user',
-        
+
         /*
          * Set the guard you wish to use. Use null for default laravel guard
          */
         'guard' => 'null',
     ],
-    
+
     /*
     |--------------------------------------------------------------------------
     | User config
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -16,7 +16,7 @@ return [
         /*
          * Set the table name that you want to authenticate against
          */
-        'table' => 'user',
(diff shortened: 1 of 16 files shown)


fletch3555:
I'm not sure how I feel about this.  On one hand, I'm perfectly fine with enabling users to change the auth guard.  On the other, having `::guard(config(....))` on EVERY call is frustratingly cumbersome...

I'll have to think on this

DxCat:
@fletch3555 that's quite true. Then how about we have a custom helper like `voyager_auth()` that will behind the scene call `Auth::guard(config('voyager.auth.guard)`. Will that be better?

fletch3555:
Still not really fond of having that everywhere.  There should be a way to tell Laravel's Auth framework to always use a specific guard.  Docs are here: https://laravel.com/docs/5.5/authentication

I know you can do it via middleware, so maybe that's an option.  Dynamically set the guard (by some config value) in the [login controller](https://github.com/the-control-group/voyager/blob/1.0/src/Http/Controllers/VoyagerAuthController.php) and add `auth:[guardname]` to the [route file](https://github.com/the-control-group/voyager/blob/1.0/routes/voyager.php).  Only possible issue is how Voyager loads routes (pulled in from under `routes/web.php`, which may add a different `auth` middleware, which may conflict in some way)

plutojw:
I used this version in my project. The default Laravel Auth will be affected. 
Below is my voyager setup.

`    'auth' => [
        'table' => 'admins',
        'guard' => 'admin',
    ],`

All Auth::user will use table Admins not Users.
Can I set only Voyager use this config?
Thanks.

niels-numbers:
According to http://mattallan.org/posts/setting-the-guard-per-route-in-laravel/ it should be possible since 5.3 to simply write

```
Route::group(['as' => 'voyager.', 'guard' => config('voyager.auth.guard')], function () {
```
to specify which guard `Auth()` is using.
But then I assume that users also would need to add this in there own web.php file:

```
Route::group(['prefix' => 'admin', 'guard' => config('voyager.auth.guard')], function () {
    // ...
});
```

plutojw:
@iwasherefirst2  Thanks your information. 

After the setup, the default Auth will be set to 'Admin', not 'Web'.
I tried the link solution.
`Route::group(['guard' => 'admin'], function () {
    Route::get('/admin/test', function () {
        return Auth::user()->toJson();
    });
});`

`Route::group(['guard' => 'web'], function () {
    Route::get('/web/test', function () {
        return Auth::user()->toJson();
    });
});`

I login Admin user via Voyager login page, above 2 functions return same user.

I can solve the issue via below code to get default auth user like the helper.php
`Auth::guard('web')->user()`

If possible, I think set Auth::user() is represent Auth::guard('web')->user(), that will be better.
Thanks.

plutojw:
I suggest to remove below change in VoyagerServiceProvider.php
`Auth::shouldUse(config('voyager.auth.guard'));`

Because above code will change the default Auth guard as config('voyager.auth.guard').
It is not necessary, because we are using helper method voyager_auth() to get the voyager auth.

`Route::get('/admin/test2', function () { return voyager_auth()->user()->toJson(); }); `

But Voyager will check default Auth in other method, such as BREAD method.

So I will rollback to use official version and setup Auth.php to change the default Auth guard.
 
`    'defaults' => [
        'guard' => 'your defined guard',
        'passwords' => 'your defined guard',
    ],`

niels-numbers:
@plutojw sorry for the confusion, but my last post was not directed at you (should have mentioned it, whoups)! I just posted an alternative way of how to change this plugin, to avoid having `voyager_auth()` everywhere.. But for your question: Did you create a specific admin model & table? This version only allows you to use a different guard then the default one in Voyager. You have to set up the other things by yourself. See https://github.com/the-control-group/voyager/issues/2568 , where I explained the workaround for the official version.

DxCat:
Closing as using a route based guard might be a better option to go for this.

niels-numbers:
@askaoru do you mean you prefer the workaround at #2568, or do you want a different pull request for that?

Issue #2342, opened 2017-12-12
Relationship Issue : displaying no results


sharhan016:
- Laravel Version: latest
- Voyager Version: latest
- PHP Version: 7.1
- Database Driver & Version: Apache/2.4.29 (Win32) OpenSSL/1.0.2l PHP/7.1.11

### Description:
I created two tables naming **items** and **product_types** ... I added two inputs in the product_type and created a relationship for items with product_types..
when I try to add an item I am able to select the product_type with a drop down menu and I'm able to select..but after saving the column product_type is displaying no results
**My item bread property:**
![item bread](https://user-images.githubusercontent.com/25862854/33872626-a04dfc46-df3d-11e7-9be6-f1cc5137d6cf.PNG)
**When adding an item:** 
![item edit](https://user-images.githubusercontent.com/25862854/33872653-beaa4eec-df3d-11e7-9483-83b76bfa9d45.PNG)
**after adding when I view the item:**
![item ew view](https://user-images.githubusercontent.com/25862854/33872692-ea535fde-df3d-11e7-9270-cbaf71d10ab6.PNG)

handiwijoyo:
Can you try to use product_type_id instead?

fletch3555:
Please provide the actual version number for laravel and Voyager.  Saying "latest" is not helpful if this issue were to remain open for weeks/months

Overfinch:
I had the same problem, make sure that you activated the checkboxes (Visibility) in field "product_type".

notisset:
Solution provided by @Overfinch works for me. Also I had to change the column's "Input Type" to "Number".

v1.0.14

emptynick:
This also seems to be fixed in https://github.com/the-control-group/voyager/pull/2837 (1.x-dev)
The default selected type is now text and not checkbox.
If it isn't fixed (in future 1.1) please let us know or open a new issue!

suuuperman:
@Overfinch  thankyou it works to me also.

PR #2535, opened 2018-01-23
Update Controller.php for fix no result issue #2342

MamunHoque:
Using BelongsTo relationship if filed type is check-box its  save default value like 1. so we must need to change check-box to others type like number/text or allow relationship filed. 
Solved : Relationship Issue : displaying no results #2342


fletch3555:
I'm not sure this is the correct fix.  This seems 100% like a user misconfiguration error.  The fix from our end should be to make it more obvious what the misconfiguration is.

fletch3555:
I think we need a different solution for this.

PR #2550, opened 2018-01-25
User model call abstraction

sdrobov:
Small patch to abstract from voyager's user model and let user to redefine it.

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -63,7 +63,8 @@ class Menu extends Model
             $user_permissions = null;
 
             if (!Auth::guest()) {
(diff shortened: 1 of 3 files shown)


fletch3555:
This is already possible in Voyager ~and documented~ (apologies, I thought it was, but can't seem to find it anywhere.  That's what I get for speaking before looking).  Just add the following to your AppServiceProvider: `Voyager::useModel('User', \App\User::class);` (and change the model namespace as appropriate for your app).  There are also a number of issues that reference this facade method: https://github.com/the-control-group/voyager/search?q=useModel&type=Issues

Issue #1886, opened 2017-09-26
Can't access to the admin page in voyager 1.0


ngomtanouba:
- Laravel Version: 5.5.13
- Voyager Version: 1.0
- PHP Version:7.0.10
- Database Driver & Version:MySql 5.7.14

### Description:
Hello!
After `composer  require tcg/voyager`  i run `php artisan voyager:install --with-dummy` .

All things are correct but at the end of my installation i got this issue:

`Attempting to set Voyager User model as parent to App\User
Dumping the autoloaded files and reloading all new files

  [Symfony\Component\Process\Exception\ProcessTimedOutException]
  The process "composer dump-autoload" exceeded the timeout of 60 seconds.`

And then can't to my admin page.
Need your help! Thanks

ngomtanouba:
@fletch3555 I delete wich one the public vendor directory or the '/vendor/tcg ? I've ran `composer dump-autoload` then  `php artisan voyager:install` but return `[Symfony\Component\Process\Exception\ProcessTimedOutException]
  The process "composer dump-autoload" exceeded the timeout of 60 ` again

Maghraby:
@fletch3555 the error because can't load jquery.
![Uploading Screen Shot 2017-09-27 at 8.12.34 PM.png…]()

Maghraby:
@fletch3555 this's because can't load jquery
<img width="1436" alt="screen shot 2017-09-27 at 8 12 34 pm" src="https://user-images.githubusercontent.com/7889084/30930556-5242d668-a3c2-11e7-97ef-ce1dc170186e.png">

ngomtanouba:
What can we do in this case? @Maghraby

fletch3555:
@Maghraby, the presence or absence of jquery has absolutely nothing to do with this issue.

@BarthFox, I meant delete `/vendor/tcg/voyager`, not `/public/vendor`.

Maghraby:
@fletch3555 jquery can't load any data
<img width="1440" alt="screen shot 2017-09-30 at 6 38 40 pm" src="https://user-images.githubusercontent.com/7889084/31047649-b07271aa-a60e-11e7-9635-33481266c26f.png">

ngomtanouba:
Hi @fletch3555  Even i delete the vendor folder it returns the same error. i installed the new version of laravel with the new version of voyager but nothing changes

yakuter:
I have the same problem and still couldn't find the solution.

[...]

PR #2551, opened 2018-01-25
Fixing timeout issue during installation

waptik:
Setting timeout to null to prevent installation from stopping at a certain point in time.
Using solution by @mpenna according to issue #1886


fletch3555:
Please don't close PRs just to add commits and open a new one.  PRs are a live diff between two branches, and as such will update automatically if you push new commits to it.  Also, I noticed you changed the base branch (where you're merging into) from 1.x-dev to 1.0.  This can be done by editing the existing PR.

PR #2557, opened 2018-01-27
Add translateable strings to database seeders

Dartui:


publishable/database/seeds/SettingsTableSeeder.php
```diff
@@ -13,120 +13,120 @@ public function run()
...
             $setting->fill([
-                'display_name' => 'Site Title',
-                'value'        => 'Site Title',
+                'display_name' => __('voyager.seeders.settings.site.title'),
+                'value'        => __('voyager.seeders.settings.site.title'),
                 'details'      => '',
                 'type'         => 'text',
                 'order'        => 1,
-                'group'        => 'Site',
+                'group'        => __('voyager.seeders.settings.site.group_name'),
```
fletch3555: Group should not be translated.  Group is used as part of the "key" when referencing the setting.

Dartui: https://github.com/the-control-group/voyager/blob/1.0/src/Http/Controllers/VoyagerSettingsController.php#L19-L29

What about this?

emptynick: Those are the settings that dont have a key.
They will be displayed, and only that, in a group named `General` or its translation.
 Imagine the following in german:
```php
$setting = $this->findSetting('site.title');
if (!$setting->exists) {
    $setting->fill([
        'display_name' => __('voyager.seeders.settings.site.title'),
        'value'        => __('voyager.seeders.settings.site.title'),
        'details'      => '',
        'type'         => 'text',
        'order'        => 1,
        'group'        => 'Seite',
    ])->save();
}
```
This will save the setting `site.title` in the group `Seite`.
You can call `setting('site.title')` **until** you edit/save settings, then you will get a key called `seite.site.title` and you will have to call it like this.

fletch3555 (changes requested):
Please remove the group from the translated string

Dartui:
@fletch3555 done!

fletch3555 (approved):

PR #2558, opened 2018-01-28
Access to cropped image names Voyager::image

giorgigigauri:
There is no way to access images with attribute names defined in BREAD.


fletch3555:
Yes there is: https://voyager.readme.io/docs/thumbnails-url

PR #2570, opened 2018-01-30
Move scripts section after additional_js

ssang:
Scripts specific to a page that go into the scripts section in the master blade file should come after the additional javascript files specified in the config file. Many page related scripts may require intitial files to be loaded in in order to be executed.

### DIFF
--- a/resources/views/master.blade.php
+++ b/resources/views/master.blade.php
@@ -130,11 +130,12 @@ if ((substr(Auth::user()->avatar, 0, 7) == 'http://') || (substr(Auth::user()->a
 
     @endif
 </script>
-@yield('javascript')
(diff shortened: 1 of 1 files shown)


fletch3555:
I don't follow your argument.  Scripts entered through "additional_js" will not have access to jQuery/etc if it's loaded first.  We can't accept this change.

ssang:
What I mean is that if I were to, for example, override the edit-add.blade.php file for a specific resource, I can't initiate a VueJS app through the javascript section in the layout. This is because my components, and even VueJS itself, which I include in my 'additional_js' configuration is currently loaded after the javascript section in the master layout file.

PR #2581, opened 2018-02-01
feature/voyager-config-upload-path

shakalaka:
Took out in voyager configuration upload path


marktopper (dismissed):

marktopper:
Does it work for Windows systems? Or does it need to use `DIRECTORY_SEPARATOR`?

src/Helpers/helperUploadPath.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+if (!function_exists('getUploadPath')) {
```
fletch3555, line 3: I think the helper method is too specific, and would be better served as a service class method or something.

src/Http/Controllers/ContentTypes/File.php
```diff
@@ -43,7 +43,7 @@ public function handle()
      */
     protected function generatePath()
     {
-        return $this->slug.DIRECTORY_SEPARATOR.date('FY').DIRECTORY_SEPARATOR;
+        return getUploadPath($this->slug, config('voyager.storage.path', '%slug%/%date:F%/%date:Y%/'));
```
fletch3555, line 46: As @marktopper said, you changed all of the DIRECTORY_SEPARATOR calls to literal strings, which very likely breaks OS cross-compatibility.

src/Helpers/helperUploadPath.php
```diff
@@ -0,0 +1,38 @@
...
+
+        foreach ($_dirs as $dir) {
+            if ('' === $dir) {
+                continue;
+            }
+
+            if (preg_match('/%(.*?)%/', $dir)) {
+                $dir = str_replace('%', '', $dir);
+
+                if ($dir === 'slug') {
```
fletch3555, line 26: I'm also not really a huge fan of the placeholder replacement.  This if/elseif/etc chain could easily get VERY long and unmanageable.

src/Helpers/helperUploadPath.php
```diff
@@ -0,0 +1,38 @@
...
+     * Return upload path.
+     *
+     * @param string $slug
+     * @param string $path
+     *
+     * @return string
+     */
+    function getUploadPath($slug, $path)
+    {
+        $_dirs = explode('/', $path);
```
fletch3555, line 15: Variable naming is too generic.  Also, $_ is generally reserved for PHP superglobals.

fletch3555 (changes requested):
I think this could be better served by just letting Laravel/Voyager upload the files to a single place.  You can then either move them around manually, or choose a subfolder in the controller of a form using the file/image upload field.

I'm certainly not opposed to letting this be configurable, but I don't think the Voyager config file is the right place.  This would be better configured in the JSON "additional options" field, which allows it to be different for each BREAD type, and keeps the usage localized to the "ContentTypes" classes

marktopper:
Closed due to inactivity

PR #2587, opened 2018-02-02
Test permissions

marktopper:


fletch3555 (approved):
We should probably be a little more verbose with our naming.  In this instance, we have "PermissionTest", but it really just tests the Voyager facade's "can" method and everything downstream from there.  It doesn't test access via `Policies` or anything else like that.

I'll merge this in for now, but just some things to think about when enhancing this test suite

PR #2605, opened 2018-02-04
Custom Breadcrumbs

emptynick:
Allows a view-file or hook to define custom Breadcrumbs if neccessary.
Does not change anything until `@section('breadcrumbs')` is used.


resources/views/dashboard/navbar.blade.php
```diff
@@ -29,6 +29,7 @@
                     @endif
                 @endfor
             </ol>
+            @show
```
marktopper, line 32: Shouldn't this be `@endsection`?

emptynick: No, `@show` behaves like `@yield('breadcrumbs')` and `@endsection`.
`@endsection` will try to write this content to `@yield('breadcrumbs')`
Ref: https://laravel.com/docs/5.5/blade#defining-a-layout

fletch3555 (approved):

PR #2607, opened 2018-02-04
Dont use dev-master

marktopper:
This was breaking all Travis tests.
I will also do one right away for `1.x-dev`


fletch3555:
This change was also merged in as part of #2606.

PR #2610, opened 2018-02-04
Fixed saving belongsto relation field issue

egordm:
I was having an issue where I couldn't save a relation. It would just take id=1 on every save. 
Apparently a select field was using wrong value as name. Further below in the file it is used correctly.

Here is a relation i was using: 

The the name attribute of select was set to user_id which is incorrect.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -26,7 +26,7 @@
 
 			@else
 			
-				<select class="form-control select2" name="{{ $options->column }}">
+				<select class="form-control select2" name="{{ $relationshipField }}">
 					@php 
 						$model = app($options->model);
 	            		$query = $model::all();


IainX4L:
Thanks.  This resolved problem I was having.

vblinden:
This was never an issue for me with the following relationship below.

image

But with the change in this pull request I can still add/edit and delete without a problem.

emptynick:
Is the Datatype of `user_id` in your `Bois` BREAD maybe checkbox?
If so, change it to `Number` and report if that fixes your problem.

egordm:
Yes making `user_id ` a number instead of a checkbox seemed to fix the problem too. 

I have also just noticed that I uploaded the wrong screenshot. Because the name was set to the name of relation instead of the field needed to be set. I apologize for that.

Issue #2578, opened 2018-02-01
Validation errors for rich text editors not showing


apsg:
- Laravel Version: 5.5.33
- Voyager Version: 1.0.14
- PHP Version: 7.1
- Database Driver & Version: mysql

### Description:

For a custom table I created a BREAD with several rich text fields. These fields cannot be null (and in BREAD manager there is info: Required: YES) but still, I can try to save new item to the database (and get mysql error "field xxx cannot be null").

Following [this](https://voyager.readme.io/docs/validation) I've added validation rule for these rich textboxes:

`{
    "validation": {
        "rule": "required",
        "messages": {
            "required": "This field is required"
        }
    }
}`

but it only prevents me from saving new item to the DB when the field is empty - it does not show any kind of error/message. Also, on validation, there are js errors showing in console:

`An invalid form control with name='description' is not focusable.`
`An invalid form control with name='about' is not focusable.`

I suppose it's some kind of issue related to using tinymce rich text editor. How can I make ANY kind of message showing on failed validation?

abacram:
Yes, apparently it's related with validateBread function from Controller.php. Validations are running with $row->display_name instead $row->field. I recommend you to wait for a release with the fix, but if you're hurry for now, i had to override /vendor/tcg/voyager/src/Http/Controllers/Controller.php and make a fix to validateBread in order to get working in my local until the fix is released.

P.D.: i don't recommend you to make any changes into the tcg vendor files, because you may lost them when you try to update or mount your project in another environment/machine.

apsg:
This issue was solved by https://github.com/the-control-group/voyager/pull/2660 in release 1.0.16, so it can be marked as closed.
Thank you!

fletch3555:
Thank you for following up.

Issue #2236, opened 2017-11-15
Voyager 1.0.7 validation errors doesn't show


acalic:
- Laravel Version: 5.4.36
- Voyager Version: 1.0.7
- PHP Version: 5.6.28
- Database Driver & Version:

### Description:
BREAD Validation errors stopped showing after updating voyager. Instead of showing the validation errors of a BREAD in related edit-add page, it shows JS console errors (app.js). I don't remember which was the voyager version number while it still worked, but it was working before. Now the validations just don't work, or they just don't show up right.

Validation example for field "Title" of "Post" BREAD:
```
{
    "validation": {
        "rule": "required|alpha|max:6"
    }
}
```

![voyager-error-js](https://user-images.githubusercontent.com/33701348/32857828-81badc3a-ca49-11e7-82bd-4caf490b943d.PNG)
![voyager-error-net-response](https://user-images.githubusercontent.com/33701348/32857829-81e700ee-ca49-11e7-8388-75c93e88a391.PNG)

### Steps To Reproduce:
Add validation options in BREAD editor, open and try to edit that BREAD, click on "Save" button.

Victor-emil:
I did some research and the bug was introduced in 1.06. The "name" attribute of all formfields has been renamed "data-name". The Javascript to handle this has also been altered to take care of the renaming. However, the edit-add.blade.php file inside posts does not reflect this change. The attribute is still called name, which means that the Javascript tries to retrieve a value from an attribute that does not exist due to the name mismatch.

acalic:
@Victor-emil This happens for all my BREADs, i used the "Post" model just as an example. Thanks for trying to help though.

Victor-emil:
@acalic. I see. I thought that was the only point of failure, but you were totally right. It happens everywhere. I found a bug somewhere else, though. In release 1.06 the ajax calls for validation returns the display_name instead of the actual field name (If an error occurs). The Javascript code will use the display name to find the correct field, however the field name and field/model display name are usually not the same. 
If you go to Edit Bread on the posts table and change the Title to title it will work, but that is not a sustainable solution. That is a temporarily fix, but the Javascript/php code should be changed. If I get the time I will try to do a PR.
Personally I think the code should be using the id of the dom element (id="title" etc.) instead of the name for fetching.

rttolipas:
Thank you for working this out @Victor-emil!

fletch3555:
[...]

PR #2612, opened 2018-02-05
Fixing bugs in Validation.

RodriigoGS:
## This pull request has the purpose of fixing the problem in the validation of the fields.

**Accomplished:**

- Modification in the array keys of the function responsible for validating the fields (Controller @ validateBread).

- Change in the data-name of all fields of the form, so that it is the same as that sent by the validator.

- Add the data-name in the select_dropdown field so that it is found by the validator.


regiszanandrea:
Related to #2205 and #2236.

apsg:
And related to https://github.com/the-control-group/voyager/issues/2578 . Please merge it soon...

regiszanandrea:
@fletch3555 Please, accept this PR.

PR #2629, opened 2018-02-09
Use example.com instead of admin.com for dummy user

vblinden:
We should use the `example.com` domain because this is a reserved domain name for such purposes (https://www.iana.org/domains/reserved). `admin.com` is not reserved.


fletch3555:
Too many of our videos and documentation use the admin.com domain.  We'll change it, but not for v1.1.

PR #2630, opened 2018-02-09
Fix to change illuminate/support dependency version to allow installation with laravel 5.6.1

spencer-aerian:
Visually tested with mysql backend but may be issues with sqlite.
Putting this here for others to run alternative tests.


fletch3555:
This change removes support for laravel 5.5, and doesn't fix any other dependencies.

PR #2631, opened 2018-02-09
Prevent default on .remove-multi-image click

smb043:


fletch3555:
Is there a problem you're trying to solve with this? We can't just blindly accept changes, despite how innocuous they may be.  We need much more information about the problem first, because it may actually be something else causing it, with a known fix.

In any case, it's failing TravisCI, so I can't merge it anyway.

fletch3555:
Scratch my previous comment about Travis.  It was just complaining because it was trying to run against Laravel 5.6, which we didn't support yet.

fletch3555:
@smb043, can you please answer my previous question?

smb043:
Hi @fletch3555. I apologize for the long response. When i click on `.remove-multi-image`, then page scroll to top, instead of stay on current scroll position

fletch3555 (approved):

PR #2658, opened 2018-02-13
Added dutch translation

vblinden:
Hello everyone,

I translated everything to Dutch. Feedback is welcome!


cyrildewit:
Hi @vblinden,

I think that there is already a Dutch translation, if I am right :(. Check the [v.1.0-dev branch](https://github.com/the-control-group/voyager/tree/1.x-dev).

See https://github.com/the-control-group/voyager/blob/1.x-dev/publishable/lang/nl/voyager.php

I made the same mistake xD.

What you could do is double check the translations.

vblinden:
Ah, I only checked the 1.0 branch. No problem! Thanks for the heads up!

cyrildewit:
What a bummer!

PR #2664, opened 2018-02-14
Change redirected path after successful user login

vblinden:
The user can now tell the system where it should redirect to via a config file. It defaults to the standard prefix '/admin'.


akazorg:
Nice feature.

src/Http/Controllers/VoyagerAuthController.php
```diff
@@ -52,6 +52,6 @@ public function postLogin(Request $request)
      */
     public function redirectTo()
     {
-        return route('voyager.dashboard');
+        return config('voyager.user.redirect');
```
fletch3555: Can you give the config call a default value?  Without it, this will blow up if the config isn't there (anyone that upgrades from an older version)

fletch3555 (changes requested):

defaultpage:
@vblinden, Can you set `route('voyager.dashboard')` as default value?
```php
return config('voyager.user.redirect', route('voyager.dashboard'));
```

fletch3555 (approved):

PR #2670, opened 2018-02-15
Fix for #1731

### LINKED ISSUE #1731: Fix Cancele Delete Settings Field

vblinden:
This fixes #1731. The setting id are now correctly updated everytime on the delete form.

Thanks to @davodavodavo3 for the orginal pull request (#1764)!

Closes #1764


resources/views/settings/index.blade.php
```diff
@@ -444,10 +444,12 @@
                 }
             });
 
-            $('.voyager-trash').click(function () {
+            $('.panel-actions .voyager-trash').click(function () {
```
fletch3555, line 449: What's this class for?

resources/views/settings/index.blade.php
```diff
@@ -444,10 +444,12 @@
...
                 var display = $(this).data('display-name') + '/' + $(this).data('display-key');
 
                 $('#delete_setting_title').text(display);
+
+                $('#delete_form')[0].action = '{{ route('voyager.settings.delete', [ 'id' => '__id' ]) }}';
```
fletch3555: Can't we combine these 2 lines and only have to set the action once? Something like:
```
$('#delete_form')[0].action = '{{ route('voyager.settings.delete', [ 'id' => '__id' ]) }}'.replace('__id', $(this).data('id'));
```

If this works, there are other instances of this that should be updated, but that can be a separate PR since it's unrelated to #1731

This also renders the form action useless.  Can you get rid of the unnecessary `route()` call and replace it with `#` or something instead?

fletch3555 (changes requested):

resources/views/settings/index.blade.php
vblinden: This fixes the error when a user clicked the trash icon in the modal the value went: undefined/undefined. This ensures that only the trash icons in the actions are used to get the data-id.

fletch3555 (approved):

PR #2675, opened 2018-02-15
Fixing moving file to folder

regiszanandrea:
Hi,

This PR fix a bug that create sub-folders when you move a file to a another folder.

Case: "I want to move the image to dealerships folder."

**Image and actual folder:**
captura de tela 2018-02-15 as 11 22 24 am

**Destination:**
captura de tela 2018-02-15 as 11 23 45 am

**Explaining actual action**

This action will create a folder called 'planilhas' and inside of this folder will create 'dealerships' folder, then it will move the file to this directory, therefore, the path of image will be '/planilhas/planilhas/dealerships/image.png/'. But, the correct should be '/planilhas/dealerships/image.png/'. As we can see on images below:

**The folder 'planilhas' created:**
captura de tela 2018-02-15 as 11 22 42 am

**The image's final path:**
captura de tela 2018-02-15 as 11 22 52 am

After applying this PR:

**Correct image path:**
captura de tela 2018-02-15 as 11 23 28 am

So, as we can see on breadcrumb, the path of image after this PR is '/planilhas/dealerships/image.png' as showed on the select option (second image).

Anything else, I'm available to answer.


vblinden:
This works for me and I can't detect any problem!

src/Http/Controllers/VoyagerMediaController.php
```diff
@@ -129,7 +129,7 @@ public function move_file(Request $request)
         $source = "{$location}/{$source}";
         $destination = strpos($destination, '/../') !== false
             ? $this->directory.'/'.dirname($folderLocation).'/'.str_replace('/../', '', $destination)
-            : "{$location}/{$destination}";
+            : "/{$destination}";
```
fletch3555, line 132: Is the starting `/` necessary? This was a relative path previously.  I'm thinking this could come back and bite us, but I don't fully understand this code right now, so if it's necessary, just say so.

regiszanandrea: I will make more tests, and I will return to you with a better answer.

regiszanandrea: @fletch3555 So, I made some tests and both works fine. So, you decide if you want that I remove the `/`. 

A example of the value of `$destination` variable when I try to move a file:

With slash:`/pages/test/l7a0l6M2OU99o25Ey2kMMGTESVH4spJNA92Y5968.jpeg`

Wihout slash:  `pages/test/l7a0l6M2OU99o25Ey2kMMGTESVH4spJNA92Y5968.jpeg`

regiszanandrea: Another thing: When I try to move a file to a parent folder, the value of `$destination` variable is : 

`/.//l7a0l6M2OU99o25Ey2kMMGTESVH4spJNA92Y5968.jpeg`

fletch3555: That's a REALLY weird path when you move the file... Does it still work?

regiszanandrea: About this: `/.//l7a0l6M2OU99o25Ey2kMMGTESVH4spJNA92Y5968.jpeg`? Yes, it's works. Without and with `/`.

regiszanandrea:
@fletch3555 Any news?

fletch3555:
I'm still not really liking that path, but if it works, I'll allow it.  I fully expect it to cause some kind of issue once released.

fletch3555 (approved):

regiszanandrea:
@fletch3555 but this path: `/.//l7a0l6M2OU99o25Ey2kMMGTESVH4spJNA92Y5968.jpeg` is generated by the voyager code, not by my change.

fletch3555:
I understand that, which is why I'm merging it.

regiszanandrea:
@fletch3555 oh ok, thanks for that.

PR #2677, opened 2018-02-16
Fixes #2075 and #1949

### LINKED ISSUE #1949: Modal wrong action routing

vblinden:
When deleting a menu, canceling it and opening another menu again. The action has two id's like this: .../2/1. This pull requests fixes that.

This will also fix #2075 and fix #1949


fletch3555:
Can you make sure this applies to all types? This behavior likely exists in all Voyager BREAD views

vblinden:
@fletch3555 Sure, I will check every delete modal and update the PR.

fletch3555:
This is essentially the same as #2670.  One uses `__id`, and the other uses the index route and appends the id.  Can you make them function the same (I prefer using the correct route, not assuming the delete route is the same as appending the id to the index route).

vblinden:
Fixed all the other instances I could find. Some have different routes and so we have to append the id to the end and not replace the `__id`.

resources/views/bread/browse.blade.php
```diff
@@ -279,16 +279,7 @@
...
-
-            if (!deleteFormAction) { // Save form action initial value
-                deleteFormAction = form.action;
-            }
-
-            form.action = deleteFormAction.match(/\/[0-9]+$/)
-                ? deleteFormAction.replace(/([0-9]+$)/, $(this).data('id'))
-                : deleteFormAction + '/' + $(this).data('id');
-            console.log(form.action);
+            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.index') }}' + '/' + $(this).data('id');
```
fletch3555: Can you make this one use the `__[x]` placeholder like the rest?

resources/views/menus/browse.blade.php
```diff
@@ -106,9 +106,7 @@
         });
 
         $('td').on('click', '.delete', function (e) {
-            id = $(e.target).data('id');
-
-            $('#delete_form')[0].action += '/' + id;
+            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.index') }}' + '/' + $(this).data('id');
```
fletch3555: This one too.

fletch3555 (changes requested):

fletch3555:
Sorry, just saw your comment about needing to append.  What do you mean they "have different routes"?  If it's an issue where some are ".delete" and some are ".destroy", then I think that's a different issue that should be changed as well.

vblinden:
Yes, we generate all the routes for the datatypes. But we set the routes for database and setting our self. So they differ from the generated ones. I could make that a different pull request, where we make the routes the same?

fletch3555:
Yes, please make them all the same (bread ones should be resource routes, so they're all `destroy`). We did this same thing a while back, so we must've just missed some.  Go ahead and test that under the 1.x-dev branch to make sure it wasn't just merged there for some reason

fletch3555:
Yes, please make them all the same (bread ones should be resource routes, so they're all `destroy`). We did this same thing a while back, so we must've just missed some.  Go ahead and test that under the 1.x-dev branch to make sure it wasn't just merged there for some reason

vblinden:
I made them all append (so we have some unity, like the resource routes). Are there any other delete modals that I missed?

fletch3555 (approved):

emptynick:
Something went wrong [here](https://github.com/the-control-group/voyager/pull/2677/files#diff-52e64b52374bf2d87cd76ecbb043ca63R109)
You set the form-action to `database.destroy` when someone trys to delete a menu.
Good thing is that `table` is not defined, so this just doesn't work.
Otherwise it would just delete a whole table

vblinden:
Wow! Good catch! Should I make a pr to fix this?

emptynick:
No, I did already: https://github.com/the-control-group/voyager/pull/2892

PR #2678, opened 2018-02-16
Add the ability to add custom actions buttons

vblinden:
Hello everyone,

At the moment there is no way to define a custom action in the BREAD browse overview. Only the default view, delete and edit actions are available. With this pull request it is possible for someone to add a for example a download action to a certain datatype.

The way you would do this is create a new class `DownloadAction` and extend it with the `AbstractAction`. You can give it a title, icon and set the attributes. You can also specify which on which DataType you want too show the action and define custom routes for each datatype.

Let me know what you guys think! Feedback is appreciated!


resources/views/bread/browse.blade.php
```diff
@@ -175,21 +175,17 @@
...
-                                                    <i class="voyager-edit"></i> <span class="hidden-xs hidden-sm">{{ __('voyager.generic.edit') }}</span>
-                                                </a>
-                                            @endcan
-                                            @can('read', $data)
-                                                <a href="{{ route('voyager.'.$dataType->slug.'.show', $data->{$data->getKeyName()}) }}" title="{{ __('voyager.generic.view') }}" class="btn btn-sm btn-warning pull-right">
-                                                    <i class="voyager-eye"></i> <span class="hidden-xs hidden-sm">{{ __('voyager.generic.view') }}</span>
-                                                </a>
-                                            @endcan
+                                            @foreach(Voyager::actions() as $action)
+                                                @php $action = new $action($dataType, $data); @endphp
```
fletch3555: Can we move the rendering of this into a view partial? Then this code would just be this: https://laravel.com/docs/5.5/blade#rendering-views-for-collections

fletch3555 (changes requested):

fletch3555:
This will need to go in `1.x-dev`.  Make sure you talk to @emptynick about his BREAD redesign so this can get implemented with it

vblinden:
We can't use the `@each` blade helper cause that doesn't inherit the variables of the parent view and we need the `$dataType` and `$data`. So now we use the `@include` blade helper with a different view.

fletch3555:
Thanks.  Can you resolve the conflict as well?  Just pull 1.x-dev into your branch, discard the conflicted change, keeping your version (with the @foreach and @include), then update your Action classes to use the `voyager::` prefix in your `getTitle()` methods

fletch3555 (approved):

alexandrekadri:
hi, I think could be usefull if add like BREAD-EX permissions. For each BREAD  exists add permissions  "Export".

fletch3555:
This PR was merged ~3 months ago.  Bugs and Feature requests should be reported as their own issues here, not on closed issues/PRs.

Issue #1861, opened 2017-09-25
Custom Primary Key , Insert mode when editing


sistemaswebbrasil:
- Laravel Version: 5.5.*
- Voyager Version: 1.0
- PHP Version: 7.0
- Database Driver & Version: mysql

![captura de tela 2017-09-25 10 42 32](https://user-images.githubusercontent.com/28270491/30811720-6a1120cc-a1df-11e7-8f3b-41e350f73554.png)
![selecao_002](https://user-images.githubusercontent.com/28270491/30811739-753611e2-a1df-11e7-8540-d7a33bd8db96.png)

### Description:
When using a primary key with another non-default name like `protected $primaryKey = "sql_rowid";` the crud enters in add mode when clicking edit.

### Steps To Reproduce:
1) Changed primary key row name at the model.
2) Clicked on "edit" for a row on the table and the "Add" view was opened.

ataylor32:
I also ran into this issue. I think the fix, at least for the "resources/views/bread/edit-add.blade.php" file, is to replace all 6 instances of `$dataTypeContent->id` with `$dataTypeContent->getKey()` and then replace all 5 instances of `isset($dataTypeContent->getKey())` with `!is_null($dataTypeContent->getKey())`. There are probably other areas that need to be updated too, though.

emptynick:
This was likely fixed in #2690

PR #2690, opened 2018-02-19
Use getkey() instead of id if custom primary key used in model.

Julien-Git1:
Hi, This is the fix i'm using for custom primary key.
(as @ataylor32 proposed in #1861).


fletch3555:
I'm getting some serious deja vu here... I SWEAR we already added this.

Maybe we just missed a few, or it got reverted in some other change and we missed that.  In either case, thanks!

fletch3555 (approved):

gausam:
@fletch3555 Probably were thinking about this? https://github.com/the-control-group/voyager/pull/1890

PR #2697, opened 2018-02-19
Use Carbon instead of gmdate (already)

defaultpage:
Issue: #2369 
Previous PR: #2370


defaultpage:
I think you lost part of the code from previous PR.

fletch3555 (approved):

marktopper (approved):

PR #2698, opened 2018-02-19
Detect proper Symlink

emptynick:
This PR detects if the `public/storage` element is a real symlink.
If its a regular directory it will rename it to `storage_old` and create a real symlink.

There are 3 cases (which I tested)

1. The element doesn't exist at all - normal operation
2. The element exists and is a symlink - nothing happens
3. The element exists, but is a normal directory - directory gets renamed and symlink is generated


src/VoyagerServiceProvider.php
```diff
@@ -155,15 +155,25 @@ protected function addStorageSymlinkAlert()
...
-            $alert = (new Alert('missing-storage-symlink', 'warning'))
-                ->title(__('voyager.error.symlink_missing_title'))
-                ->text(__('voyager.error.symlink_missing_text'))
-                ->button(__('voyager.error.symlink_missing_button'), '?fix-missing-storage-symlink=1');
-
-            VoyagerFacade::addAlert($alert);
+        if (request()->has('fix-missing-storage-symlink')) {
+            if (file_exists(public_path('storage'))) {
+                if (readlink(public_path('storage')) == public_path('storage')) {
+                    rename(public_path('storage'), 'storage_old');
```
defaultpage, line 161: > [`rename()`](http://php.net/manual/en/function.rename.php): If renaming a file and newname exists, it will be **overwritten**. If renaming a directory and newname exists, this function will emit a **warning**.

Can you add `date('Y-m-d H-i-s')` or timestamp after `old`?

emptynick: Normally we tell users to delete this directory.
I think 99% of the users who have this problem dont even know/care that they have this folder

fletch3555:
One question.  Why are we fixing it automatically if it's the wrong type (directory vs symlink), but prompting the user with a button to fix it if it's outright missing?  Seems like they should either always be prompted, or always be auto-fixed

emptynick:
We dont fix it automatically.
It is just an additional condition: if symlink does not exist *or* the symlink is a folder. 
If one of this is true it will display the message to fix it.
Click that and it will rename the folder (if it is a folder) and create the symlink (as before)

fletch3555:
Ah, just looked again, and you're 100% right.  I didn't see it since the whole block was restructured.  At first glance, it looked like you just added the `rename()` line(s)

fletch3555 (approved):

marcelosantos89:
This fix breaks storage:link done in Windows causing the following error: readlink(): Invalid argument ... this is not a problem for production but it is for development

emptynick:
What is `public_path('storage')` in your case?
I actually wrote it on windows in dev-env

marcelosantos89:
printing that function gives me: /opt/application/public/storage
I'm using docker but the folder on my machine is more like a shortcut than a symlink thats why the errror but i removed it in my PC and its ok now

fletch3555:
Yeah, with docker mapped volumes, Windows hosts can read container symlinks, but Linux containers can't read host ntfs junctions(the windows version of a symlink)

PR #2710, opened 2018-02-21
Enable translation on belongTo field when in browse or view mode

Julien-Git1:
Hi all,
I needed to also see translation of fields with relationships (belongTo) at least in browse or read mode.

I used this modification to get what i wanted, maybe it can help ?

Bye!

capture


Julien-Git1:
Edited and PR done.

fletch3555:
Thanks

PR #2747, opened 2018-03-01
Correct tabindex on database/ select elements

### LINKED ISSUE #2744: tabindex="-1" on database/ select elements
- Laravel Version: 5.5.35
- Voyager Version: 1.0.16
- PHP Version: 7.1.8
- Database Driver & Version: MySQL 5.6.35

### Description:

When creating or editing a database, the select elements for both Type and Index have `tabindex="-1"`. Why? It is slightly annoying to me that I cannot tab through each input and instead have to click into those selects (particularly Type). Is there some reason for this, or is it oversight?

### Steps To Reproduce:

1. Log in to Voyager
1. Go to Tools -> Database
1. Click "+ Create New Table"
1. Click into the input under "Name"
1. Press tab repeatedly and notice that the select elements are skipped in the tab order

mattrabe:
Removed `tabindex="-1"` in two locations

I made this change and was able to still create and edit tables without issue. Being a keyboard junkie I found the ability to tab through the columns very nice :)

Fixes #2744

### DIFF
--- a/resources/views/tools/database/vue-components/database-column.blade.php
+++ b/resources/views/tools/database/vue-components/database-column.blade.php
@@ -31,7 +31,7 @@
     <td>
         <select :value="index.type" @change="onIndexTypeChange"
                 :disabled="column.type.notSupportIndex"
-                class="form-control" tabindex="-1">
+                class="form-control">
             <option value=""></option>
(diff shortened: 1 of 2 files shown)


marktopper:
Fixes #2744

marktopper (approved):

PR #2754, opened 2018-03-02
Allows editing of hasMany relations in BREAD-generated forms

MartynasJanu:
Creating a hasMany relation wouldn't allow the parent object form to edit the relations. For example, you can't add players to a team when only a player can play only in one team. You could only set the team when editing a player.

This adds the multiselect in the BREAD-generated form to choose child relations and controller code to handle the changes.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -76,15 +76,19 @@
 	            		<p>{{ $string_values }}</p>
 	            	@endif
 	            @else
-	            	@if(empty($selected_values))
-		            	<p>No results</p>
-		            @else
-		            	<ul>
-			            	@foreach($selected_values as $selected_value)
-			            		<li>{{ $selected_value }}</li>
-			            	@endforeach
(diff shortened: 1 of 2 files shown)


MartynasJanu:
Delete please

fletch3555:
I'm not sure what you changed between this and #2755, or if one was a duplicate, but if you closed this in order to change something, please don't do that.  I don't see any new commits (though it would update automatically if there were), and Github will let you edit the title and description after it's created, so there's no need to intentionally create duplicate PRs

MartynasJanu:
There was a styleCI error which I fixed but being to new to github I didn't understand at first how to amend a commit of a branch that already had a PR open. Now I know, so won't happen again :)

fletch3555:
No worries.  I'm glad you realized it.  Also, StyleCI will fix itself after we merge a PR, so you don't have to worry too much about fixing them.

MartynasJanu:
Oh, good to know. Thanks :)

PR #2755, opened 2018-03-02
Allows editing of hasMany relations in BREAD-generated forms

MartynasJanu:
Creating a hasMany relation wouldn't allow the parent object form to edit the relations. For example, you can't add players to a team when only a player can play only in one team. You could only set the team when editing a player.

This adds the multiselect in the BREAD-generated form to choose child relations and controller code to handle the changes.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -43,6 +43,7 @@ abstract class Controller extends BaseController
     public function insertUpdateData($request, $slug, $rows, $data)
     {
         $multi_select = [];
+        $has_many_relations = [];
 
         /*
          * Prepare Translations and Transform data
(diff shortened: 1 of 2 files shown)


HappyLDE:
Why is this closed? We need this like air.

emptynick:
This is not closed?

devswert:
Please merge this in the master branch 😞

marktopper:
Is anyone up for fixing the merge conflicts on this? Then I will do a test and merge it if it works properly.

hotsaucejake:
It seems this line is the conflict: https://github.com/the-control-group/voyager/blob/1.1/src/Http/Controllers/Controller.php#L59

This functionality would be huge for my use case.  I have a production server running v0.11 (using custom views that handle this) that I would love to update to v1.1 and use this.

SteamWind:
@hotsaucejake You could start a review for better tracking of the progress 😉

src/Http/Controllers/Controller.php
```diff
@@ -56,7 +57,12 @@ public function insertUpdateData($request, $slug, $rows, $data)
 
             // if the field for this row is absent from the request, continue
             // checkboxes will be absent when unchecked, thus they are the exception
-            if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
+            // Same exception applies to hasMany relations
+            if (!$request->hasFile($row->field) &&
+                !$request->has($row->field) &&
+                $row->type !== 'checkbox' &&
+                ($row->type !== 'relationship' || $options->type !== 'hasMany')
```
emptynick, line 64: Is this condition really what you wanted to do?

emptynick:
Unfortunately this PR has conflicting files and requested changes.
I'm going to close this PR because it has been untouched for more than a year.
We can reopen this if you fix the conflicts/requested changes.

PR #2793, opened 2018-03-09
Making non required migrations publishable

nickbryan:


fletch3555 (approved):

fletch3555:
We're going to make a small change to this.  The act of publishing the migrations will be done only when `--with-dummy` is selected.  We don't want unneeded migrations to be published repeatedly.  I'm going to merge this in now, and @emptynick will get the additional change in before we release v1.1

PR #2796, opened 2018-03-10
Split language files

emptynick:
This PR splits up every language file (`voyager.php`) to single grouped files.
This removes the necessity to use `__('voyager::voyager.group.key')`
Now we are able to use `__('voyager::group.key')`

All language-calls _should_ be replaced


marktopper (dismissed):
Great work!

fletch3555:
@emptynick, would you mind fixing the conflict? Once that's done, I'll get this merged in

fletch3555 (dismissed):

fletch3555 (approved):

PR #2797, opened 2018-03-10
Wrong fields alignemnt (padding) in create/edit form

reficul:
if you not pass "col-md-12" and if there are some custom display in other fields you have wrong alignment
schermata 2018-03-10 alle 13 23 08

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -59,8 +59,7 @@
                                 @if ($options && isset($options->formfields_custom))
                                     @include('voyager::formfields.custom.' . $options->formfields_custom)
                                 @else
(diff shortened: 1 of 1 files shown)


emptynick:
What about
`<div class="form-group col-md-{{ $display_options->width or 12 }}">`

fletch3555 (changes requested):
I agree that @emptynick's suggestion is better.  @reficul, please update this PR to use his suggestion.

fletch3555 (approved):

osc2nuke:
How i use that in the BREAD options?
`<div class="form-group col-md-{{ $display_options->width or 12 }}">`

Nevermind just figured it:
`{
    "display": {
        "width": 2
    }
}`

I could not find it in the docs, sorry!

emptynick:
https://voyager.readme.io/docs/display-options

osc2nuke:
Sigh.......... i really need to go for these glasses.

PR #2804, opened 2018-03-11
Adding legend option to edit-create blade

reficul:
you can add a new custom data field to define a legend for grouping fields:
```
    "legend": {
        "text": "Area File",
        "bgcolor": "#f0f0f0",
        "align": "center"
    }
```
to show something like:
schermata 2018-03-11 alle 12 32 47

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -56,7 +56,7 @@
                                     $options = json_decode($row->details);
                                     $display_options = isset($options->display) ? $options->display : NULL;
                                 @endphp
-                                @if (isset($options->legend->text))
+                                @if (isset($options->legend) && isset($options->legend->text))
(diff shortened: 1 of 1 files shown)


marktopper:
Once merged this needs to be added to the documentation.

resources/views/bread/edit-add.blade.php
```diff
@@ -56,6 +56,9 @@ class="form-edit-add"
                                     $options = json_decode($row->details);
                                     $display_options = isset($options->display) ? $options->display : NULL;
                                 @endphp
+                                @if (isset($options->legend->text))
```
marktopper: If `$options->legend` is not defined, I believe that it will complain about trying to get undefined property of null.

marktopper (changes requested):

marktopper (approved):

resources/views/bread/edit-add.blade.php
fletch3555: I believe `isset()` should apply to the whole chain. https://stackoverflow.com/a/41099030

PR #2808, opened 2018-03-12
Ui locale selection

emptynick:
This allows to change the locale for users.
Locales are readed by the name of the files in `publishable/lang`

The locale is stored in a `settings` column (cast to array) for future usage as suggested by @marktopper


marktopper (dismissed):

fletch3555 (approved):

particle6:
I know this has already been merged. But if a user does not have this setting, the admin will throw an error upon login.

PR #2809, opened 2018-03-12
Dummy Data removed from normal installation

emptynick:
Now installing Voyager without `--with-dummy` flag does **not** do the following (anymore):

- Publish posts/pages/categories migrations and run them
- Publish demo assets (`storage/app/public`)
- Register widgets in config by default
- Set Menu Items
- Create content
- And set permissions

The Seeders may need some love.
Also everything needs some testing.
I did try to install with and without dummy-data and it worked as expected.


emptynick:
Closing in favor of #2829

PR #2832, opened 2018-03-16
Load all permission first

jannejava:
Loading all user permissions and data types at once instead of querying one by one. For me, this has cut down the number of database queries from 44 to 22. Have a look and see it could be something.


emptynick:
Bet you wanted to put this into `1.x-dev` ;)

src/Policies/MenuItemPolicy.php
```diff
@@ -8,7 +8,8 @@
 
 class MenuItemPolicy extends BasePolicy
 {
-    protected static $datatypes = [];
+    protected static $datatypes = null;
```
emptynick: Any particular reason you changed this from `[]` to `null`?

emptynick (approved):
Tested and working fine!

fletch3555 (approved):

PR #2840, opened 2018-03-18
Make Actions column not searchable

MrCrayon:
Searching "del" or any other string contained in actions results as no filter because actions column is searched.


emptynick (dismissed):
Makes sense and works as expected

fletch3555 (approved):

Issue #2861, opened 2018-03-21
Slow rendering on list caused by readmore


MrCrayon:
- Laravel Version: 5.5.39
- Voyager Version: 1.0.16
- PHP Version: 7.1.14
- Database Driver & Version: 10.1.30-MariaDB

### Description:
Rendering in bread browse of a table with several records with server pagination off is really slow, I found out the main culprit is the hard coded class="readmore" for text, text_area and rich_text_box.

Here is the Performance analysis with Chrome for around 600 records 1 text field display:

![readmore-enabled](https://user-images.githubusercontent.com/6875243/37694082-7b15b5de-2cc4-11e8-8181-bb7d8b689c17.png)

and here is the analysis overriding the template file and removing class="readmore"

![readmore-disabled](https://user-images.githubusercontent.com/6875243/37694109-ba0211e8-2cc4-11e8-9775-81dfeefc44d3.png)

It would be great to have the readmore functionality as an option since is badly affecting performances.

I'm going to make a pull request shortly about this.

### Steps To Reproduce:
Create a bread with a text field, leave server side pagination off.
Fill the table with hundred of records (even if affects also smaller quantities)
Go on the listing page and wait page to render

easymomo:
I don't know if it will help you but I alwas select "Server Side Rendering" in the BREAD when I know I will have a lot of data. I only keep BREAD like "Categories" with client side rendering

MrCrayon:
Of course with Server side pagination enabled you'll only have 20 records displayed so there is no problem but the issue is that there shouldn't be any problem even with server side pagination disabled.
This is my result for 1700 records with server side pagination disabled, readmore removed and an array passed to datatable instead of creating the table in html server side.

![schermata da 2018-04-15 15-45-57](https://user-images.githubusercontent.com/6875243/38779117-2cd5a672-40c4-11e8-90dd-ecd866504009.png)

danielbidala:
I think we should keep open this issue. I don't know what's going on here exatly but if I remove readmore classes from browse.blade.php my loading time decrease by 4 second!

I use server side rendering.

Rendering take more than 4seconds to me with readmore classes . The server response is only 46ms.

emptynick:
Closing as https://github.com/the-control-group/voyager/pull/3671 was merged.

PR #2862, opened 2018-03-21
remove hardcoded readmore class and add custom classes through bread details

MrCrayon:
This solves #2861 and also adds the functionality to be able to add custom classes in the listing.

Adding this in bread details for a filed:

```
{
    "browse": {
        "thead": {
            "classes": [
                "col-sm-8"
            ]
        },
        "tbody": {
            "classes": [
                "text-primary",
                "text-uppercase"
            ]
        }
    }
}
```
Is going to add class="col-sm-8" to th tag, so we can format column widths, hide some columns for small screens and all that cool stuff.
And is going to add the other two classes to the associated td tag.

And of course who wants to use readmore can add that back.

I'm not sure about naming conventions, there is a display property already used to add  some custom properties to the add-edit template, maybe display_browse could be used in this case.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -57,7 +57,8 @@
                                             <th></th>
                                         @endcan
                                         @foreach($dataType->browseRows as $row)
-                                        <th>
+                                            <?php $options = json_decode($row->details); ?>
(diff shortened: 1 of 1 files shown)


emptynick:
So I dug into `readmore` and found out, that its not working at all.
We are shortening the strings to 200 chars max, and I assume readmore works with strings > 200.

We can discuss if we remove `readmore` completely.
As for your PR, we need to document this

PR #2863, opened 2018-03-21
Fix migration role_id relationship

luritas:
When I rollback migrations, error display below
```
Illuminate\Database\QueryException  : SQLSTATE[HY000]: General error: 1832 Cannot change column 'role_id': used in a foreign key constraint 'users_role_id_foreign' (SQL: ALTER TABLE users CHANGE role_id role_id INT DEFAULT NULL)
```
I changed reverse migration code on `migrations/2017_11_26_013050_add_user_role_relationship.php`

### DIFF
--- a/migrations/2017_11_26_013050_add_user_role_relationship.php
+++ b/migrations/2017_11_26_013050_add_user_role_relationship.php
@@ -27,8 +27,8 @@ class AddUserRoleRelationship extends Migration
     public function down()
     {
         Schema::table('users', function (Blueprint $table) {
-            $table->integer('role_id')->change();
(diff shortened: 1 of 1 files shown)


marktopper:
Seems like you are trying to merge in a `1.x-dev` fork into `1.0`.
This should have been sent to `1.x-dev` instead.

fletch3555 (approved):

strickdj:
@luritas I still get this error. I had to change the code to the following:
``` Schema::table('users', function (Blueprint $table) {

        });
	        $table->integer('role_id')->change();
        });
```

PR #2864, opened 2018-03-21
 fix: package.json to reduce vulnerabilities 

x1mdev:
The following vulnerabilities are fixed with an upgrade:
- https://snyk.io/vuln/npm:bootstrap:20160627

Latest report for x1mdev/voyager:package.json:
https://snyk.io/test/github/x1mdev/voyager?targetFile=package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "ace-builds": "^1.2.6",
     "animate.css": "^3.5.2",
-    "bootstrap": "^3.3.7",
+    "bootstrap": "^4.0.0",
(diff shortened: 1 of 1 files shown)


emptynick:
You cant just replace Bootstrap 3 with 4.
This will blow up our design.

x1mdev:
@emptynick I see.. let me check for some other possibilities.

emptynick:
It says it was fixed in `3.4.0` but I doubt that there is such a version.
The only way is to upgrade to `4.0.x` but this requires a lot of work

marktopper:
Updating to bootstrap `v4.0.x` requires more work than this.

Issue #2661, opened 2018-02-13
Sort by relationship column


defaultpage:
- Laravel Version: 5.5.34
- Voyager Version: 1.0.15
- PHP Version: 7.1.0
- Database Driver & Version: mysql

### Description:
Sorting by relationships, not work for me. Maybe it's because of my changes, but I think I saw the problem after installation too. Can you check if the sorting works for you?

We can see `getRelationships` method where I found next code:
```php
        $relationships = [];

        $dataType->browseRows->each(function ($item) use (&$relationships) {
            $details = json_decode($item->details);
            if (isset($details->relationship) && isset($item->field)) {
                 ...
            }
        });

        return $relationships;
```
But in relationship details:
```
{"model":"App\\Product","table":"products","type":"belongsTo","column":"product_id","key":"id","label":"name","pivot_table":"admins","pivot":"0"}
```

### Steps To Reproduce:
1. Create relationship (https://voyager.readme.io/docs/relationships)
2. Sort table by new relationship

ArtemMolotov:
Example of temporary fix: 
```php
trait FixBreadRelationshipSortFilter
{
    public static function isJoined($query, $table){
        $joins = collect($query->getQuery()->joins);
        return $joins->pluck('table')->contains($table);
    }

   ...
    public function index(Request $request)
    {
        ...
            if ($orderBy && $row = $dataType->browseRows()->whereField($orderBy)->first()) {
                $querySortOrder = (!empty($sortOrder)) ? $sortOrder : 'DESC';
                if ($row->type === 'relationship'){
                    $details = json_decode($row->details);
                    $tmpOrderBy = $orderBy;
                    $orderBy = "{$details->table}.{$details->label}";
                    $model->belongsTo($details->model);
                    if (!static::isJoined($query, $details->table)){
                        $query = $query->leftJoin($details->table, $details->column, '=', $details->table . '.' . $details->key);
                    }
                }
                $dataTypeContent = call_user_func([
                    $query->with($relationships)->orderBy($orderBy, $querySortOrder),
                    $getter,
                ]);
                if ($row->type === 'relationship'){
                    $orderBy = $tmpOrderBy;
                }
            }
        ...
    }
}
```

P.S.: Very bad code, but...

vblinden:
Could you further explain what you mean with "Sort table by new relationship"? This is unclear to me. Thanks!

defaultpage:
When you click on 'Продукт' nothing happens.

[...]

PR #2871, opened 2018-03-22
Load relationship

### LINKED ISSUE #2870: Relationship not loaded in browse in some conditions
- Laravel Version: 5.5.39
- Voyager Version: 1.0.16
- PHP Version: 7.1.14
- Database Driver & Version: 10.1.30-MariaDB

### Description:
When using built in relationships defined in BREAD details field _label_ chosen doesn't show in browse 
if _timestamps_ are defined and no sorting is used.
The problem is a missing _->with($relationships)_ in VoyagerBreadController:

```
            if ($orderBy && in_array($orderBy, $dataType->fields())) {
                $querySortOrder = (!empty($sortOrder)) ? $sortOrder : 'DESC';
                $dataTypeContent = call_user_func([
                    $query->with($relationships)->orderBy($orderBy, $querySortOrder),
                    $getter,
                ]);
            } elseif ($model->timestamps) {
                $dataTypeContent = call_user_func([$query->latest($model::CREATED_AT), $getter]);
            } else {
                $dataTypeContent = call_user_func([$query->with($relationships)->orderBy($model->getKeyName(), 'DESC'), $getter]);
            }
```
I think it got lost with pull request  #937 , @victorlap noticed it but it looks like nobody replied so I don't know if there is a reason behind it even if I can't think of any.

### Steps To Reproduce:
Install Voyager with dummy
Edit posts bread adding to category_id:
```
{
    "relationship": {
        "key": "id",
        "label": "name",
        "method": "category"
    }
}
```
Browse Posts, only _category_id_ is shown instead of _name_

MrCrayon:
This fixes #2870

### DIFF
--- a/src/Http/Controllers/VoyagerBreadController.php
+++ b/src/Http/Controllers/VoyagerBreadController.php
@@ -48,10 +48,10 @@ class VoyagerBreadController extends Controller
 
         // Next Get or Paginate the actual content from the MODEL that corresponds to the slug DataType
         if (strlen($dataType->model_name) != 0) {
-            $model = app($dataType->model_name);
-            $query = $model::select('*');
-
             $relationships = $this->getRelationships($dataType);
+
+            $model = app($dataType->model_name);
+            $query = $model::select('*')->with($relationships);
 
             // If a column has a relationship associated with it, we do not want to show that field
             $this->removeRelationshipField($dataType, 'browse');
@@ -65,13 +65,13 @@ class VoyagerBreadController extends Controller
             if ($orderBy && in_array($orderBy, $dataType->fields())) {
                 $querySortOrder = (!empty($sortOrder)) ? $sortOrder : 'DESC';
                 $dataTypeContent = call_user_func([
-                    $query->with($relationships)->orderBy($orderBy, $querySortOrder),
+                    $query->orderBy($orderBy, $querySortOrder),
                     $getter,
                 ]);
             } elseif ($model->timestamps) {
-                $dataTypeContent = call_user_func([$query->with($relationships)->latest($model::CREATED_AT), $getter]);
+                $dataTypeContent = call_user_func([$query->latest($model::CREATED_AT), $getter]);
             } else {
-                $dataTypeContent = call_user_func([$query->with($relationships)->orderBy($model->getKeyName(), 'DESC'), $getter]);
+                $dataTypeContent = call_user_func([$query->orderBy($model->getKeyName(), 'DESC'), $getter]);
             }
 
             // Replace relationships' keys for labels and create READ links if a slug is provided.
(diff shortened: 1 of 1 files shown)


fletch3555 (changes requested):
Instead of adding the relationships there, can you remove it from the other two and add it to the end of `$model::select('*')` on line 52, after moving that line down below the definition of `$relationships`?

MrCrayon:
Thanks for suggestion, that looks much better.
I'm still quite new to Laravel Framework and Voyager so I try to change as less as possible.
I moved relationship up, if that's ok for you.

defaultpage:
@MrCrayon, @fletch3555, works for me. Thanks.

marktopper:
We're no longer accepting changes for the 1.0 branch since we're so close to a 1.1 release. Can you please update the PR to resolve the conflicts?

MrCrayon:
I'm quite new on github, I checked the **Resolve conflicts** but in 1.x-dev changes should be applied to a different file _VoyagerBaseController_ so I'm not sure how I can resolve that.
Maybe it would be easier to submit a new pull request on 1.x-dev.
Suggestions?

marktopper:
If you could do that, then it would be great.
Else we will leave this here, and hopefully I will get time to look into it :)

MrCrayon:
Sure I'll check it later, thanks :)

fletch3555:
Closing in favor of #2983

morettyCat:
Hi there, I think something is going wrong in this section of code. Let me explain:

When you go to "Browse" some model and then you click on row's name on header's table, results are not being listed correctly.

### Steps to reproduce

1. Create a _First Model_ (paged server side) with timestamps
2. Create a _Second Model_
3. Create a _Pivot Table_
4. Create a relationship BelongsToMany in _First Model_ pointing to the _Second Model_ (using pivot table). 
5. Go to browse _First Model_
6. Click on the field that shows relationship in order to sortBy this field
7. Look how it comes **unordered** :(


### My suspicions

As far as I played in order to solve this, I was able to find something interesting:

When the request came with this :

/admin/firstmodel?page=1&sort_order=desc&**order_by=firstmodel_belongstomany_secondmodel_relationship**

then the VoyagerBaseController.php on the if condition on line 65:

65. `if ($orderBy && in_array($orderBy, $dataType->fields())) {`

gets resolved on the timestamp "else" on line 71:

71. `}elseif ($model->timestamps) {`

and that's not correct, because "**what if we have timestamps but we asked the UI to sort results by relationship?**" the logic in this file says that it will be sorted by timestamps, but that's not correct as we specifically selected a field to sort by.

### Temporary solution

Delete the eleseif statement placed above in order to go to the last "ELSE" which applies the order correctly.

I expect it's clear... it's my first post. Congratulations for that AMAZING project.

Any real solution? I supose that deleteing the elseif that deserves to a logical condition is not a real solution...

defaultpage:
@morettyCat Hi! If I understand you correctly, then your problem is identical to [this one (#2661)](https://github.com/the-control-group/voyager/issues/2661).

morettyCat:
Yes @ArtemMolotov ! I did not read this issue before, sorry for that, i'll try to solve this issue by this method (but as far as I can see, i'm on 1.1 and your link relates to 1.0 if I'm not wrong). But i'll read it before asking again, so sorry!

PR #2872, opened 2018-03-22
Fix pt br lang

Braiani:
There is a misspelling on the Dimmer file under pt_br lang folder. This PR fix it.


marktopper (approved):
The change seems right when checking up the `pt` dictionary.

Issue #2182, opened 2017-11-07
Add Optional Details to Relationship Field


abacram:
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7
- Database Driver & Version:
MySQL
### Description:
Is it possible to add an Optional Details field to a Relationship Field type in the database tool?
I think it could be nice in order to improve display options.

And also i think it would be helpful to add a link to the relationship that is provided in the model we are working (add/updating). I think that it was possible in older versions, is that right?

bimo13:
Hi @abacram , have you found a workaround for this issue ? thanks in advance :)

p6x0016:
Hi, 
Im just new here, i did @abacram want but its just a workaround.

Follow the steps below:

1) Create your relationship field in BREAD manager / Database.
2) Check your database (phpmyadmin) and find the row in the "data_rows" table - use search via data_type_id
3) Once found, edit the relationship field.
4) Edit the "details" field, make sure it will still be enclosed to be read as JSON.
example: `{"model":"App\\projects","table":"projects","type":"belongsTo","column":"id","key":"id","label":"name","pivot_table":"categories","pivot":"0","display":{"width":"3"}}`

**"display":{"width":"3"}**
The highlighted code is the added code to allow the Optional Details to work.

Hope this help everyone checking this.

xavibaz:
Hi,

I'm in 1.1.1 now, and the problem persists. If you add ```width``` to the ```details``` field manually in the ```data_rows``` table it works, but when bread is updated, ```width``` is gone.

Digging in the code, I found out that ```DataType::getRelationships``` called in ```DataType::updateDataType``` overwrites the content without having in mind "extra details".

The details @MrCrayon explained in #2874 saying that this was fixed with that commit seem wrong to me (maybe it was like that in 1.0), because now it is managed by VoyagerBreadController and there is no such call to that function. 

**I believe one solution would be** to add a new field in the relationship part (see images) to add options like in the other data_rows.

![image](https://user-images.githubusercontent.com/34057208/40430658-e93a1cb8-5ea5-11e8-9414-2b35741a693d.png)

![image](https://user-images.githubusercontent.com/34057208/40430988-afce7b3a-5ea6-11e8-8a68-54566c7ff0ad.png)

And then merge those options with the ones related to the relationship.

[...]

PR #2874, opened 2018-03-22
merge details with referenced field

MrCrayon:
Before removing the overridden DataRow gets details fields and merges it in the displayed DataRow.
Solves #2182

### DIFF
--- a/src/Http/Controllers/Traits/BreadRelationshipParser.php
+++ b/src/Http/Controllers/Traits/BreadRelationshipParser.php
@@ -18,6 +18,14 @@ trait BreadRelationshipParser
                 $options = json_decode($row->details);
                 $relationshipField = @$options->column;
                 $keyInCollection = key($dataType->{$bread_type.'Rows'}->where('field', '=', $relationshipField)->toArray());
+                $originalDetails = '';
+                if ($dataType->{$bread_type.'Rows'}->has($keyInCollection)) {
+                    $originalDetails = $dataType->{$bread_type.'Rows'}->get($keyInCollection)->only('details')['details'];
+                }
+                if (!empty($originalDetails)) {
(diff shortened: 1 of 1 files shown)


emptynick:
I dont really understand what this does.
~~No matter what, #2182 is about adding details to a relationship, not merging them back once deleted.~~

MrCrayon:
@emptynick I should have probably explained it further.
When you create a relationship field you can't choose details because that field is hidden and used to define the relationship, but you can set details for the field associated to that relationship.
As example let's say I have Posts with category_id, I can set details for that field, then I add a relationship belongsTo Category using category_id as foreign key.

When Bread for Posts is shown removeRelationshipField is called and that removes category_id with all information stored in its details field.

My commit gets on the fly the details from category_id and merge it with relationship field details, so that when relationship field is shown it carries the details of the related field.

Basically is the same as the workaround proposed in issue #2182 but it's done on the fly without modifying db data.

BTW I checked that there is no conflict between documented keywords used in details field and keywords used to define relationship.

emptynick:
Ok I understand, but this only works on Belongs To Relationships then? Or does it also take the details from a foreign column?

MrCrayon:
Yes, because for other relationship types there is no field referenced in the same table.
A better solution would be to move relationship data in another field and allow details to be customized like any other field.
This is just a temporary solution, I think all BREAD system is under reconstruction so a more complicated solution might not be needed.

emptynick:
Each relationship has its own `data_row` so we _could_ display a normal JSON-editor, maybe in a `collapsible` so it is hidden first

MrCrayon:
The problem I see is that you will have at the same time two ways to change that data, JSON-editor and current interface.
One way to solve this duplicity is to remove the relationship data before displaying JSON-editor and merge it again before saving to db.

PR #2876, opened 2018-03-22
Add resize option to thumbnails

reficul:
There are no way to resize an image to make a thumbnail, you can only scale or crop original image.
With this PR I add the "resize" option also to thumbs.

usage:
```
    "thumbnails": [
        {
            "name": "resizedthumb",
            "resize": {
                "width": null,
                "height": "250"
            }
        }
    ]
```

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -290,13 +290,13 @@ abstract class Controller extends BaseController
                                             }
                                         }
                                     )->encode($file->getClientOriginalExtension(), $resize_quality);
-                                } elseif (isset($options->thumbnails) && isset($thumbnails->crop->width) && isset($thumbnails->crop->height)) {
+                                } elseif (isset($thumbnails->crop->width) && isset($thumbnails->crop->height)) {
                                     $crop_width = $thumbnails->crop->width;
                                     $crop_height = $thumbnails->crop->height;
                                     $image = Image::make($file)
                                         ->fit($crop_width, $crop_height)
                                         ->encode($file->getClientOriginalExtension(), $resize_quality);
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/Controller.php
```diff
@@ -296,6 +296,25 @@ function (Constraint $constraint) use ($options) {
                                     $image = Image::make($file)
                                         ->fit($crop_width, $crop_height)
                                         ->encode($file->getClientOriginalExtension(), $resize_quality);
+                                } elseif (isset($options->thumbnails) && isset($thumbnails->resize) && (isset($thumbnails->resize->width) || isset($thumbnails->resize->height))) {
```
emptynick: Theres already a check if `$options->thumbnails` is set on [Line 268](https://github.com/reficul/voyager/blob/8433b9e82dc495dc61be6e9a732afba72e685886/src/Http/Controllers/Controller.php#L268)
I know, its the same for #293 maybe you can remove that check also!

src/Http/Controllers/Controller.php
```diff
@@ -437,7 +456,26 @@ function (Constraint $constraint) use ($options) {
                                 $image = Image::make($file)
                                     ->fit($crop_width, $crop_height)
                                     ->encode($file->getClientOriginalExtension(), $resize_quality);
-                            }
+                            } elseif (isset($options->thumbnails) && isset($thumbnails->resize) && (isset($thumbnails->resize->width) || isset($thumbnails->resize->height))) {
```
emptynick: Same here, see [Line 428](https://github.com/reficul/voyager/blob/8433b9e82dc495dc61be6e9a732afba72e685886/src/Http/Controllers/Controller.php#L428)

emptynick (changes requested):

marktopper:
Sorry, we are closing issues for `v1.0` do to the release of `v1.1` that is out in a bit.
You are welcome to remake this for `v1.1`.

PR #2885, opened 2018-03-23
Additional relationship attributes/accessors

emptynick:
## Needs documentation

This allows to use Accessors as display-values in relationships.

### Usage:
In your model define an Accessor like described in the [Laravel docs](https://laravel.com/docs/5.6/eloquent-mutators#accessors-and-mutators)
```
public function getFullNameAttribute()
{
    return "{$this->first_name} {$this->last_name}";
}
```
Then add the attribute to `$additional_attributes` like:
`public $additional_attributes = ['full_name'];`


fletch3555 (approved):

mdenitti:
respect!

EjjE101:
This is a great sollution, thanks!

NightZpy:
This solution doesn't works for BelongsToMany relations.

SteamWind:
Doesn't work with HasMany too:
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'fullinfos' in 'field list' (SQL: select `fullinfos` from `mytable` where `relation_id` = 3)
```
But this was declared in the relation's Model:
```php
    public $additional_attributes = ['fullinfos'];
[...]
    public function getFullinfosAttribute()
    {
        return $this->first_attr . ' ' . $this->second_attr;
    }
```
I'll check this merge to see if I can do something

emptynick:
This just means that a `$model->fullinfos` does not hit your Accessor, for whatever reason.

SteamWind:
@emptynick I think I found where it doesn't work. Somewhere in relationship.blade.php, it pluck the request from the query builder with the label, which is fullinfos. This generate a query with fullinfos in the select, what we need to do is to retrieve the model and then pluck it (the model knows what fullinfos is but not the query builder...).
I'll try it deeper and I should post a pull request if it works...

SteamWind:
Got it, in relationship.blade.php:
```php
$selected_values = $model::where($options->column, '=', $relationshipData->id)->pluck($options->label)->all();
```
we need to retrive the result before plucking like so:
```php
$selected_values = $model::where($options->column, '=', $relationshipData->id)->get()->pluck($options->label)->all();
```
This works perfectly. I'm gonna make a pull request.

mdenitti:
Great work guys! this feature is getting more and more traction 👍 and is pretty dam usefull 🥇

PR #2886, opened 2018-03-23
Fix for spatial column type save error on mysq 5.7

bab55z:
This commit fixes the following error.
When saving spacial type column using `MySQL 5.7`, it returns `error near the spacial column value`.

Reproducing error
```
Voyager 1.x-dev
Mysql 5.7.14
PHP 7.2.3
```


emptynick:
I assume this will also work with lower version than 5.7?

bab55z:
@emptynick  Yes it is working tested also on 5.6 and runing fine
[sqlfiddle](http://sqlfiddle.com/#!9/746e8/2/0)

RoyIrwan (approved):
Tested on 10.2.12-MariaDB and works like a charm

fletch3555 (approved):

PR #2889, opened 2018-03-24
Fix for spatial traits, wrong coordinates order in getCoordinates() function

bab55z:
Getting `lat `and `lng `data from point value with the `list` function is ambiguous and incoherent between `PHP5 `and `PHP7`
This lead to wrong or right location coords depending on the verion of php you use
Excerpt from PHP.net documentation below ([php list function](http://php.net/manual/en/function.list.php))

> Warning
> In PHP 5, list() assigns the values starting with the right-most parameter. In PHP 7, list() starts with the left-most parameter.
> 
> If you are using plain variables, you don't have to worry about this. But if you are using arrays with indices you usually expect the order of the indices in the array the same you wrote in the list() from left to right, which is not the case in PHP 5, as it's assigned in the reverse order.
> 
> Generally speaking, it is advisable to avoid relying on a specific order of operation, as this may change again in the future.


MrCrayon:
IMHO I don't think this patch is going to change anything since $lat and $lng are plain variables, therefore, not affected.

emptynick:
@MrCrayon you are right, see 
http://sandbox.onlinephpfunctions.com/code/ce68db17444bff512b6e4d93ec33432429ec405e

`If you are using plain variables, you don't have to worry about this.`

bab55z:
I'll investigate a bit more but currently this what i am facing.

I am runnin my code on php 7.2 as specified
Lets suppose you want to store New York location in a point data type. 
New York position on map is (40.712765, -73.884957)
lat -> 40.712765
lng -> -73.884957
the point is correctly stored 

But there is a issue when you retrieve the `lat `and `lng `from db using `list function` you get this.

`PHP5 list `
 [40.712765, -73.884957] -> **corresponds to new york**

`PHP7 list `
 [-73.884957, 40.712765] -> **corresponds to Antarctic** 

I think this should be addressed because it can leads an app to provide unreal data and is not acceptable in the field of mapping.

I'll investigate and provide update, if anyone can also figure out what it happening i'll be good also

I had to override, some coordinates field type files to get consistent map positions across crud operations

MrCrayon:
I'm not using this Trait so I can't say for sure is working.
Can you provide the piece of code where you are using getCoordinates()?

RoyIrwan (approved):
Tested against PHP 7.2, and it works. Without this update, I got blank google map on view and edit (BREAD)

MrCrayon:
After running a script on different servers, find results below, I understood what's the real issue.
It has nothing to do with php version or the use of list but list parameters were simply swapped, fix should be:
```diff
-                        list($lat, $lng) = explode(' ', $point);
+                        list($lng, $lat) = explode(' ', $point);
```

script:
```php
<?php
$point = "lat lng";
list($lat, $lng) = explode(' ', $point);
$lngLat = explode(' ', $point);

echo phpversion(). "\n";
echo $lat . ' ' . $lng . "\n";
echo $lngLat[0] . ' ' . $lngLat[1] . "\n";
```

results:
```
5.4.45-0+deb7u11
lat lng
lat lng

5.6.11-1ubuntu3.4
lat lng
lat lng

7.0.28
lat lng
lat lng

7.1.15
lat lng
lat lng
```

bab55z:
@MrCrayon you are right, and your fix is right and ok.
But what about the comment on **php.net**. They state we should not rely on order of parameters returned by list function because it can be changed at any **php release**.

> Generally speaking, it is advisable to avoid relying on a specific order of operation, as this may change again in the future.

We can make sure future releases of php won't have any side effect on the `spatial trait` by not using the `list function ` as the order is not reliable
my solution is the array indexes as in my PR, any other solutions or we should simply swap the parameters of `list `(taking risks of inconsistences across **php releases**)

MrCrayon:
@bab55z it's not the order of parameters is the "order of operation", that's why it affects only if you use array with indices.
Try this with php5 and php7:
```php
$point = "lng lat";
$lngLat = [];
list($lngLat[], $lngLat[]) = explode(' ', $point);

echo phpversion(). "\n";
echo $lngLat[0] . ' ' . $lngLat[1] . "\n";
```

bab55z:
I meant ditching the list function
```
$point = "lng lat";
$lngLat = explode(' ', $point);
echo $lngLat[0];
echo $lngLat[1];
```

But you are right, we simply need to invert parameters order.

MrCrayon:
What I was trying to point out is that you misunderstood what php documentation stated, using list or using an array as in your first commit it's exactly the same thing, but IMHO using list is a bit more readable.

But the important thing is that another bug is about to be eliminated :thumbsup:

bab55z:
yes i misunderstood, that lead me to another field to find a solution. but every things is right now thanks for explanations.

emptynick:
Makes sense.
The saving-order was changed in https://github.com/the-control-group/voyager/pull/2425

RoyIrwan (approved):
This is also works against 10.2.12-MariaDB

fletch3555 (approved):

PR #2893, opened 2018-03-25
Update .php_cs file

KABBOUCHI:
Switch PHP CS Fixer syntax from v1 to v2

Using [PSR2](https://styleci.readme.io/docs/presets#section-psr2) rules minus [lowercase_constants](https://styleci.readme.io/docs/fixers#section-lowercase_constants) and [method_argument_space](https://styleci.readme.io/docs/fixers#section-method_argument_space)


KABBOUCHI:
please can you check StyleCI configuration to make sure we are using the same rules

marktopper:
Thanks, this seems to work for the new PHP CS Fixer.
Tested it and it does not mess with the currently StyleCI styled syntax - so far so good.

marktopper (approved):

PR #2908, opened 2018-03-28
Allows empty fields to be updated as null. Fixes #1402 

### LINKED ISSUE #1402: Impossible to update to null Edit BREAD
- Laravel Version: 5.4.23
- Voyager Version: 0.11.14
- PHP Version: 7.0.15
- Database Driver & Version: mysql 5.6.35

### Description:
Impossible to update to null Edit BREAD

### Steps To Reproduce:
I created a bread on a new table with a varchar(50) type field. I put "xxx" and save. All is ok.
I make a mistake and want to update the field to null. So i edit the table bread and erase the field entry and save.
But after saving, the field is always with the old value !

Best regards

xaviqv:
With Voyager 1.0 I can't update a filled up BREAD field leaving it empty. It keeps it's last value instead.

In Voyager file "src\Http\Controllers\Controller.php", inside the method insertUpdateData() there is a condition that makes any empty field to be escaped from the loop process:
`!$request->has($row->field)` 

In Laravel ^5.4, $request->has() evals to false on empty fields, so I replaced it to:

`!$request->exists($row->field)`

$request->exists() returns true if the field is in request although it may be empty.

It seems to work fine in my bread tables, but I'm not sure if it may affect negatively any other kind of field. What do you think?

It fixes #1402

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -50,8 +50,8 @@ abstract class Controller extends BaseController
 
             // if the field for this row is absent from the request, continue
             // checkboxes will be absent when unchecked, thus they are the exception
-            if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
-                continue;
+            if (!$request->hasFile($row->field) && !$request->exists($row->field) && $row->type !== 'checkbox') {
+				continue;
             }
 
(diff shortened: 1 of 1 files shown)


emptynick:
Sorry, but this is a duplicate of https://github.com/the-control-group/voyager/pull/2377
We really appreciate your effort though!

PR #2912, opened 2018-03-29
English translation fixes

ghost:
Moved English translation for `create_model_table` from lang/en/bread.php to lang/en/database.php and fixed minor translation bug on views/bread/browse.php

### DIFF
--- a/publishable/lang/en/bread.php
+++ b/publishable/lang/en/bread.php
@@ -19,7 +19,6 @@ return [
     'success_update_bread'   => 'Successfully updated the :datatype BREAD',
     'controller_name'        => 'Controller Name',
     'create_migration'       => 'Create migration for this table?',
-    'create_model_table'     => 'Create model for this table?',
     'description'            => 'Description',
     'display_name'           => 'Display Name',
     'display_name_plural'    => 'Display Name (Plural)',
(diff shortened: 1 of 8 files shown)


emptynick (approved):

fletch3555:
I disagree with moving the translation.  That specific key is only used on the BREAD create page.  If you move that one, you'll need to move the `migration` one right above it, which doesn't seem like it fits in the database file.

ghost:
This translation is used by the "Creat New Table" page. If it is not in the database translation then I believe the translation location needs to change on that page

emptynick:
The string he moved (`voyager::database.create_model_table`) is used in the database tool [here](https://github.com/the-control-group/voyager/blob/1.x-dev/resources/views/tools/database/vue-components/database-table-editor.blade.php#L17).
Those are just 2 different things

fletch3555:
Perhaps it should be duplicated then.  If it exists in 2 places with different/unrelated uses, then they should be different keys (potentially with the same value)

fletch3555:
We're just going to ignore my comments here.... @emptynick has shown me the error of my ways.

This looks good to me.

fletch3555 (approved):

PR #2914, opened 2018-03-29
Condition for permissions that don't exist fixed

ghost:
Condition for permissions that don't exist fixed for menu items.
Credits: @emptynick


emptynick (approved):
Tested and working

fletch3555 (approved):

PR #2922, opened 2018-04-01
Added the admin confirmation password as mentioned in issue #2896

### LINKED ISSUE #2896: Ask for password confirm on voyager:admin command
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7.2

It's a **suggestion not a bug**.

### Description:
I issued the command: `php artisan voyager:admin your@email.com --create` and misspelled my password for example: "passwordù", than when I tried to login did not worked and I had to remove the account and make it again.

It should ask to retype the password so the user can notice errors before creating the account.

swarnava112:
I have added the admin password confirmation when creating an admin from php artisan command.

Closes #2896

### DIFF
--- a/src/Commands/AdminCommand.php
+++ b/src/Commands/AdminCommand.php
@@ -50,6 +50,12 @@ class AdminCommand extends Command
             $this->option('create')
         );
 
+        // codegeek2014 #2896
+        if($user == null)
+        {
+            exit;
+        }
+
         // Get or create role
         $role = $this->getAdministratorRole();
 
@@ -117,10 +123,18 @@ class AdminCommand extends Command
         if ($create) {
             $name = $this->ask('Enter the admin name');
             $password = $this->secret('Enter admin password');
+            $confirmPassword = $this->secret('Confirm Password'); // codegeek2014 #2896
 
             // Ask for email if there wasnt set one
             if (!$email) {
                 $email = $this->ask('Enter the admin email');
+            }
+
+            // codegeek2014 #2896
+            if( $password != $confirmPassword)
+            {
+                $this->info('Passwords donot match');
+                return null;
             }
 
             $this->info('Creating admin account');
--- a/src/Commands/AdminCommand.php
+++ b/src/Commands/AdminCommand.php
@@ -129,10 +129,9 @@ class AdminCommand extends Command
                 $email = $this->ask('Enter the admin email');
             }
 
-            // Passwords donot match
+            // Passwords don't match
             if ($password != $confirmPassword) {
-                $this->info('Passwords donot match');
-                
+                $this->info("Passwords don't match");
                 return;
             }
 
--- a/src/Commands/AdminCommand.php
+++ b/src/Commands/AdminCommand.php
@@ -129,7 +129,7 @@ class AdminCommand extends Command
                 $email = $this->ask('Enter the admin email');
             }
 
-            // passwords donot match
+            // Passwords don't match
             if( $password != $confirmPassword){
                 $this->info('Passwords donot match');
                 return;
--- a/src/Commands/AdminCommand.php
+++ b/src/Commands/AdminCommand.php
@@ -129,9 +129,10 @@ class AdminCommand extends Command
                 $email = $this->ask('Enter the admin email');
             }
 
-            // Passwords don't match
+            // Passwords donot match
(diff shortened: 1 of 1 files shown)


emptynick:
Please don't put comments into your changes if you dont have to e.g. explain a complex condition or something.
Otherwise our code would be 90% comments

src/Commands/AdminCommand.php
```diff
@@ -117,12 +122,19 @@ protected function getUser($create = false)
...
+            $confirmPassword = $this->secret('Confirm Password'); // codegeek2014 #2896
 
             // Ask for email if there wasnt set one
             if (!$email) {
                 $email = $this->ask('Enter the admin email');
             }
 
+            // codegeek2014 #2896
+            if( $password != $confirmPassword){
+                $this->info('Passwords donot match');
```
emptynick: Please change this to something like `Passwords don't match`

emptynick (changes requested):

swarnava112:
i have changed the comments.

src/Commands/AdminCommand.php
```diff
@@ -117,12 +122,19 @@ protected function getUser($create = false)
...
+            $confirmPassword = $this->secret('Confirm Password'); 
 
             // Ask for email if there wasnt set one
             if (!$email) {
                 $email = $this->ask('Enter the admin email');
             }
 
+            // passwords donot match
+            if( $password != $confirmPassword){
+                $this->info('Passwords donot match');
```
emptynick: Please change this to something like `Passwords don't match`

emptynick (dismissed):

src/Commands/AdminCommand.php
swarnava112: updated the comment.

swarnava112:
why is the StyleCI check failing? What am I doing wrong?..

fletch3555:
It wants a line before your return.  But you don't need to worry about StyleCI.  It'll fix it itself after this gets merged.

What you do need to do is complete the change requested by @emptynick.  You still have "donot" in your code

src/Commands/AdminCommand.php
```diff
@@ -117,12 +122,19 @@ protected function getUser($create = false)
...
+            $confirmPassword = $this->secret('Confirm Password');
 
             // Ask for email if there wasnt set one
             if (!$email) {
                 $email = $this->ask('Enter the admin email');
             }
 
+            // Passwords don't match
+            if ($password != $confirmPassword) {
+                $this->info('Passwords donot match');
```
fletch3555: The text in this line needs to change, as requested by @emptynick

fletch3555 (dismissed):

src/Commands/AdminCommand.php
```diff
@@ -129,9 +129,10 @@ protected function getUser($create = false)
                 $email = $this->ask('Enter the admin email');
             }
 
-            // Passwords don't match
+            // Passwords donot match
```
swarnava112: changed don't to donot

src/Commands/AdminCommand.php
```diff
@@ -129,10 +129,9 @@ protected function getUser($create = false)
                 $email = $this->ask('Enter the admin email');
             }
 
-            // Passwords donot match
+            // Passwords don't match
             if ($password != $confirmPassword) {
-                $this->info('Passwords donot match');
-                
+                $this->info("Passwords don't match");
```
swarnava112: changed both donot to don't

src/Commands/AdminCommand.php
```diff
@@ -50,6 +50,11 @@ public function handle()
             $this->option('create')
         );
 
+        // the user not returned
+        if (!$user) {
+            exit;
+        }
```
emptynick, line 56: Sorry to bother you again, but what do the above lines do/why are they necessary now?

swarnava112: if the password doesnot match the a null `$user` is returned, hence if `!$user` then exit. And sorry for the late reply..

emptynick: Ah yeah, got it

marktopper (approved):
The entire command could be refactored a bit, but its good for now. Thanks for adding this feature.

PR #2927, opened 2018-04-03
Add preserveFileUploadName also on "File" type

reficul:
This PR continue to use random as default but I think that is better for 1.1 to set original filenames as default and random as option. There is the whole question of “SEO” and the possibility for users to have more recognizable files that I think should lead to reverse things.


marktopper:
Sorry, we are closing issues for `v1.0` do to the release of `v1.1` that is out in a bit.
You are welcome to remake this for `v1.1`.

PR #2928, opened 2018-04-03
Fixes checkboxes not saving the value

Braiani:
This PR will fix the bug when saving an entry that has a checkbox, that always saved as false.

### DIFF
--- a/src/Http/Controllers/ContentTypes/Checkbox.php
+++ b/src/Http/Controllers/ContentTypes/Checkbox.php
@@ -9,6 +9,6 @@ class Checkbox extends BaseType
      */
     public function handle()
     {
-        return (int) ($this->request->input($this->row->field) == 'on') ? true : false;
+        return (int) ($this->request->input($this->row->field) == 'on');
     }
 }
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/ContentTypes/Checkbox.php
```diff
@@ -9,6 +9,6 @@ class Checkbox extends BaseType
      */
     public function handle()
     {
-        return (int) $this->request->input($this->row->field, 0);
+        return (int) ($this->request->input($this->row->field) == 'on') ? true : false;
```
fletch3555: testing with `==` will already return a boolean... why do you need the ternary operator returning `true` or `false`..?

fletch3555 (changes requested):

src/Http/Controllers/ContentTypes/Checkbox.php
Braiani: I changed it...

emptynick:
Do we need to cast it to `int` or is it safe to let the DB decide this?

fletch3555 (approved):

marktopper (approved):

PR #2931, opened 2018-04-04
fix: belongsToMany relationships can now be removed

DiederikvandenB:
Removing all entries of a belongsToMany relationship in voyager results in the field being removed from the request that is to be saved in to the database. Therefore, a relationship cannot be removed, because that would render `$content` as `is_empty` and the current code then removes the `$row`. This fixes that issue.


pamoedo:
It happened to me also. This fixes issues. Thanks for pr!

sitnikovalex:
Thanks for the solution!

marktopper (approved):

PR #2945, opened 2018-04-09
Removed what felt like test variable

fredbradley:
This seemed to be an accident that it was still left there? Especially with the `strtolower()`

Ready to be removed in next update.


bbashy:
Original example;

`Yes, Delete it! post`
to
`Yes, Delete it!`


ref: https://github.com/the-control-group/voyager/blob/1.0/publishable/lang/en/voyager.php#L37

marktopper (approved):

marktopper:
Thanks

PR #2953, opened 2018-04-11
Add classname to view button in browse template

bbrala:
Add a classname to the view button in the browse template to be consistent with the other buttons. This is expecially handy when trying to select the buttons with javascript.

I was selecting a button with javascript because i didn't want to override the whole template, unfortunately selecting the view button is a bit tricky. This makes this a lot easier. Additionally, this makes the classnames consistent with the other buttons.

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -186,7 +186,7 @@
                                                 </a>
                                             @endcan
                                             @can('read', $data)
-                                                <a href="{{ route('voyager.'.$dataType->slug.'.show', $data->{$data->getKeyName()}) }}" title="{{ __('voyager.generic.view') }}" class="btn btn-sm btn-warning pull-right">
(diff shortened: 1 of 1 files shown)


emptynick:
Makes sense to me.
Only thing is that you created your branch from `1.0` and want to merge into `1.x-dev`, thats why its conflicting.
Because we are working on version `1.1` you might want to make your changes in [this line](https://github.com/the-control-group/voyager/blob/1.x-dev/src/Actions/ViewAction.php#L25)

bbrala:
Oops, got confused, quite a few remotes on my repository since i working with @vblinden, will send in a new PR.

emptynick:
You can just push a commit to your fork/branch, it will automatically update here as well.

bbrala:
Since the change is so small, i'll remake to make sure rebasing doesn't cause problems.

PR #2954, opened 2018-04-11
Fix classname for view button in browse template

bbrala:
Add the correct classname to the view button in the browse template to be consistent with the other buttons. This is expecially handy when trying to select the buttons with javascript.

I was selecting a button with javascript because i didn't want to override the whole template, unfortunately selecting the view button is a bit tricky. This makes this a lot easier and makes the classnames consistent with the other buttons.


marktopper (approved):

marktopper:
Thanks

PR #2955, opened 2018-04-11
Add a translation label for each translation

marijnz0r:
This pull request adds a label for each translatable field in a model. The label contains the 2 letters of the language to be translated. This adds a hint for a user to see which fields are to be translated for a model.

### DIFF
--- a/resources/assets/js/multilingual.js
+++ b/resources/assets/js/multilingual.js
@@ -28,211 +28,215 @@
 
     "use strict";
 
-        var pluginName = "multilingual",
-            defaults = {
-                editing:       false,                       // Editing or View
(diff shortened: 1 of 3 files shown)


vblinden:
Looks good to me!

I don't know if you need to commit the `app.js` and `app.css` files. @fletch3555 Can you confirm?

fletch3555:
No, please don't commit the app.js/css files. We have a build too that will auto-compile them after the merge.  Committing them will cause merge conflicts

marijnz0r:
Hi @fletch3555, I updated the pull request to not contain the build files.

marktopper (approved):

PR #2963, opened 2018-04-13
Fix bug, more than one point column

ebelendez:
With this fix tables may have more than one point column and the map will display properly.


marktopper:
Any chance you could look into the merge conflicts?

resources/views/bread/read.blade.php
```diff
@@ -180,4 +180,5 @@
         });
 
     </script>
+    @includeWhen($dataTypeContent->getCoordinates(), 'voyager::formfields.coordinates_js')
```
marktopper: Method `getCoordinates` does not exist according to test `TCG\Voyager\Tests\RouteTest::testGetRoutes`.

> ErrorException: Call to undefined method Illuminate\Database\Query\Builder::getCoordinates() (View: /home/travis/build/the-control-group/voyager/resources/views/bread/read.blade.php)

marktopper (changes requested):

ebelendez:
Sorry is my first PR. Now is working good but there is a problem with 1.x-dev branch, I'm getting a "Permission does not exist" when editing any BREAD. (Even without my changes)

marktopper:
Do you see what permission it complains about?

And can you tell me the exact things you do to replicate this error, then I will look into it 👍

resources/views/bread/read.blade.php
fletch3555: That's because it's acting on a `Builder` object, not a `Model` object.  Likely missing a `->get()` somewhere

ebelendez:
I install Voyager with composer a worked fine.
Then I delete vendor/tcg/voyager directory and did a checkout of voyager and pull 1.x-dev.
When I try to edit BREAD its says "browse_bread" permission does not exits. 
src/Voyager.php line 230
image 001

fletch3555:
@ebelendez, that's not how you should be changing branches.  Instead, run `composer install tcg/voyager:dev-1.x-dev`, and that will fetch the branch instead of the published versions.

As for your issue, did you follow the install steps in our documentation? Specifically, did you run `php artisan voyager:install`?

ebelendez:
Good to know but I got the same results using composer require tcg/voyager:dev-1.x-dev
And yes, I used "artisan voyager:install" just forgot to mention.

marktopper:
Please check your `permissions` table, it seems to complain about missing the `browse_bread` row.

ebelendez:
Yes I know that key does not exist in permissions table but I don't know how to fix that.. Were you able to reproduce the problem?

marktopper:
Have you tried what @fletch3555 said? Running `php artisan voyager:install`?

ebelendez:
I tried with a fresh Laravel install and everything works fine.
The only way I found out to reproduce the problem is copying database/seeds dir from the problematic installation to a new Laravel 5.5 install! 
Any way I also found out that if I delete this dir (in the problematic install) before I run voyager::install everything works right, so I think we can forget about this.
Can we proceed with the merge?

ebelendez:
Where / how can I add this instructions?

**Point type columns and Google Maps**
To show a map in a field value this steps must be taken:
-	Get a key at [Google Developers](https://developers.google.com/maps/documentation/javascript/get-api-key) that is enabled for: Maps and Static Maps.
-	Add 4 variables to .env: 

```
GOOGLE_MAPS_KEY= 
GOOGLE_MAPS_DEFAULT_CENTER_LAT=
GOOGLE_MAPS_DEFAULT_CENTER_LNG=
GOOGLE_MAPS_DEFAULT_ZOOM=
```

-	Create a table with one or more point type columns (include create model). 
-	Add BREAD to the table and select Input Type "Coordinates" for the point type columns.
-	Add Spatial trait and $spatial property to the Model.

```
..
use TCG\Voyager\Traits\Spatial;

class Test extends Model
{
    use Spatial;
    public $spatial = ['pt1', 'pt2'];
}
```

-	The value of $spatial must be an array off the point type columns.
-	Now you can see the map when you Create / Edit / View this fields.

ebelendez:
So.. do I have to do something?

emptynick:
Closing in favor of https://github.com/the-control-group/voyager/pull/4474

PR #2967, opened 2018-04-13
Tagging feature for "belongs to many" relations

amrnn90:
This PR adds an option for belongsToMany relationships that enables the user to create new items directly using select2 and ajax.
Before using this, it's necessary to enable BREAD for the other table and make fields other than "label" nullable or give them default values.


marktopper (approved):

marktopper:
Thanks, works like a charm

PR #2977, opened 2018-04-15
Permission seems wrong

osuagwu:
You have:
$current = $user->id === $model->id;
But why would you expect model id to match user id. Although this is randomly possible it does not mean that the lucky user is the user. It can lead to access being granted to the wrong person.
Is the following what  you meant ?
$current = $user->id === $model->user_id;

### DIFF
--- a/src/Policies/UserPolicy.php
+++ b/src/Policies/UserPolicy.php
@@ -17,8 +17,9 @@ class UserPolicy extends BasePolicy
     public function read(User $user, $model)
     {
         // Does this post belong to the current user?
-        $current = $user->id === $model->id;
-
+        //$current = $user->id === $model->id;//This is bad, and could give away access to a wrong person 
+        $current = $user->id === $model->user_id; //Is this what  you meant?
+        
         return $current || $this->checkPermission($user, $model, 'read');
     }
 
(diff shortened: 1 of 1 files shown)


fletch3555:
This is the UserPolicy, which only applies to "user" BREAD views.  Due to this `$model` will always be the `User` object being accessed (not necessarily the "current" user).  This test is checking if the current user is accessing their own user view(s).  We want them to always be allowed to do that for purposes of the "/admin/profile" view.  I think if you were to test this yourself, you'll find that it throws a "field 'user_id' not found" error.

osuagwu:
I see, but your function description or the way its written do not show what you explained. When using your program we just reuse the functionality you already have instead of making duplicate, so I think it is good for functions to be clear about what they do unless of course you are preventing us from extending you program. Type-hinting the $model as User or doing $current =$user->is($model), would prevent us from misusing your function.

fletch3555:
This is just a policy class following Laravel conventions.  We can't typehint the `$model` parameter, as it won't match the signature of the method it's overriding.  This particular class is specific to Users, which it's reasonable for us to expect you to know.

osuagwu:
OK

PR #1989, opened 2017-10-12
Feature/rtl support v1

mbdwey:
To enable RTL set voyager.multilingual.rtl flag  to true from voyager config file

PR #2993, opened 2018-04-18
fix use of voyager.prefix in AddBreadMenuItem

MrCrayon:
removed hardcoded `admin` as menu in AddBreadMenuItem and role in AddBreadPermission and add configuration
added rtl to voyager_dummy config missed from #1989

**This PR changes two voyager config keys so I think config will need to be republished**

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -174,11 +174,19 @@ return [
     |
     */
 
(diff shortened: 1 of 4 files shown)


marktopper (approved):
Nice improvements 💪

PR #3013, opened 2018-04-20
Use method chaining when interacting with the browser in some tests

cyrildewit:
**Priority:** low

I came across some tests where we are interacting with the browser kit a lot. If we replace all those `$this` with method chaining, it will be easier to understand which actions are taken. The only downside is that it's harder to change the actions very quickly and adding comments to the actions. But how many times are we changing the admin page structure? So I don't think this is a problem.

There are some more tests, but these contains comments by each step. Removing `$this` would only make it worser.

### DIFF
--- a/tests/LoginTest.php
+++ b/tests/LoginTest.php
@@ -13,11 +13,11 @@ class LoginTest extends TestCase
 
     public function testSuccessfulLoginWithDefaultCredentials()
     {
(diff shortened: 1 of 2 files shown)


marktopper (approved):

marktopper:
Thanks 👍

PR #3014, opened 2018-04-20
Fixes User-Role being removed

### LINKED ISSUE #2451: Call to a member function relationLoaded() on null
- Laravel Version: 5.5
- Voyager Version: 1.0.0
- PHP Version: 7.1

### Description:
When going to http://localhost:8000/admin I get this error:
Symfony\Component\Debug\Exception\FatalThrowableError thrown with message "Call to a member function relationLoaded() on null"

Stacktrace:
#0 Symfony\Component\Debug\Exception\FatalThrowableError in C:\xampp\htdocs\celer\vendor\tcg\voyager\src\Traits\VoyagerUser.php:53

### LINKED ISSUE #3010: Can't change admin password.
- Laravel Version: 5.6.*
- Voyager Version: 1.1
- PHP Version: 7.1.3
- Database Driver & Version: mysql 5.7.21

### Steps To Reproduce:
- setup new laravel project
- composer require
- create database
- setup env to connect database
- voyager install --with-dummy
- login with admin@admin
- click profile
- change password, save.

emptynick:
The user-role was removed when a user edited his profile.

Fixes #3010 and fixes #2451

Problem introduced in #2931

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -59,7 +59,7 @@ abstract class Controller extends BaseController
             if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
                 // if the field is a belongsToMany relationship, don't remove it
                 // if no content is provided, that means the relationships need to be removed
-                if (isset($options->type) && $options->type !== 'belongsToMany') {
+                if ((isset($options->type) && $options->type !== 'belongsToMany') || $row->field !== 'user_belongsto_role_relationship') {
                     continue;
                 }
             }


cyrildewit:
~I can confirm that this fix solves the issue! Thanks @emptynick!~

MrCrayon:
It fixes #2451 but it brings back bug #2931 solved: not being able to reset the field.

emptynick:
No, #2931 is still implemented.

MrCrayon:
But it won't work for user_belongsto_role_relationship, try to remove all additional roles after you add one.

marktopper (approved):

AskAlice:
still getting this error, i assume composer doesn't have an updated package?
edit: `composer require tcg/voyager:1.1.x-dev`

marktopper:
@AskAlice Try: `composer require tcg/voyager:1.1-dev`

jpolive:
I can confirm that this change reintroduces the problem in #2931.

The OR in the IF statement at src/Http/Controllers/Controller.php:63 makes it so that any $row that reaches that line goes to `continue;` even when `$options->type` is belongsToMany.

Maybe this exception could be solved with an hidden input for `user_belongsto_role_relationship` in the edit-add view for users instead of adding it to the Controller?

MrCrayon:
@jpolive The problem started in the first place because role field was removed so that users can't change their role, even if that was only a cosmetic change since there is no check on the controller side, so that's still possible but just hidden.
So if you add an hidden field you can solve the problem of role not being removed but still there is no check that it's not being changed. 
I hope my PR #3016 will be considered.

seoeaa:
New installation according to the instructions and when changing the language in the profile error

Symfony\Component\Debug\Exception\FatalThrowableError thrown with message "Call to a member function relationLoaded() on null"

Stacktrace:
#0 Symfony\Component\Debug\Exception\FatalThrowableError in C:\Server\OSPanel\domains\laravel.dev\vendor\tcg\voyager\src\Traits\VoyagerUser.php:121

DiederikvandenB:
Just wanted to drop a line to say sorry for introducing this bug in #2931. Normally I'd fix it myself, but since I am on holiday without internet for the coming two weeks I don't think I can be of much use.

PR #3016, opened 2018-04-20
Fix user edit removing roles

### LINKED ISSUE #2476: Unable to delete last BelongsToMany relation in BREAD
- Laravel Version: 5.5.28
- Voyager Version: 1.0.12
- PHP Version: 7.0.23
- Database Driver & Version: Server version: 5.7.18-16 Percona Server (GPL), Release 16, Revision d7301f8

Description:
Bug. It is impossible to delete last BelongsToMany relation between objects using BREAD.
The one row in pivot table always stay.

Steps To Reproduce:
1. Create table Items ( id, name, created_at, updated_at)
2. Create table tags (id, name, created_at, updated_at)
3. Create pivot table  item_tags (item_id, tag_id)
4. Create BREAD for Items
5. Create BelongsToMany relation in items BREAD
6. Add one  item and  add tag to that item 
8. Edit item and try to delete tag

Voyager shows a Sucess result, but - the tag was not remove.

### LINKED ISSUE #3140: All additional roles can't be removed
- Laravel Version: 5.5.40
- Voyager Version: 1.1
- PHP Version: 7.0.28
- Database Driver & Version: 5.7.22

### Description:

When I try to remove all additional roles from an user, the additional roles remain the same. When I removed one additional role but not all, the changes are storaged.

### Steps To Reproduce:
Remove all additional roles from an user.
Save changes.
Additional roles remain the same.

### LINKED ISSUE #3230: Remove last value on belongsToMany relationship
- Laravel Version: 5.5.40
- Voyager Version: 1.1.1
- PHP Version: 7.0
- Database Driver & Version: mysqli / mysqlnd 5.0.12-dev

### Description:
If you delete all values from the belongToMany field, one of them is not deleted.

### Steps To Reproduce:
Create BelongsToMany Relationship in some Bread.
Add two relations in this entity, save.
Open it, remove relations. Save.
One relations don't removed.

### Cause

vendor/tcg/voyager/src/Http/Controllers/Controller.php, line 62.
`
 if ((isset($options->type) && $options->type !== 'belongsToMany') || $row->field !== 'user_belongsto_role_relationship') `

Because of the second part of the expression, the condition is always true.
remove this
`
 $row->field !== 'user_belongsto_role_relationship'`
from condition - relations work fine.

WTF is this condition here?

### LINKED ISSUE #3350: BelongsToMany select multiple bug
## Version information

 - Laravel: v5.5.*
 - Voyager: v1.1.2
 - PHP: 7.2
 - Database: [PostgreSQL] [10.4]

## Description

When using BelongsToMany relationship, the select multiple TextBox that automatically appears in the BREAD "Edit view" doesn't "update" if the previously selected options are all removed. My guess for why this happens is that the code simply picks whatever options you select for the update while ignoring a situation whereby the user decides to remove all previously selected options. Is there a way to fix this please?

MrCrayon:
Fix user being able to edit its own role without edit users permission
Currently roles are only hidden but there is no permission check if user sends role_id.

This fix might be a bit hacky but I think it was important to fix ASAP on Voyager this big security issue that became even worse in 1.1 since being permissions hidden on profile edit gives the false confidence that it's secured.
Also I tried to find a solution that would have the less impact to the current structure.

Any suggestion  for a better solution is welcome.

Fixes #3140, fixes #3424, fixes #2476, fixes #3230, fixes #3350
Closes #3177, closes #3112


cyrildewit:
I think you forgot to update the tests.

MrCrayon:
@cyrildewit yes you are right, I tested it all on a fresh new voyager installation but of course tests will need to be changed too since some routes changed.
I'll have a look at it.

MrCrayon:
@emptynick any opinion about this PR?
This one doesn't have any problem about updating roles afaik and it also solve the security problem that I think it's quite important.

resources/views/users/edit-add.blade.php
```diff
@@ -60,7 +71,7 @@
                                 <input type="password" class="form-control" id="password" name="password" value="" autocomplete="new-password">
                             </div>
 
-                            @can('editRoles', $dataTypeContent)
+                            @if(auth()->user()->hasPermission('edit_users'))
```
fletch3555: You should be able to still use `@can()`.  Just pass `auth()->user()` instead of `$dataTypeContent`

routes/voyager.php
```diff
@@ -129,6 +129,16 @@
             Route::post('/', ['uses' => $namespacePrefix.'VoyagerCompassController@index',  'as' => 'post']);
         });
 
+        //Profile
+        Route::group([
+            'as'     => 'users.',
+            'prefix' => 'profile',
+        ], function () use ($namespacePrefix) {
+            Route::get('/{id}', ['uses' => $namespacePrefix.'VoyagerController@showProfile', 'as' => 'showProfile']);
```
fletch3555: The inclusion of `id` seems unnecessary.  `profile` implies `my user account`.  Additionally, you only allow access to the current user in the VoyagerController methods you've added

src/Http/Controllers/VoyagerController.php
```diff
@@ -68,8 +68,54 @@ public function upload(Request $request)
         return "<script> parent.helpers.setImageValue('".Voyager::image($fullFilename)."'); </script>";
     }
 
-    public function profile()
+    public function profile(Request $request)
```
fletch3555: what's the purpose of this method if you have `showProfile` as well?

src/Http/Controllers/VoyagerController.php
```diff
@@ -68,8 +68,54 @@ public function upload(Request $request)
...
+            abort(404);
+        }
+    }
+
+    public function editProfile(Request $request, $id)
+    {
+        if ($id == auth()->user()->id) {
+            $voyagerBaseController = new VoyagerBaseController();
+
+            return $voyagerBaseController->edit($request, $id);
```
fletch3555: This is SUPER hacky.... Not sure I can condone it

src/Http/Controllers/VoyagerController.php
```diff
@@ -68,8 +68,54 @@ public function upload(Request $request)
...
+        } else {
+            abort(404);
+        }
+    }
+
+    public function updateProfile(Request $request, $id)
+    {
+        if ($id == auth()->user()->id) {
+            // If can't edit users can't change his own role but need to preserve roles set
+            if (!auth::user()->hasPermission('edit_users')) {
```
fletch3555: I'd rather see permissions checks use the Laravel Authorization functionality than direct permissions checks.  Keeps it consistent

src/Http/Controllers/VoyagerController.php
```diff
@@ -68,8 +68,54 @@ public function upload(Request $request)
...
+                $role = Voyager::model('Role');
+                $roles = auth::user()->belongsToMany($role, 'user_roles')
```
fletch3555: Why are you using `belongsToMany()` here....? That's usually reserved for use in a method on the model class itself (which does exist by the way)

src/Http/Controllers/VoyagerController.php
```diff
@@ -68,8 +68,54 @@ public function upload(Request $request)
...
+
+                $params = $request->all();
+                $params['role_id'] = Auth::user()->role_id;
+                $params['user_belongstomany_role_relationship'] = $roles;
+
+                $request->replace($params);
+            }
+            $voyagerBaseController = new VoyagerBaseController();
+
+            return $voyagerBaseController->update($request, $id);
```
fletch3555: Again, not really sure I like the idea of calling out to another controller like this.

fletch3555 (changes requested):

MrCrayon:
Thanks for all your suggestions, appreciated.
I'm still quite new on Laravel so I'm still learning the right way to do things.

src/Http/Controllers/VoyagerController.php
MrCrayon: I guess a cleaner option would be to create a VoyagerProfileController that extends VoyagerBaseController and then use parent::method.
Actually in my application I put all that in UserController.

src/Http/Controllers/VoyagerController.php
MrCrayon: You are right, it was much easier, I thought relationships were not defined and I didn't check.
I took the code from relationship.blade.php

MrCrayon:
I think it's much better now :grin: 
For a manual update _\TCG\Voyager\Http\Controllers\VoyagerUserController_ needs to be added as Users Bread Controller.
I hope I picked the right way to check permission, voyager docs refers to
```Voyager::can('edit_posts');```
but I saw that has been deprecated.

**Comment by cyrildewit** [comment]
@MrCrayon yeah, I saw it too. Maybe we should remove them from the documentation or add something about that they are going to be removed in version x.x.x.

**Comment by emptynick** [comment]
The problem here is, that you updated the Seeder so that the `users` BREAD uses your Controller.
I would consider this as a breaking change.

**Comment by MrCrayon** [comment]
I have to test it but I think that if you apply the changes and don't update Controller used for User in BREAD it's simply not changing anything, so I don't know how breaking that would be.
Instead if you want to make sure that the new UserController is used we could change it in routes.

I admit I don't know much about releasing policies my point of view is that since this is a security issue the sooner is patched the better and I don't think the problem is going to be solved without a custom controller for User.

**Comment by carlost89** [comment]
Hi, I need help

In order to solve this problem I created a controller called UserController that extends VoyagerBaseController, then I overrode the update method just adding this code:

`if(!Auth::user()->can('editRoles', $data)) {
                $request->request->add(['role_id' => $data->role_id]);
                $request->request->add(['user_belongstomany_role_relationship' => $data->roles()->pluck('id')->toArray()]);
            }`

after

`$data = call_user_func([$dataType->model_name, 'findOrFail'], $id);`

and I was able to solve the problem.

Now I see that you're working in something very similar from what I did. The question is ¿How can I implement the changes you're doing here in order to solve the problem?

Thank you.

File: publishable/database/seeds/DataTypesTableSeeder.php [hunk 8]
```diff
@@ -19,7 +19,7 @@ public function run()
                 'icon'                  => 'voyager-person',
                 'model_name'            => 'TCG\\Voyager\\Models\\User',
                 'policy_name'           => 'TCG\\Voyager\\Policies\\UserPolicy',
-                'controller'            => '',
+                'controller'            => '\TCG\Voyager\Http\Controllers\VoyagerUserController',
```
fletch3555: This can't be right, look at the lines above it.  The backslashes need to be escaped

fletch3555 (dismissed):

emptynick (approved):

MrCrayon:
@fletch3555 @emptynick in your merge https://github.com/the-control-group/voyager/pull/3016/commits/dc2818e62e76e78809484c07cc483c538f7e1c71 was removed, introducing back the  vulnerability that that commit was fixing.

Without that check a malevolent user can change is main role while editing profile tampering with html form, I just tried in a fresh installation.

emptynick:
What did you try?
I just removed the `@can('editRoles', ...)` stuff from `resources/views/users/edit-add.blade.php`, changed all my roles and hit save. It wont update any of it because of
https://github.com/the-control-group/voyager/blob/4ee291ff24a43a56bec5009601f7ce874b725838/src/Http/Controllers/VoyagerUserController.php#L19-L24
Actually, this code survived a pen-test:
https://compass-security.com/fileadmin/Datein/Research/Advisories/CSNC-2018-039_voyager_privilege_escalation.txt

MrCrayon:
Check this https://github.com/the-control-group/voyager/commit/dc2818e62e76e78809484c07cc483c538f7e1c71 commit, that part was removed.
If I go in edit profile as normal user and I add a field with name="user_belongsto_role_relationship" I can change role_id.

emptynick:
Is the controller for the user-bread set to `TCG\Voyager\Http\Controllers\VoyagerUserController`?

MrCrayon:
Yes, I did it on a fresh install. You can try: inspect with chrome and add the field.

emptynick:
Which html code/parameters do you inject exactly?

MrCrayon:
```html
<input class="form-control" name="user_belongsto_role_relationship">
```

emptynick:
https://github.com/the-control-group/voyager/commit/7bb20732365222c5704d571379301919e97267e5

PR #3020, opened 2018-04-20
Fix user role being removed #2

emptynick:
This "reverts" my own fix #3014 
And implements another condition which tests if `role_id` is null. If so, it will use the old one.
Not sure if its safe to check if `slug == users`

Tagging @MrCrayon @cyrildewit and @HuntersSerge


marktopper:
This seems like a better solution! 👍

MrCrayon:
I have to test it but I think this will still remove the additional roles that are hidden in profile editing.

emptynick:
Thats true. Lets see

MrCrayon:
IMHO there is no way to fix it as it is without some more radical change as in my PR.

Right now with the current UserPolicy I can't change my role even if I can create and change roles and I can edit users and that for me already doesn't make much sense.

Also I don't think it's a good idea to check _user_belongstomany_role_relationship_ in insertUpdateData function because it will leave a specific behavior that might not be needed or wanted extending User and Role models/controllers.

emptynick:
Ok I think this is better now.
I reverted my old changes (belongsToMany can be removed again) and moved `data->save()` after relationship-handling.
Roles now get reverted when the user-model is saving, when the user-id == data-id and the user can't change his own roles.

@MrCrayon 
> Also I don't think it's a good idea to check user_belongstomany_role_relationship in insertUpdateData function because it will leave a specific behavior that might not be needed or wanted extending User and Role models/controllers.

[What?](https://github.com/the-control-group/voyager/pull/3016/files#diff-7a84f7979dcdf410a58625359347974aR26)

carlost89:
Hi, I need help

In order to solve this problem I created a controller called UserController that extends VoyagerBaseController, then I overrode the update method just adding this code:

`if(!Auth::user()->can('editRoles', $data)) {
                $request->request->add(['role_id' => $data->role_id]);
                $request->request->add(['user_belongstomany_role_relationship' => $data->roles()->pluck('id')->toArray()]);
            }`

after

`$data = call_user_func([$dataType->model_name, 'findOrFail'], $id);`

and I was able to solve the problem.

Now I see that you're working in something very similar from what I did. The question is ¿How can I implement the changes you're doing here in order to solve the problem? or ¿When do you think this is going to be approved to merge?

Thank you.

emptynick:
Clsoing in favor of #3177

PR #3022, opened 2018-04-20
Don't show a widget if it's not accessible for the current user

### LINKED ISSUE #2248: Override widgets upon user's role
- Laravel Version: 5.5.21
- Voyager Version: 1.0.7
- PHP Version: 7.1.10
- Database Driver & Version:

### Description:

How (and where) can I set that?
\Config::set('voyager.dashboard.widgets', getWidgets() );

### LINKED ISSUE #2275: Dashboard shows things that shouldn't be available to Role
- Laravel Version: 5.5.21
- Voyager Version: 1.0
- PHP Version: 7.1.10
- Database Driver & Version: MySQL 14.14

I have a role that only should allow access to Posts. But I have to also allow the role to be able to "Browse Admin". So when that user logs in to the dashboard they see this:

screen shot 2017-11-22 at 2 54 10 pm

They shouldn't see the Users or Pages section at all. It doesn't seem like the dashboard is respecting the role privileges. If either of those is clicked, they get this:

screen shot 2017-11-22 at 2 54 01 pm

cyrildewit:
**Priority:** medium

I have been working the whole day on a fix for #2275. I've tried everything, but this is the only solution that I can find.

Currently, if a user doesn't have the permission to browse the `posts` page for example, he can still see the `PostDimmer` widget. This pull request will fix that. It removes the widget from the `voyager::dimmers` group, if the user can't access it.

I decided not to wrap the `dimmer.blade.php` with `@can('read', $model)...@endcan`, because it would double check the permission. If you think, that this should be added, I can add it.

~I choosed to merge this pull request into the `v1.x-dev` branch, because now the widgets requires an `getRelatedModel` method.~

Some feedback would be highly apprenticed! Because I believe this PR is incomplete.

---

Fixes #2275
Fixes #2248

### DIFF
--- a/resources/views/dimmers.blade.php
+++ b/resources/views/dimmers.blade.php
@@ -3,10 +3,10 @@ $dimmers = \Arrilot\Widgets\Facade::group('voyager::dimmers');
 
 // Loop over the defined dashboard widgets. Get the related model and remove the
 // widget from the group if the current user hasn't the permission to read it.
-foreach (config('voyager.dashboard.widgets') as $widgetClass) {
+foreach (config('voyager.dashboard.widgets', []) as $widgetClass) {
     $relatedModel = app($widgetClass)->getRelatedModel();
 
-    if (! Auth::user()->can('read', $relatedModel)) {
+    if (! Auth::user()->can('browse', $relatedModel)) {
         $dimmers->removeByName($widgetClass);
     }
 }
--- a/resources/views/dimmers.blade.php
+++ b/resources/views/dimmers.blade.php
@@ -1,7 +1,17 @@
 @php
 $dimmers = \Arrilot\Widgets\Facade::group('voyager::dimmers');
-$count = $dimmers->count();
 
+// Loop over the defined dashboard widgets. Get the related model and remove the
+// widget from the group if the current user hasn't the permission to read it.
+foreach (config('voyager.dashboard.widgets') as $widgetClass) {
+    $relatedModel = app($widgetClass)->getRelatedModel();
+
+    if (! Auth::user()->can('read', $relatedModel)) {
+        $dimmers->removeByName($widgetClass);
+    }
+}
+
+$count = $dimmers->count();
 $classes = [
     'col-xs-12',
     'col-sm-'.($count >= 2 ? '6' : '12'),
--- a/resources/views/dimmers.blade.php
+++ b/resources/views/dimmers.blade.php
@@ -1,12 +1,9 @@
 @php
 $dimmers = \Arrilot\Widgets\Facade::group('voyager::dimmers');
 
-// Loop over the defined dashboard widgets. Get the related model and remove the
-// widget from the group if the current user hasn't the permission to read it.
+// Remove all the inaccessible widgets from the `voyager::dimmers` group.
 foreach (config('voyager.dashboard.widgets', []) as $widgetClass) {
-    $relatedModel = app($widgetClass)->getRelatedModel();
-
-    if (! Auth::user()->can('browse', $relatedModel)) {
+    if (!app($widgetClass)->isAccessible()) {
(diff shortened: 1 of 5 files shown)


resources/views/dimmers.blade.php
```diff
@@ -1,7 +1,17 @@
 @php
 $dimmers = \Arrilot\Widgets\Facade::group('voyager::dimmers');
-$count = $dimmers->count();
 
+// Loop over the defined dashboard widgets. Get the related model and remove the
+// widget from the group if the current user hasn't the permission to read it.
+foreach (config('voyager.dashboard.widgets') as $widgetClass) {
+    $relatedModel = app($widgetClass)->getRelatedModel();
+
+    if (! Auth::user()->can('read', $relatedModel)) {
```
marktopper: How about making the permission check being inside the dimmer? Making the dimmer have the logic. Makes it easier to overwrite for specific cases.

marktopper (changes requested):

resources/views/dimmers.blade.php
cyrildewit: I already tried that, but the problem is. We can't display the widget by condition. When adding a simple `@can` check to dimmer.blade.php the grid will look like: `dimmer 1 -- empty collumn -- dimmer 3` for example. I also tried to use the `->wrap()` method to only add a `div` if the widget is accessible, but this method isn't passing the widget class throught. I could do a pull request to `arrilot/laravel-widgets` if this fixes our problem.

MrCrayon: I think since widgets give a count and are linked to the listing page browse permission should be checked instead of read.

cyrildewit: Oo ##&@*. Good catch!

cyrildewit:
~I think that I have better solution for this. I only have to work it out. I will send a pull request soon.~

resources/views/dimmers.blade.php
marktopper: I don't see the problem, now you have

```
$relatedModel = app($widgetClass)->getRelatedModel();
```

Then your could have something like this instead:
```
if (!app($widgetClass)->accessable()) {
    $dimmers->removeByName($widgetClass);
}
```

`accessable` is just a method you could create, or call it whatever.

cyrildewit: Ahh, I misunderstood you. Yes, that approach looks much cleaner! Nice!

Now I have this:

```php
/**
 * Determine if the widget is accessible.
 *
 * @return bool
*/
public function isAccessible()
{
    return \Auth::user()->can('browse', Voyager::model('Page'));
}
```

And in `dimmers.blade.php`:

```php
$dimmers = \Arrilot\Widgets\Facade::group('voyager::dimmers');

// Remove all the inaccessible widgets from the `voyager::dimmers` group.
foreach (config('voyager.dashboard.widgets', []) as $widgetClass) {
    if (!app($widgetClass)->isAccessible()) {
        $dimmers->removeByName($widgetClass);
    }
}
```

Alternatively, we could check the existance of the `isAccessible` method before calling it, but this could lead to unexpected behaviour.

cyrildewit:
Do we need an interface for the widgets? So that it will be required to have an `isAccessible()` method. Or should we create an abstract class that can be used as base class.

fletch3555:
That really depends on a few things.  If the widgets are only ever loaded explicitly (i.e. not loaded through the DI container), then either is probably fine and personal preference can prevail.  If the functionality of this common method is the same, then I would prefer a parent class.  If it will likely differ in each class, then I'd prefer an interface.

cyrildewit:
The content of `isAccessible` is always different. User's have to add there own check.

If they want to show it nomatter what, they can return the value `true`.

cyrildewit:
Is this still an appropriate fix for #2275? Or can I close this in favor of #3058!

PR #3031, opened 2018-04-21
upload method modified to upload image with original name and no white space

jeancarlosnovaes:
This is important for SEO and was a problem that I wanted to solve, so I think it can be useful for other developers.

upload-file-image
code


emptynick:
You deleted your branch or fork.
So I'm going to close this one. Also because there is https://github.com/the-control-group/voyager/pull/3378 where the behaviour is configurable.

PR #3034, opened 2018-04-22
WIP: File picker using media manager

lucentlogix:
Example implementation on post add-edit.

I'm guessing you guys already have a road-map for this but if not then maybe this can be a start. This only works for single image fields.


lucentlogix:
If you're wondering, ebspeter is my work account and lucentlogix my private.

I wasn't sure what branch to issue the PR to, but then i looked at some recently merged and understood that i probably had it right from the beginning. 

Some more information on how it works:

_Image widget on Post add-edit_
image

_On click the media manager open in a modal, double click an image to select and close modal_
image

lucentlogix:
@fletch3555 @emptynick 

It would be of great value to hear your thoughts about this approach. 

I'll also need support for multiple images for my use case, if you agree then i'll work on expanding this to also cover that. Else i guess i'll just keep it custom for the app i'm building. 

I really like this project and I'm happy if i can contribute!

prdpspkt (dismissed):
Nice Feature, It is most wanted one.

emptynick:
Sorry for the late reply.
If you need help for multiple images, I think its best to talk in Slack again.

@prdpspkt Please dont approve things because you like them. Approving means you tested them and they are working fine, but this is still work in progress.

SanjinOsmanbasic:
I really think that this feature would upgrade Voyager and increase its usability.

emptynick:
Superseeded by https://github.com/the-control-group/voyager/pull/3889

PR #3041, opened 2018-04-23
Cannot use TCG\Voyager\Facades\Voyager as Voyager

zanysoft:
we facing error:
Cannot use TCG\Voyager\Facades\Voyager as Voyager because the name is already in use

Solution:
Remove "use TCG\Voyager\Facades\Voyager;"  (you no need to declare because both files are in same directory)
or
use as " use TCG\Voyager\Facades\Voyager as NewAlias; "

### DIFF
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -4,7 +4,7 @@ namespace TCG\Voyager;
 
 use ArrayAccess;
 use Illuminate\Database\Eloquent\Model;
-use TCG\Voyager\Facades\Voyager;
+//use TCG\Voyager\Facades\Voyager;
 
 class Translator implements ArrayAccess
 {


src/Translator.php
```diff
@@ -4,7 +4,7 @@
 
 use ArrayAccess;
 use Illuminate\Database\Eloquent\Model;
-use TCG\Voyager\Facades\Voyager;
+//use TCG\Voyager\Facades\Voyager;
```
marktopper, line 7: The Voyager facade is used, so you cannot just remove it.

marktopper (changes requested):

src/Translator.php
fletch3555: Also, a pull request shouldn't claim something is unused and comment it out.  Git history exists for a reason, just remove the line.  In this case, however, you can't remove the line (or even comment it out) because it's needed like @marktopper said

fletch3555:
I don't understand what the issue is or how this fixes it. As mentioned in the comments, the `use` statement is definitely needed.

cyrildewit:
@fletch3555 I am facing this error when running the test suite of v1.1. Strange..

fletch3555:
Ah, I see the problem.  Yeah, can't `use ......\Voyager` because `[currentNamespace]\Voyager` already exists.  Removing it isn't correct, but you CAN rename it (`use TCG\Voyager\Facades\Voyager as VoyagerFacade`)

MrCrayon:
~~Actually since Voyager is in the same namespace there is no need for the _use_ statement and can (/should?) be removed.~~

~~Same reason why in VoyagerController there is no:~~
```php
use TCG\Voyager\Http\Controllers\Controller
```

fletch3555:
@MrCrayon, it's not in the same namespace though.  The service class used by the facade is; but the facade itself is not

MrCrayon:
@fletch3555 my bad, I need another coffee :grinning: 
Then yes using _VoyagerFacade_ is the right thing.

fletch3555:
Closing in favor of #3082

PR #3044, opened 2018-04-23
Swedish translation

jannejava:
Saw that there is not a Swedish translation of Voyager. Needed one for me so maybe someone finds this useful.


marktopper (approved):

marktopper:
Thanks

PR #3046, opened 2018-04-23
Made the autoloading of migrations optional

JapSeyz:
Adding the autoloading of the migrations blacklists certain classes even if the multilingual functions are not used.

I use Laravel Translation manager for a part of the application, and since I cannot change the name of the table/class from Voyager anymore, it throws an error because the "CreateTranslationsTable" class already exists inside your autoloaded vendor-migrations directory.

It breaks backwards compatibility of any app that already used that class-name, as I had to go research around to problem-solve why my migrations suddenly wouldn't run.

### DIFF
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -97,7 +97,7 @@ class VoyagerServiceProvider extends ServiceProvider
 
         $this->loadTranslationsFrom(realpath(__DIR__.'/../publishable/lang'), 'voyager');
 
-        if(config('voyager.load_migrations') {
+        if(config('voyager.load_migrations')) {
             if (config('app.env') == 'testing') {
                 $this->loadMigrationsFrom(realpath(__DIR__.'/migrations'));
             }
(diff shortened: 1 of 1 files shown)


src/VoyagerServiceProvider.php
```diff
@@ -97,11 +97,13 @@ public function boot(Router $router, Dispatcher $event)
 
         $this->loadTranslationsFrom(realpath(__DIR__.'/../publishable/lang'), 'voyager');
 
-        if (config('app.env') == 'testing') {
-            $this->loadMigrationsFrom(realpath(__DIR__.'/migrations'));
-        }
+        if(config('voyager.load_migrations')) {
```
fletch3555: The default value should be true, else it's a breaking change, and breaks tests

fletch3555 (changes requested):

src/VoyagerServiceProvider.php
JapSeyz: Has been addressed in the newest commit

JapSeyz:
@fletch3555 I'll limit my composer.json to the version before this was introduced, could you ping me when/if this is merged, or I need to change anything else?

fletch3555:
I'm not sure how I feel on this one.  On one hand, you can easily just copy the migration to your app and leave it blank (effectively nullifying it's behavior).  On the other, I'm not opposed to adding flexibility.  In either case, we'll either need the config added to the `voyager.php` config file, or noted in the docs.  Otherwise, nobody will know it exists.

emptynick:
This was implemented in https://github.com/the-control-group/voyager/pull/3584 and is configurable.
It was merged into 1.x

PR #3048, opened 2018-04-23
admin role removed

Goniss:
I know that this problem was discussed before and I tried several solutions but none worked for me. This problem occurs every time I edit my admin profile and like the other the role_id is set to NULL, so each time I edit the profile, I have to go change the id_role in phpMyadmin. 
So is there a fix ?
Thanks.


fletch3555:
This should be an issue, not a PR... also, this has already been fixed (#3014) and is pending a new release

PR #3058, opened 2018-04-24
Don't show a widget if it's not accessible for the current user #2

### LINKED ISSUE #2248: Override widgets upon user's role

### LINKED ISSUE #2275: Dashboard shows things that shouldn't be available to Role

cyrildewit:
### Summary

All users can see the dashboard widgets currently, even when they don't have the permission to browse the widgets content.

At the moment, all dashboard widgets are registered inside the `VoyagerDummyServiceProvider`, but because the user object will always stay null, we can't check for access.

What I did, is adding a method called `shouldBeDisplayed` to the widget classes. Developers can do all kind of checks over here. For our need we can add a check for the `browse` permission. Then I added the `dimmers` method to the `Voyager` class. This will loop over the configured dashboard widgets and calls this method before registering it. A widget will be skipped if the method doesn't exists.

My first attempt: #3022

---

Fixes #2275
Fixes #2248


fletch3555 (changes requested):
You're going to have to fix the tests to make TravisCI happy.  The current failures are because the test user doesn't have permissions, since they weren't needed for the dashboard page.  You'll have to update the tests to assign the appropriate permissions so the test user can see all 3 widgets.  Also, I'd appreciate it if you added a new test that verifies you CAN'T see it when you don't have permissions.

cyrildewit:
I have changed the tests again.

It's currently as follows: I've created a unit test that will test the `dimmers` method on specific behaviour. For example: if `shouldBeDisplayed` returns false, it should not be inside the returned collection.

I also created a feature test that will make sure these widgets are being displayed on the dashboard page. It will also test each widget on access. A user without the `browse_pages` permissions, shouldn't see the `PageDimmer` widget.

I also removed the test that will verify if the `display()` method renders the widgets, because `arrilot/laravel-widgets` is already testing it over [here](https://github.com/arrilot/laravel-widgets/blob/master/tests/WidgetGroupTest.php#L21). Beside that, we should be able to change the blade partial without having to update the`rendered_widgets.html` file in my opinion.

cyrildewit:
@fletch3555 what's your opinion about this: currently if the method `shouldBeDisplayed` doesn't exists, the widget will not be registerd.

Or do you prefer registering the widget anyway?

fletch3555:
Can we have a "BaseDimmer" that contains a "return true" default for that method? Then all dimmers will have it.  We'll just have to add that to the docs.

fletch3555 (dismissed):

fletch3555 (approved):

fletch3555:
Bumping to `1.x` for a future v1.2 release since the new BaseDimmer and function call would break custom dimmers that may exist.

Issue #3059, opened 2018-04-24
Saving settings without uploading an image 


jannejava:
- Laravel Version: 5.6.17
- Voyager Version: 1.1
- PHP Version:7.2
- Database Driver & Version:MySQL

### Description:

I think a may have found a bug in the last release.

Settings of type image in /admin/settings get a NULL-value if no image is uploaded but other values are updated.

Since there is no upload the field for eg site_logo never get posted, and since it's not existing in the request it gets emptied somehow.

Regards,

Jan

### Steps To Reproduce:
1. Upload a site.logo under /admin/settings
2. Click submit
3. When page reloaded, click submit again
4. Uploaded site.logo value has disappeared.

bryanezavala:
**Still does not solve this error?**

emptynick:
This PR is not merged yet, but does certainly fix the problem.

jheckizon:
Same issue with BREAD

When saving the data, the file was always removed if there is no attached to the file upload field.

## Before updating

![save-1](https://user-images.githubusercontent.com/8746166/40155955-1a97f46a-59c9-11e8-9b54-b9a298f1a579.JPG)
![save-2](https://user-images.githubusercontent.com/8746166/40155957-1acacc8c-59c9-11e8-94bb-d5ec9f9dd82f.JPG)

## After updating

![after-update-1](https://user-images.githubusercontent.com/8746166/40155958-1af60e1a-59c9-11e8-818c-9fb54aed586a.JPG)
![after-update-2](https://user-images.githubusercontent.com/8746166/40155959-1b258a5a-59c9-11e8-8704-c752f2aa0aac.JPG)

fletch3555:
Closed in #3060

pladgepeople:
Still does not solve this error?

fletch3555:
If you have issues, please open a new issue or ask in our Slack channel

PR #3060, opened 2018-04-24
Attempt to fix issue #3059

jannejava:
If no image is uploaded the setting’s value become null and overwrites the previous value.


src/Http/Controllers/VoyagerSettingsController.php
```diff
@@ -84,18 +84,24 @@ public function update(Request $request)
...
+                continue;
+            }
+
             $content = $this->getContentBasedOnType($request, 'settings', (object) [
                 'type'    => $setting->type,
-                'field'   => str_replace('.', '_', $setting->key),
+                'field'   => $field,
                 'details' => $setting->details,
                 'group'   => $setting->group,
             ]);
```
emptynick: Move your condition from line 91 to line 101 and use
`if ($setting->type == 'image' && $content == null)`
This makes everything else unnecessary.

emptynick (changes requested):

jannejava (commented):
Of course, I see what you mean. Should I do another pull request?

emptynick:
You just can push to your branch, it will update here automatically.

jannejava:
Did the update of the pull request work?

fletch3555:
Yes it did.

Also, please try to avoid committing whitespace changes (new lines, indentation/spacing changes, etc).  All they do is make the diff harder to read, and make StyleCI revert them after this gets merged.  I understand some IDEs do this automatically, but you can just be more selective in what you commit (e.g. you can select specific lines instead of entire files).

jannejava:
Sorry for that! It must be php-cs-fixer, I have it set on PSR-2 but you are maybe using Symfony style? Anyway, I will try to disable it for this project. 

Do you have any idea if this would fix the disappearing images on the settings page? (it’s a little annoying, when the images disappear ;-))

emptynick:
Yep thats working

emptynick (approved):

cyrildewit (approved):
Looks good to me!

PR #3063, opened 2018-04-25
Fix BelongsToMany relationships

hojgr:
This commit fixes belongsTo relationships. 

Without this change, all belongsTo relationships are never persistet, because they are ignored (condition on line 63. the key in input is "*_belongsto_*_relationship", but `$row->field` is target column name, i.e. `user_id`.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -57,7 +57,7 @@ abstract class Controller extends BaseController
             if ($row->type == 'relationship' && $options->type != 'belongsToMany') {
                 $row->field = @$options->column;
             }
-            
+
             // if the field for this row is absent from the request, continue
             // checkboxes will be absent when unchecked, thus they are the exception
             if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -54,21 +54,17 @@ abstract class Controller extends BaseController
         foreach ($rows as $row) {
             $options = json_decode($row->details);
 
-            // if the field for this row is absent from the request, continue
-            // checkboxes will be absent when unchecked, thus they are the exception
-            if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
-                // if the field is a belongsToMany relationship, don't remove it
-                // if no content is provided, that means the relationships need to be removed
-                if ((isset($options->type) && $options->type !== 'belongsToMany') || $row->field !== 'user_belongsto_role_relationship') {
-                    continue;
-                }
-            }
-
-            $content = $this->getContentBasedOnType($request, $slug, $row, $options);
-
             if ($row->type == 'relationship' && $options->type != 'belongsToMany') {
                 $row->field = @$options->column;
             }
+            
+            // if the field for this row is absent from the request, continue
+            // checkboxes will be absent when unchecked, thus they are the exception
+            if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
+                continue;
+            }
+
+            $content = $this->getContentBasedOnType($request, $slug, $row, $options);
 
             /*
              * merge ex_images and upload images


emptynick:
We cant accept that for now.
Try to update your own profile with the changes you made, I think it will fail again.

hojgr:
I'm sorry, I don't understand what the issue is. Can you please elaborate?

emptynick:
You removed the condition check implemented [here](https://github.com/the-control-group/voyager/pull/3014)

hojgr:
Yes, I did, as it is not necessary anymore. The condition's purpose was to attemt to fix belongsTo relationships not working. But the fix didn't work for all use cases of belongsTo.

My proposed change fixes it for all belongsTo relationship by performing the `$row->field` mutation before checking if the parameter is present in request - the mutation was taking place too late for belongsTo relationship to work. Now it happens just in time.

jpolive:
The change to 
```
if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
    continue;
}
```
will break the user role issue again #2451, it is an exception to how the relation should behave.
It will also break resetting belongsToMany relationships, they need to be set to null further down in the loop, they can't `continue`.

Can I suggest leaving the user role exception there? Something like
```
if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
    // if the field is a belongsToMany relationship, don't remove it
    // if no content is provided, that means the relationships need to be removed
    if (isset($options->type) && $options->type !== 'belongsToMany') {
        continue;
    }
    // if the field is related to user role, skip it
    if ($row->field === 'user_belongstomany_role_relationship' || $row->field === 'role_id') {
        continue;
    }
}
```
but keeping the other change you made.

hojgr:
@jpolive Are you sure that it will break user roles? In an application where this patch has been applied *all relationships* work. That includes changing user roles - I just checked that.

That exception there is necessary, to fix user roles **when belongsTo relationships are broken**. This fix fixes **broken belongsTo relationships**, so it fixes the user role issue, too, thus eliminating the need for the workaround introduced in #2451 .

Can you please elaborate on
> It will also break resetting belongsToMany relationships, they need to be set to null further down in the loop, they can't continue.
?

How can moving a condition 2 statements up break something further down the loop?

jpolive:
@hojgr your right, but on a **belongsToMany** relationship with no content (ie. on a multi select field you remove all chosen options so the field becomes empty) we need the 
`$multi_select[] = ['model' => $options->model, 'content' => $content, 'table' => $options->pivot_table];`
at the end of the current foreach loop to set the `content` to `null` if we `continue` we will never hit that line which will happen with the change to the **if** statement to always `continue`.

So we need to add a condition to not skip **belongsToMany** rows:
```
if (isset($options->type) && $options->type !== 'belongsToMany') {
    continue;
}
```

But adding the condition above will break the user role thing so you need to add the other condition as well:
```
if ($row->field === 'user_belongstomany_role_relationship' || $row->field === 'role_id') {
    continue;
}
```
I flipped the condition from `$row->field !== 'user_belongsto_role_relationship` because that also breaks absent **belongsToMany** relationships.

The moving a condition 2 statements up is correct and I believe we should keep that change. 
What is breaking things is the change to the `continue`.

hojgr:
OK, I'm starting to understand what the issue is. I'll look into it soon and attempt to come up with a solution. I don't particularly like the fact that we are hardcoding user-role relationship to solve an issue. I think a solution that would be more broad would be a lot better - there is no reason a user-role relationship should be non-standard.

MrCrayon:
@hojgr  @emptynick  This is always true
```
$row->field !== 'user_belongsto_role_relationship'
```

That's why it's preventing role_id and user_belongstomany_role_relationship to be cleared while editing profile with the side effect of preventing any other belongstomany relationship to be reset.

Untill another fix is merged, I think we just need this:
```diff
-                if ((isset($options->type) && $options->type !== 'belongsToMany') || $row->field !== 'user_belongsto_role_relationship') {
+                if ((isset($options->type) && $options->type !== 'belongsToMany') || $row->field === 'role_id' || $row->field === 'user_belongstomany_role_relationship') {
```
As @jpolive pointed out

emptynick:
Closing because you deleted your branch.
https://github.com/the-control-group/voyager/pull/3016 will completely solve this problem.

PR #3070, opened 2018-04-26
Fix saving multiple Images. Fixes #3065

### LINKED ISSUE #3065: Invalid argument supplied for foreach() when edit and don't reupload images
- Laravel Version: 5.5.40
- Voyager Version: 1.1
- PHP Version:7.1
- Database Driver & Version:

### Description:
When i edited an item and didn't reupload images field, it was an error : Invalid argument supplied for foreach()

bajramemini:
There is an error if you save multiple Images without adding new ones. This should Fix #3065


f0reclone:
For anyone searching, it's about line 19 where he checks if there are images.

cotiga (approved):
It work, thanks

fletch3555:
@bajramemini please try to avoid committing so many whitespace changes.  They add nothing and only make StyleCI revert them back after they get merged.  You should add only the lines you've explicitly changed to fix the issue, instead of the whole file.  If you're not familiar with how to do this through the Git CLI, most Git GUI tools will give you a method to do this.

bajramemini:
Hello @fletch3555 I reverted the whitespace changes and only changed what is needed. Regards.

emptynick (approved):

fletch3555 (approved):

Issue #3071, opened 2018-04-26
SQLSTATE[HY000]: General error: 1832 Cannot change column 'role_id': used in a foreign key constraint 'users_role_id_foreign' (SQL: ALTER TABLE users CHANGE role_id role_id INT DEFAULT NULL


asakaze:
- Laravel Version: 5.6.17
- Voyager Version: 1.1.0
- PHP Version:
- Database Driver & Version:

### Description:
Rollback for migration 2017_11_26_013050_add_user_role_relationship.php doesn't work. Out of the box I get error:

Illuminate\Database\QueryException  : SQLSTATE[HY000]: General error: 1832 Cannot change column 'role_id': used in a foreign key constraint 'users_role_id_foreign' (SQL: ALTER TABLE users CHANGE role_id role_id INT DEFAULT NULL)

  at /vendor/laravel/framework/src/Illuminate/Database/Connection.php:664
    660|         // If an exception occurs when attempting to run a query, we'll format the error
    661|         // message to include the bindings with SQL, which will make this exception a
    662|         // lot more helpful to the developer instead of just the database's errors.
    663|         catch (Exception $e) {
  > 664|             throw new QueryException(
    665|                 $query, $this->prepareBindings($bindings), $e
    666|             );
    667|         }
    668|

### Steps To Reproduce:
Use php artisan migrate:rollback.

agoalofalife:
This is a bug

agoalofalife:
You must remove the following line
https://github.com/the-control-group/voyager/blob/1.1/migrations/2017_11_26_013050_add_user_role_relationship.php#L31

agoalofalife:
@asakaze push PR https://github.com/the-control-group/voyager/pull/3073

fletch3555:
#3073 merged, closing

andreebit:
I don't think this is the best way to solve this issue, but I found a workaround. The only thing you have to do is to copy this migration to your project migrations folder.

Hope it will help you.

`
class AddUserRoleRelationshipFix extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('users', function (Blueprint $table) {
            $table->dropForeign(['role_id']);
        });

        Schema::table('users', function (Blueprint $table) {
[...]

PR #3073, opened 2018-04-27
fix issue

agoalofalife:
fix issue https://github.com/the-control-group/voyager/issues/3071


cyrildewit (commented):
Simply removing this line is not the right solution. Because we're changing the `role_id` collumn in the `up` method to an unsigned integer, we should also rollback this change in the `down` method. Even when the collumn is dropped in the `2016_01_01_000000_add_voyager_user_fields` migration.

cyrildewit:
A better, but less nice looking solution would be:

```php
});

    $table->integer('role_id')->change();
});
```

I think that @strickdj already posted this solution at #2863.

agoalofalife:
@cyrildewit offer the full version here?
```php

     */
    public function down()
    {
      });

       $table->integer('role_id')->change();
     });
    }
```

agoalofalife:
@cyrildewit does it work? 
you tried to roll back?
if Yes I am ready to improve your PR

cyrildewit:
I tested my suggestion and it worked. Yes, just put that code between `public function down(){` and `}`.

```php
public function down()
{
    });

        $table->integer('role_id')->change();
    });
}
```

agoalofalife:
@cyrildewit no problem)
done

cyrildewit (approved):

fletch3555:
We try to avoid changing migrations as existing users may not get the updated version.  Given this is only the `down` side of a migration that doesn't get published, I'm inclined to let it through.  Please just be cognizant of that for future changes to migrations.

fletch3555 (approved):

shaunthegeek:
fixed, but has not release?

suggest: should not close issue, until release. create a label "to be release" for it.

emptynick:
Thats how Github works.
If a PR fixes an issue, it will auto-close that issue if the PR is merged into the default branch.

PR #3077, opened 2018-04-27
Patch user roles creation

xaviqv:
- Laravel Version: 5.4
- Voyager Version: 1.1.0

In my migration list, users.id was not defined as unsigned integer, so the foreign key in user_roles table "user_id" could not be created due to this migration error:
image

### DIFF
--- a/migrations/2017_11_26_015000_create_user_roles_table.php
+++ b/migrations/2017_11_26_015000_create_user_roles_table.php
@@ -13,12 +13,16 @@ class CreateUserRolesTable extends Migration
      */
     public function up()
     {
-        Schema::create('user_roles', function (Blueprint $table) {
+		Schema::table('users', function (Blueprint $table) {
+			$table->unsignedInteger('id')->change();
+		});	
+		
+		Schema::create('user_roles', function (Blueprint $table) {
             $table->integer('user_id')->unsigned()->index();
-            $table->foreign('user_id')->references('id')->on('users')->onDelete('cascade');
             $table->integer('role_id')->unsigned()->index();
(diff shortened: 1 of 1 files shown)


migrations/2017_11_26_015000_create_user_roles_table.php
```diff
@@ -13,12 +13,16 @@ class CreateUserRolesTable extends Migration
      */
     public function up()
     {
+        Schema::table('users', function (Blueprint $table) {
```
fletch3555, line 16: We shouldn't be modifying the `users` table in a `create_user_roles_table` migration.  That will just confuse people

Also, we can't edit pre-existing migrations as upgrading users won't be able to run it.

fletch3555 (changes requested):
This migration has existed for months and you're the first one claiming this doesn't work.  I agree that the types should match, but I'm thinking there's another issue at play.  It's possible your database is configured too strictly.  In either case, I can't approve this as-is.

fletch3555:
What part of that error message indicates it's a field type conflict? All I see is an error saying you're trying to create a table with no fields

xaviqv:
@fletch3555 Maybe you're right with my database. I thought that It can be useful to anybody else that's getting the same error.

I got a foreign key error when I moved the foreign key creations to an additional Schema::table method in same migration that took me to this solution.

I comprehend what you say about modifying older migrations, but for anyone else that could get an error on this migration, modifying it is the only solution, rather than relying on a later one that won't execute.

Anyway, It will be okay for me to close and discard this change if in general usage it is wrong.

PR #3082, opened 2018-04-28
Fix namespace naming conflict

fletch3555:
Cannot `use` Voyager Facade class since the service class behind the facade is in the same namespace as this class.  Same as #3041

This PR is the result of comments in #3041 and should supersede it.


cyrildewit:
Can someone merge this? Thanks!

fletch3555:
I can't merge my own PRs, which means @marktopper will have to do so.  I've assigned it to him.

marktopper (approved):

Issue #3045, opened 2018-04-23
Improving the test suite


cyrildewit:
### Summary

I would like to discuss the current test environment. The main goal of this issue is:

- Split tests into `feature` and `unit tests`
- Add more unit tests to cover unexpected bugs
- Add missing tests (#990)
- Improve the test suite

The plan is as follows. Currently, I have not done anything with the existing tets, because I want to decide together with the community which classes we need to unit test and which ones we need to feature test. The current tests remain the same, but we will have to relocate them. I will keep the beneath sections up to date. You can share your opinion with comments and with GitHub comment actions:  👎  👍.

Please be aware that I'm not a professional tester!

Some great posts:
- https://grobmeier.solutions/php-beginners-guide-to-unit-testing.html
- https://matthiasnoback.nl/2014/07/a-better-testing-experience/
- https://matthiasnoback.nl/2014/07/test-doubles/
- https://leftontheweb.com/blog/2017/10/11/your_unit_test_methods_need_docblocks_too/

### Status by class

I am going to include every file in this list, even when they are not testable. They will be skipped after we've discussed them.

- [x] `TCG\Voyager\Actions\AbstractAction` (merged: #3061)
- [x] `TCG\Voyager\Actions\ActionInterface` (not testable)
- [ ] `TCG\Voyager\Actions\DeleteAction`
- [ ] `TCG\Voyager\Actions\EditAction`
- [ ] `TCG\Voyager\Actions\ViewAction`

_The full list can be found in the information bag at the bottom of this post._

### General to-dos

- [x] Clean up the tests (mostly duplicated code) (pending: #3099)
- ~Move `./tests/migrations` to ``./tests/database/migrations` to be more consistent (pending: #3099)~
- ~Replace full out written class paths with `ClassName::class` (#3030)~
---

<details><summary>Information bag</summary><p>

_This hidden place will be used to hide less important information._

### Full list of classes

- [ ] `TCG\Voyager\Alert\Components\AbstractComponent`
- [ ] `TCG\Voyager\Alert\Components\ButtonComponent`
- [ ] `TCG\Voyager\Alert\Components\ComponentInterface`
- [ ] `TCG\Voyager\Alert\Components\TextComponent`
- [ ] `TCG\Voyager\Alert\Components\TiteComponent`
- [ ] `TCG\Voyager\Alert\Commands\AdminCommand`
- [ ] `TCG\Voyager\Alert\Commands\ControllersCommand`
[...]

PR #3099, opened 2018-04-30
Various enhancements

cyrildewit:
This pull request is part of #3045 and contains various enhancements.

* Replaced `DB::Raw` with `DB::raw`
* Removed unnecessary code in tests (like: calling `$this->install();`)
* Removed the `users` and `password_resets` migrations, because we can use `$this->loadLaravelMigrations();`
* Removed the test `User ` model, since it wasn't used
* Extracted the `DisabledTestException` class to his own file
* And some small things


cyrildewit:
I'm aware of the errors. Strange.. It runned locally.

Edit: reverted the breaking change

tests/TestCase.php
```diff
@@ -16,7 +16,11 @@ public function setUp()
     {
         parent::setUp();
 
-        $this->loadLaravelMigrations();
+        if (app()->version() < 5.4) {
```
fletch3555: We no longer support <5.4, so you can actually remove the conditional code entirely.  How was this "breaking"?

tests/TestCase.php
```diff
@@ -75,6 +79,18 @@ protected function getEnvironmentSetUp($app)
 
     protected function install()
     {
+        if (app()->version() >= 5.4) {
```
fletch3555: This `if` isn't necessary since it will always be true

tests/TestCase.php
cyrildewit: @fletch3555 The `loadLaravelMigrations` method didn't exists.

cyrildewit: Maybe something with `orchestra/testbench-browser-kit`

fletch3555: This would be my guess: https://github.com/orchestral/testbench-core/blob/b6ecfd57c72c0dff3288811ed4b8c68ffbd75afc/src/Concerns/WithLaravelMigrations.php#L16

I'm thinking `orchestra/testbench-browser-kit` may have changed something, so it no longer includes that trait (or one of the traits it includes changed to no longer include the above trait).

cyrildewit: I found out that the `WithLaravelMigrations` trait was added to `v3.4.1` (https://github.com/orchestral/testbench-browser-kit/blob/v3.4.1/src/TestCase.php).

`v3.4.0`: https://github.com/orchestral/testbench-browser-kit/blob/v3.4.0/src/TestCase.php

cyrildewit:
@fletch3555 I fixed it, but I want to try some other things out.

**Edit:** done

cyrildewit:
Actually why are we using the lowest versions of the testbench packages? What if they contain bug fixes?

cyrildewit:
@fletch3555 TravisCI is finally passing 🎉 . Are the new version constraints all right?

composer.json
```diff
@@ -27,11 +27,11 @@
         "phpunit/phpcov": "~3.0|~4.0",
         "phpunit/phpunit": "~5.7.14|~6.1|~7.0",
         "laravel/framework": "~5.4.36|~5.5.0|~5.6.0",
-        "orchestra/testbench": "~3.4.0|~3.5.0|~3.6.0",
+        "orchestra/testbench": "~3.4.11|~3.5.0|~3.6.0",
```
fletch3555, line 30: `~` is effectively a wildcard selector, so `~3.4.0` already includes `~3.4.11`.  This change is unnecessary (same for the 2 testbench entries below this.

fletch3555:
@cyrildewit, can you please update this PR by pulling `1.1` into it and fixing merge conflicts?

composer.json
cyrildewit: If you call `composer install --prefer-dist` it will install the lowest version of `orchestra/testbench` = `3.4.0`, but this version isn't compatible. To fix this, we need to set the minimum to `3.4.11`. Otherwise TravisCI will fail.

fletch3555 (approved):

PR #3124, opened 2018-05-03
Mimic behaviour from ContentTypes\Image into ContentTypes\MultipleImage

kylemilloy:
There's an issue where if `upload_max_filesize` in your ini is too low the Intervention Image facade throws the following error inside of `Http\Controllers\ContentTypes\MultipleImage:

screen shot 2018-05-02 at 4 05 48 pm

This occurs because instead of getting an `image/*` mimetype Laravel recieves an `application/octet-stream` as the file only partially uploaded and tries to pass this corrupted file to the the Intervention Image facades make method. 

In `Http\Controllers\ContentTypes\Image` we use a hasFile gate in the handle method. If this gate receives the aforementioned corrupted file then it returns false and fails silently. In this PR we replicate this behaviour for `Http\Controllers\ContentTypes\MultipleImage` by checking to first see if the file being uploaded is a valid image and fails silently. 

Failing silently may not be 100% ideal and in the future an error may need to be sent back and logs written to provide a more detailed error.


kylemilloy:
Whoops... should have double checked my branch...ignore.

Issue #2887, opened 2018-03-24
Upload Multiple Images error


piterog:
- Laravel Version: 5.4
- Voyager Version: v1.0.13
- PHP Version: 7
- Database Driver & Version:

### Description:
When I'm trying to upload multiple images using voyager I have this message:

> Unsupported image type. GD driver is only able to decode JPG, PNG, GIF or WebP files.

How you can see, some images erase the datatype... and all images are .jpg
![apagar](https://user-images.githubusercontent.com/6227151/37859986-7daea9ca-2efb-11e8-921d-11b253fd9bd1.JPG)

### Steps To Reproduce:
1. Upload images

fletch3555:
Simply having a `.jpg` extension does not make it a `JPEG` file.  Please confirm the files are actually jpeg files

piterog:
Yes sir.. all files are jpeg file.

piterog:
Hey guys, i have some news.. there are specific images that can not be sent, at first they are normal .jpg, but they do not send anyway, does anyone have an idea of the problem?

kylemilloy:
I had this issue and it turned out the upload_max_filesize in php.ini was too small which fixed the issue after I shrunk the file. Did this solve it for you?

piterog:
@kylemilloy thanks for the answer man, but still the same error :( Don't work to me

kylemilloy:
Hmmmm...I'd check your **upload_max_filesize** and **post_max_size** just to be sure that any of the two aren't limiting you. 

If you change the limit don't forget to restart php so the changes take effect.

piterog:
Yeh i have do this... post_max_size = 0;
and upload_max_filesize i have commented...

The problem still the same upthere!

kylemilloy:
By default upload_max_filesize is 2M...if you've commented it out this will be what the value is. Set it to something a little more reasonable like 32M, 128M, etc. I notice you've got a file that you're uploading called "picjumbo"...how "jumbo" are we talking? What's the file size?

Restart php again and run phpinfo (either via command line or with a .php file) to confirm that the new value is being used...something like `php -i | grep upload` If the upload_max_filesize isn't being set then it's being overwritten somewhere else...we'll cross that bridge if we come to it...

emptynick:
I'm pretty sure that this is not the problem. Otherwise PHP/Laravel would die with a `PostTooLargeException`

kylemilloy:
[...]

PR #3125, opened 2018-05-03
Fix image bugs and improve menu performance

kylemilloy:
There's an issue where if `upload_max_filesize` in your ini is too low the Intervention Image facade throws the following error inside of `Http\Controllers\ContentTypes\MultipleImage`:

screen shot 2018-05-02 at 4 05 48 pm

This occurs because instead of getting an `image/*` mimetype Laravel recieves an `application/octet-stream` as the file only partially uploaded and tries to pass this corrupted file to the the Intervention Image facade's make method.

In `Http\Controllers\ContentTypes\Image` we use a hasFile gate in the handle method. If this gate receives the aforementioned corrupted file then it returns false and fails silently. In this PR we replicate this behaviour for `Http\Controllers\ContentTypes\MultipleImage` by checking to first see if the file being uploaded is valid and fail silently as well if it is not.

Failing silently may not be 100% ideal and in the future an error may need to be sent back and logs written to provide a more detailed error...however this PR just focuses on copying existing behaviour.

Should resolve this issue: https://github.com/the-control-group/voyager/issues/2887


kylemilloy:
Added caching to the menu system. Editing, adding, deleting and reordering clears the menu cache. If you're using redis/memcached or anything that supports cache tagging it'll clear out tagged caches...otherwise it is more aggressive and flushes the cache.

### Before cache.
screen shot 2018-06-29 at 12 01 34 am

### After...
screen shot 2018-06-29 at 12 01 42 am

emptynick:
I fixed the merge-conflicts.
Tested and looks good to me. Thanks!

emptynick (approved):

PR #3145, opened 2018-05-06
Update voyager_tinymce.js with setup callback

CSchloegel:
This callback allows developers to access the TinyMCE Editor BEFORE it has been instantiated. Useful for adding plugins etc...


emptynick (approved):

Xmk:
Where it?
https://github.com/the-control-group/voyager/blob/1.1/resources/assets/js/voyager_tinymce.js

emptynick:
In 1.x because its a breaking change.
https://github.com/the-control-group/voyager/blob/4e5e8c80507b2d5152994ad07da80a626e106958/resources/assets/js/voyager_tinymce.js#L31-L35

Xmk:
I need to autoload the data\blob images on the server, but the Voyager simply does not allow me to do this with human methods!

emptynick:
Switch to 1.x if you need this callback.

Xmk:
@emptynick thanks, I try

niels-numbers:
> Useful for adding plugins etc...

How do you add a plugin with the  setup callback? I couldn't find that in the TinyMCE docs.  Maybe this would be useful to add to the docs aswell.

PR #3163, opened 2018-05-10
PR for issue #3097

### LINKED ISSUE #3097: Editing a table with a row type of file, saves the value as "[]"
- Laravel Version: 5.6
- Voyager Version: 1.1.0
- PHP Version: 7.1
- Database Driver & Version: MySQL 5.7.19

### Description:
When editing a table with a row->type of "file". If you already have a file uploaded and have changed other content but have left the file as-is, the file field is updated to a string of "[]".

### Steps To Reproduce:

1. Using the BREAD builder, create a new entry in a table which has a row type of File. Populate all fields, then save.
2. Edit the new entry, change some fields, but leave the file field as-is. It should still reference the original file you uploaded, then save.
3. The file field should now have been saved as "[]", meaning you have lost the refence to the original file and its path etc.

I have found a fix for this, but I'm not sure if this needs looking into further down the line.

In vendor/tgc/voyager/src/Http/Controllers/Controller.php - insertUpdateData(), Line 88.

Change:

`if (is_null($content)) {`

to 

`if (is_null($content) || $content == '[]') {`

This check catches the empty json array and passes in the orignal value as intended.

```
// If the file upload is null and it has a current file keep the current file
if ($row->type == 'file') {
  $content = $data->{$row->field};
}
```

reficul:
When editing a table with a row->type of "file". If you already have a file uploaded and have changed other content but have left the file as-is, the file field is updated to a string of "[]".

Fixes #3097


emptynick (approved):

fletch3555 (approved):

fletch3555:
Closing in favor of #3521

PR #3164, opened 2018-05-10
Added preserveFileUploadName also for FILE

reficul:
This PR continue to use random as default but I think that is better for 1.1 to set original filenames as default and random as option. There is the whole question of “SEO” and the possibility for users to have more recognizable files that I think should lead to reverse things.


reficul (commented):
TravisFix

fletch3555 (approved):

PR #3170, opened 2018-05-11
Allows null options on relationship

tjmckenzie:
If field row is not required, provide None/Null option for BelongsTo and BelongsToMany


resources/views/formfields/relationship.blade.php
```diff
@@ -29,8 +29,13 @@
 				<select class="form-control select2" name="{{ $options->column }}">
 					@php 
 						$model = app($options->model);
-	            		$query = $model::all();
-	            	@endphp
+						$query = $model::all();
+					@endphp
+
+					@if($row->required === 0)
+						<option value="">None</option>
```
cyrildewit: Would it be an idea to replace the static `None` with an automatic translation?

cyrildewit (changes requested):

resources/views/formfields/relationship.blade.php
emptynick: Theres already a translation. `generic.none` iirc

tjmckenzie: Good idea. I have just added it. I was glad to see "none" was already described in the generic translation files.

cyrildewit (approved):
Looks good to me!

tjmckenzie:
Any ETA on a merge / release?

fletch3555 (approved):

PR #3176, opened 2018-05-12
fix save multi image in admin panel

veggigit:


emptynick:
Sorry, but this is a duplicate of https://github.com/the-control-group/voyager/pull/3060

PR #3177, opened 2018-05-14
Profile feature

### LINKED ISSUE #2625: Permissions error when user can add/edit but not browse
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7.1

### Description:
When saving your profile, users who do not have access to Browse Users are returned to the `users/index` view which throws a permissions error. 

### Steps To Reproduce:

- Create a new user with a role of `user`
- Make sure the user role does not have any BREAD permissions for the `users` database
- Login as the new user
- Go to > Profile > Edit My Profile
- Click Save

Error is returned: `AccessDeniedHttpException This action is unauthorized.` which is correct as this user does not have access to Browse users.

### Solution:

Update the `VoyagerBreadController` function `update()` from 

`->route("voyager.{$dataType->slug}.index")`

to 

`->route(Voyager::can('browse_' . $dataType->slug) ? "voyager.{$dataType->slug}.index" : "voyager.dashboard")`

This redirects to the dashboard instead of to the index slug, which should catch the permissions error. 

Is there a better / more elegant way of doing this?

Braiani:
This PR introduce the `VoyagerProfileController` and implement the suggestion on #2625 (fixes #2625).
Now an user that can't BREAD the users table can edit his own profile.
+ Added a verification that prevents the user gain unauthorized permissions. With that, even if the user inject an input called `role_id` it won't take effect, as the request is validated and the main role is edited.
+ The same occurs with the additional roles that are taken out of the request.


cyrildewit:
This looks like a clean solution! Did someone else already confirm the fixes? I can't test this fix right now, because my Homestead is erroring. I might be able to get it working tonight.

Tagging @MrCrayon

fletch3555:
If someone else confirms this is working well, I'll gladly get it merged in.  I'm also unable to test it right now.

MrCrayon:
I'm out and I can't test it but I think this PR should also address the preserving of user_belongstomany_role_relationship values. 
Right now it should work only because of the temporary fix that doesn't allow user_belongstomany_role_relationship to be reset.

cyrildewit:
> Right now it should work only because of the temporary fix that doesn't allow user_belongstomany_role_relationship to be reset.

@MrCrayon do you mean this temporary fix by @emptynick? https://github.com/the-control-group/voyager/pull/3014

cyrildewit:
What about replacing the `/admin/users/1/edit` url with `/admin/profile/edit`? It looks cleaner in my opinion. We can reuse `edit-add` view.

MrCrayon:
@cyrildewit yes, that one

src/Http/Controllers/VoyagerProfileController.php
```diff
@@ -0,0 +1,61 @@
...
+
+        //Verify if user tries to modify his own role
+        if ($request->has('role_id')) {
+            $request->merge(['role_id' => $request->user()->role->id]);
+        }
+
+        if($request->has('user_belongstomany_role_relationship')){
+            $request->offsetUnset('user_belongstomany_role_relationship');
+        }
+
```
cyrildewit: I think you missed the `user_belongsto_role_relationship` field.

```php
if ($request->has('user_belongsto_role_relationship')) {
    $request->offsetUnset('user_belongsto_role_relationship');
}
```

I was able to change my main role to something different by adding `<input type="text" class="form-control" id="user_belongsto_role_relationship" name="user_belongsto_role_relationship" value="1">` to the page.

MrCrayon: @cyrildewit I think you discovered a new bug.
AFAIK the belogsto (belongstomany is different) is never sent in an edit request because the name is set as:
```php
<select class="form-control select2" name="{{ $options->column }}">
```
in relationship.blade.php, where $option->column is the column referred in the same table (es. role_id)
Therefore the belogsto should never be taken in consideration in the request.

cyrildewit:
I thought I made a mistake, so I removed the review. When I can confirm this bug with full certainty, I will re-add the review.

cyrildewit:
## user_belongsto_role_relationship

Okey, I rechecked the bug and it's really a bug.

### Steps to Reproduce

1. Go to profile
2. Edit profile
3. Add following field to the webpage `<input type="text" class="form-control" name="user_belongsto_role_relationship" value="!! RANDOM_NUMBER !!">`

src/Http/Controllers/VoyagerProfileController.php
cyrildewit: I already reviewed this, but I kind of panicked when I thought I made a mistake and removed it without thinking....

I think you also need to remove the `user_belongsto_role_relationship` field from the request.

```php
if ($request->has('user_belongsto_role_relationship')) {
    $request->offsetUnset('user_belongsto_role_relationship');
}
```

I was able to change my default role by inserting the following input to the webpage: `<input type="text" class="form-control" name="user_belongsto_role_relationship" value="!! RANDOM_NUMBER !!">`

src/Http/Controllers/VoyagerProfileController.php
```diff
@@ -0,0 +1,61 @@
...
+        if ($val->fails()) {
+            return response()->json(['errors' => $val->messages()]);
+        }
+
+        if (!$request->ajax()) {
+            $this->insertUpdateData($request, $slug, $dataType->editRows, $data);
+
+            event(new BreadDataUpdated($dataType, $data));
+            
+            $route = $request->user()->can('browse', $data) ? 'voyager.users.index' : 'voyager.profile';
```
cyrildewit: Can we not just always redirect the user back to his profile?

cyrildewit (changes requested):

src/Http/Controllers/VoyagerProfileController.php
Braiani: Thanks for the reply.. I'll check here and submit another commit solving this problem!

src/Http/Controllers/VoyagerProfileController.php
Braiani: I used it because if the user comes to here from the users browse page and edit his own user, it'll be redirect to profile page... In fact, I'm thinking on improving this based on the incoming page the user access the user edit page... Any suggestion?

Braiani:
> Right now it should work only because of the temporary fix that doesn't allow user_belongstomany_role_relationship to be reset.

> @MrCrayon do you mean this temporary fix by @emptynick? #3014

I'll work on it.

@fletch3555 could you wait until I solve this problems without close this PR?

src/Http/Controllers/VoyagerProfileController.php
MrCrayon: I think it can be done, if we exclude saving it in session, adding a returnUrl hidden input in edit page, so you know if you are editing coming from profile or user list.

I liked the more generic idea to have user sent back as @carlost89 suggested in #2625 :
```php
return redirect()->back()
```
when they don't have browse permission, but that would need to be applied in VoyagerBase.

MrCrayon:
I'm not a big fan of duplicating code that's why I approched the problem that way in #3016.
But I like the idea of Profiles being someway managed differently than users.

I was thinking if we create a BREAD Profiles that uses the same Model (TCG\Voyager\Models\User) and same Policy, we could set in BREAD that role_id is not editable.
This way without adding any other checks it should work.
Plus it would be possibile to have a user that can manage other users editing their profile but it wouldn't be able to change their roles.

Thoughts?

P.S.
I tried doing this and it looks like it's working but I didn't do extensive tests.

src/Http/Controllers/VoyagerProfileController.php
cyrildewit: @MrCrayon suggestion looks good:

```php
return redirect()->back();
```

fletch3555: You should never reach the point where you don't know where to redirect back to with this.  You can view your profile, edit your profile, and save your edited profile.  We just need to choose whether you go back to the "edit" form page, or the "view" page.  My preference for the view page.  Using `redirect()->back()` solves the problem to an extent as it will always go back to the same place since there's only one form that should ever submit to this, right?

I agree we don't need the conditional here, but I don't quite agree with `redirect()->back()` either.  Convince me?

cyrildewit: @fletch3555 we could decouple the relation between `/admin/profile` and `/admin/users/1`?

Yes we have to create two controllers, but I don't see any problems.

### My Profile feature

View URL: `/admin/profile`
Edit URL: `/admin/profile/edit` (redirects to: `/admin/profile`)

### User BREAD feature

Browse URL: `/admin/users`
View URL: `/admin/users/{id}`
Edit URL: `/admin/users/{id}/edit` (redirects to: `/admin/users/{id}`)

fletch3555: Sorry for the delay, but yes @cyrildewit, that's exactly what this PR does.

fletch3555:
@MrCrayon, I'm not a fan of encouraging multiple BREADs per table in this version of Voyager, since it's not a feature supported by the UI.  @emptynick was working a version for a future release that is a complete rewrite of the UI and does support multiple BREADs, so we can definitely revisit that idea when that's more mature.

For now, I'd like to continue with the ProfileController in this PR (`/admin/profile` != `/admin/user/{x}`).

@Braiani, please let me know when you're ready and we'll review it then.  Thanks!

src/Http/Controllers/VoyagerProfileController.php
cyrildewit: @fletch3555 not exactly. If you edit your profile at the moment, you will be pointed to `/admin/users/{id}/edit` instead of `/admin/profile/edit`.

fletch3555: That's not what this PR should be doing.  Perhaps a link was missed.

If you edit a user (that happens to be your own), you should go to `/admin/users/{id}/edit` like normal.
If you edit your profile, you should visit `/admin/profile/edit` (with the changes in this PR at least).

Submitting the edit form on either of those pages should return you to their respective "view" page (`/admin/users/{id}` or `/admin/profile`)

cyrildewit: Do you mean that this PR should not add the extra route `/admin/profile/edit`?

Braiani: I was talking with @emptynick yesterday about this feat. I'll do some improvements with the routes to use `admin/profile/edit` when accessing the profile edit page and `admin/user/{id}/edit` when is editing an user.
I'll do it on this week!

@fletch3555 is there anything you want me to improve on the profile feature?

Braiani:
@fletch3555, Could you check if my improvement is good enough? If you think there's more to do, just say what is needed!
@emptynick could you check it too?

tanys123:
Hi guys, how is the review progress? Is there anything that i can help on?

Braiani:
Thanks @emptynick for the fixes!

MrCrayon:
This PR introduces profiles, so a user can safely modify it's own profile without being able to modify associated roles.
Said that I think there are some points that needs to be addressed:

1. I'm still able to go to admin/users/MYID/edit and exploit bug there.
I think user policy need to be changed and a policy for profile need to be created.
2. Value of user_belongstomany_role_relationship still need to be preserved.
Right now the value is not cleared only because temporary fix #3014 doesn't let any belongsToMany relationship to be cleared.
3. Revert 3014

fletch3555:
@MrCrayon, @emptynick and I were talking earlier, and believe a couple simple changes would meet your needs. We keep the profile page here, and add a redirect in the UsersController as follows:

If $requestedUserId == $currentUserId, then redirect to profile page

This way, a user will require the `edit_users` permission to access `admin/users/{id}/edit`, but can access `/admin/profile/edit` without it.  This would be a breaking change due to the needed creation of a UsersController, and would therefore wait until v1.2.

In parallel, we can add some validation around what fields are submitted and saved, which should meet your immediate concerns, but that's different than this PR(even though this stemmed from that discussion).

Thoughts?

MrCrayon:
@fletch3555 I'm not sure what you mean with redirect to profile page.

Can you have a look at #3016?
AFAIK that works, so maybe somethng can be taken from there.

Meanwhile I'm in slack group.

emptynick:
Also fixed in #3016.
Merged into 1.x

PR #3182, opened 2018-05-15
Add alias to Intervention Image class in MultipleImage content type

### LINKED ISSUE #3179: Cannot use Intervention\Image\Facades\Image as Image because the name is already in use
- Laravel Version: 5.6
- Voyager Version: 1.1
- PHP Version: 7.0.9
- Database Driver & Version:

### Description:
when i select multiple image in voyager  and then add the voyager show this error: 
"Cannot use Intervention\Image\Facades\Image as Image because the name is already in use".
please let me know what is the problem behind it and how i can fix this..

cyrildewit:
When using the `Intervention\Image\Facades\Image` facade directly inside the `MultipleImage` content type, we are getting the following error: `Cannot use Intervention\Image\Facades\Image as Image because the name is already in use`. This error appears because there is already a class named `Image` in the same namespace (the content type: Image).

It was reported by @PinkuChanda at #3179 and solved by @emptynick.

I think merging this pull request without squashing will add @emptynick as Co Author in the commit.

---

Co-Authored-By: Christoph Schweppe
Fixes #3179

### DIFF
--- a/src/Http/Controllers/ContentTypes/MultipleImage.php
+++ b/src/Http/Controllers/ContentTypes/MultipleImage.php
@@ -5,7 +5,7 @@ namespace TCG\Voyager\Http\Controllers\ContentTypes;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
 use Intervention\Image\Constraint;
-use Intervention\Image\Facades\Image;
+use Intervention\Image\Facades\Image as InterventionImage;
 
 class MultipleImage extends BaseType
 {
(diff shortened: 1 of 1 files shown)


PinkuChanda (approved):
Its Working Properly. Thanks for the quick fixes. 
#Happy_Coding 

Regards 
Pinku Chanda

emptynick:
Can this fail if you have an `Image` field + a `Multiple Images` field in one BREAD?
Because we then have two times
`use Intervention\Image\Facades\Image as InterventionImage;`

PinkuChanda:
Yes image and multiple image is working properly at same bread.. i think it is fine now.

fletch3555:
No it won't fail with both in the same BREAD.

The issue is a namespace conflict, and each PHP class has it's own locally-scoped namespace.  Adding `use` statements effectively import the other classes into that local namespace.  The `Image` import was conflicting because there's a PHP class called `Image` in the same directory (automatically imported into the same namespace).  This PR should be the end of these issues.

fletch3555 (approved):

PR #3192, opened 2018-05-16
change voyager.user.namespace config behavior

eidng8:
change `voyager.user.namespace` config behavior to override, defaults to `auth.providers.users.model`.

Why do I do that? Try the following:
```
laravel new AwesomeApp
cd AwesomeApp
composer require tcg/voyager
php artisan app:name AwesomeApp
php artisan voyager:install

```

And it yells! Can't find `App\User`!

This PR should break no system, existing or in the future.  I think :smirk:

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -13,7 +13,7 @@ return [
     'user' => [
         'add_default_role_on_register' => true,
         'default_role'                 => 'user',
-        'namespace'                    => App\User::class,
+        'namespace'                    => null,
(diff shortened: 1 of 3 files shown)


cyrildewit (approved):

fletch3555 (approved):

fletch3555:
Sorry for the branch confusion.  I was going to claim this would be a breaking change due to the config change, but realized the code is only a fallback and the config will already exist for all existing users.  I'll get this merged in once TravisCI is happy

cyrildewit (approved):

PR #3193, opened 2018-05-16
Tune string columns

eidng8:
- change non-linguistic columns to use `latin1`
- shrink linguistic columns to a sensible size

### DIFF
--- a/migrations/2016_01_01_000000_create_data_types_table.php
+++ b/migrations/2016_01_01_000000_create_data_types_table.php
@@ -15,12 +15,12 @@ class CreateDataTypesTable extends Migration
         // Create table for storing roles
         Schema::create('data_types', function (Blueprint $table) {
             $table->increments('id');
-            $table->string('name')->unique();
-            $table->string('slug')->unique();
+            $table->string('name', 191)->unique();
+            $table->string('slug', 191)->unique();
             $table->string('display_name_singular');
             $table->string('display_name_plural');
             $table->string('icon')->nullable();
-            $table->string('model_name')->nullable();
+            $table->string('model_name')->charset('latin1')->collate('latin1_general_ci')->nullable();
             $table->string('description')->nullable();
             $table->boolean('generate_permissions')->default(false);
             $table->timestamps();
@@ -30,8 +30,8 @@ class CreateDataTypesTable extends Migration
         Schema::create('data_rows', function (Blueprint $table) {
             $table->increments('id');
             $table->integer('data_type_id')->unsigned();
-            $table->string('field');
-            $table->string('type');
+            $table->string('field')->charset('latin1')->collate('latin1_general_ci');
+            $table->string('type')->charset('latin1')->collate('latin1_general_ci');
             $table->string('display_name');
             $table->boolean('required')->default(false);
             $table->boolean('browse')->default(true);
--- a/migrations/2016_05_19_173453_create_menu_table.php
+++ b/migrations/2016_05_19_173453_create_menu_table.php
@@ -14,7 +14,7 @@ class CreateMenuTable extends Migration
     {
         Schema::create('menus', function (Blueprint $table) {
             $table->increments('id');
-            $table->string('name')->unique();
+            $table->string('name', 100)->unique();
             $table->timestamps();
         });
 
@@ -22,10 +22,10 @@ class CreateMenuTable extends Migration
             $table->increments('id');
             $table->unsignedInteger('menu_id')->nullable();
             $table->string('title');
-            $table->string('url');
-            $table->string('target')->default('_self');
-            $table->string('icon_class')->nullable();
-            $table->string('color')->nullable();
+            $table->string('url')->charset('latin1')->collate('latin1_general_ci');
+            $table->string('target')->charset('latin1')->collate('latin1_general_ci')->default('_self');
+            $table->string('icon_class')->charset('latin1')->collate('latin1_general_ci')->nullable();
+            $table->string('color')->charset('latin1')->collate('latin1_general_ci')->nullable();
             $table->integer('parent_id')->nullable();
             $table->integer('order');
             $table->timestamps();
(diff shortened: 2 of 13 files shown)


yangqi:
Shrinking column size is unnecessary, the default length is 255. Setting default charset and collation would make it incompatible with other database, like sqlite, postgresql and sql server, which are all supported by laravel.

eidng8:
Well, I though Voyager is mysql only, at least for the moment. And I made this change with that presumption in mind.

If dbms compatibility is concerned, this would be a lot more difficult.

As far as I know, pg & mssql don't have this kind of problem. But I wouldn't detect db type in migrations. This leaves a dead lock here.

As of the column size, it's a trade off.

eidng8:
Details on column sizes:

- `name`: please enlighten me with anyone's name longer than 100 characters
- `slug`: I've never seen any slug as long as 200 characters, however, non-latin slugs are commonly used
- `model_name`: I presume it's PHP's class name with namespace. Though it's allowed to be very long, but I haven't seen anyone using long class paths.
- `url`: Once they are encode with `rawurlencode()` , no non-latin characters can't get in it. Also, valid URL can get well over 255 characters. I've just left it to your predetermined 255 size.
- `target`: It's for `<a>` element I presume. And I don't expect someone will use non-latin for it. I can remove the collation if you want to.
- `icon_class`: It's for HTML element's `class` I presume. So no non-latin here.
- `color`: It's color code or commonly used color names I presume. Also no non-latin here.
- `key`: Although it's possible to use non-latin key names, I presume the `display_name` is better for that. So I've changed it to `latin1`. For tables without `display_name`, I presume they are used in program logic, so it'll be save to use `latin1` too
- `table_name`, `column_name`, & `locale`: I don't see any necessity to use non-latin here.

yangqi:
Well, voyager is supposed to support laravel 5.4 to 5.6, which support other rmdb than mysql. Voyager should support those rmdb as well, otherwise should list the limitation.

Besides all the compatibility issue, I see all these changes unnecessary, all those column definitions can be changed within voyager, users will change them if needed.

eidng8:
No they can't. Just try a fresh install and you'll know it. `php artisan voyager:install` fails due to key length limit.

Also, if Voyager is really keen on support all rdbms as Laravel, we've got one failure here: MySQL. Users can't even install it without hacking a bit.

yangqi:
I mean column length. I just tried, there is no issue change column length in voyager.

eidng8:
You mean using the Database Manager page? If so, how am I suppose to use that page if I can't even install voyager?

If that's not what you mean, I'm sorry I can't get it. What do you "change column length in voyager"?

eidng8:
Actually, changing your migration was not the first flash in my mind. I was first went through you docs to find a way to publish migrations to my project. But I can't find a way.

So, my first choice is, publish migration to hosting project, let me do what I have to do, without touching your code base. IMO, this is the Laravel way.

yangqi:
What do you mean you can't install voyager? If there is issue installing voyager, please describe in detail.

eidng8:
Oops, sorry for not mentioning that. This is what I've done:

```
laravel new Voyager
composer required tcg/voyager
php artisan voyager:install
```

And the key length too long error popped up.

Then I followed the link suggested in the document, changing default string length to 191 in the `AppServiceProvider`, and `php artisan voyager:install` again. The key length too long error popped up again. This time pointed at [this line](https://github.com/the-control-group/voyager/blob/3dd1dcf4f8498d8fe5b92bcfdcf04886e010d2c1/migrations/2017_01_14_005015_create_translations_table.php#L26).

I've did the above on a box:
- Windows 7 x64
- MySQL 5.7.21
- PHP 7.2.4

yangqi:
Well, this is an issue in laravel that has been documented.

https://laravel.com/docs/master/migrations#indexes

> Laravel uses the utf8mb4 character set by default, which includes support for storing "emojis" in the database. If you are running a version of MySQL older than the 5.7.7 release or MariaDB older than the 10.2.2 release, you may need to manually configure the default string length generated by migrations in order for MySQL to create indexes for them. You may configure this by calling the Schema::defaultStringLength method within your AppServiceProvider

> Alternatively, you may enable the innodb_large_prefix option for your database. Refer to your database's documentation for instructions on how to properly enable this option.

eidng8:
Yes, they've said that, so do you. I'm not blaming you on this.

To me, there are only two ways to ease the pain in the future.

1. Just like what I did, trading off.
2. Publish migrations too, so users can choose to DIY.

And I prefer option 2.

yangqi:
Laravel already provides a better solution to override default string length in AppServiceProvider, so no need to change every migration.

eidng8:
IMO, that is worse! Changing it globally brings big impact, much bigger than fixing a few columns' size & collection.

Also, the translation table uses tri-column index, which exceeds the limit even after shrinking. As pointed out in the above install steps I've tried.

yangqi:
No, it's better because this issue only affects older versions of mysql or mariadb, and it's up to each user to decide whether they need to have this work around. Also it's much easier in the future to take out the work around if they upgrade their database. Anyway, this discussion is not related, now it's clear that the issue is not in voyager.

fletch3555:
Couple notes here.

1. We don't edit migrations, as that would break upgrades (they wouldn't get the changes since the migration already ran.
2. Voyager supports all the same RDBMS that Laravel supports (though we haven't explicitly tested them all.  (If you know of specific issues, please report them in github issues.)
3. That key length issue you mentioned with mysql is 100% related to Voyager (similarly phrased error message as the common MySQL one also mentioned above).  The issue (and workaround) have been mentioned in github issues here in the past, so you're welcome to search for it (I don't have time right now).  If you find a better (more permanent) solution than the workaround mentioned, please feel free to open a PR with the fix.  We simply haven't gotten around to it yet, and it's been a lower priority since there's a known workaround.
4. I disagree with many of your arguments here.  These are certainly not critical optimizations, as they won't benefit a majority of our user base.  If you need them to make your app function, then you're certainly welcome to add a migration to your app to modify the collation/field length for these fields.  Until a better argument for this is made, I'm going to close this.

PR #3207, opened 2018-05-17
add user model fallback to voyager:admin

eidng8:


cyrildewit (approved):
This looks fine!

fletch3555 (approved):

fletch3555:
This probably should've just been bundled with #3192, but that's okay.

eidng8:
Yes it should. But I haven't found out this issue by then. And that PR is being reviewed. So I've created a new one.

PR #3208, opened 2018-05-17
Delete item error if there are empty file fields

reficul:
I can’t delete a bread item with empty file fields. Voyager try to delete files also if not present. 
Then the foreach fire an error: 
*Invalid argument supplied for foreach()*


emptynick:
Superseded by https://github.com/the-control-group/voyager/pull/3580
Sorry for not looking at your PR earlier.

PR #3214, opened 2018-05-19
Bug Fix: Convert belongsToMany relationship data to array before sync

### LINKED ISSUE #3213: belongsToMany relationship fails with multiple values
- Laravel Version: 5.5.40
- Voyager Version: 1.1.1
- PHP Version: 7.1
- Database Driver & Version: 5.7

### Description:
Creating a many-to-many relationship fails when updating model with multiple values. Error message reads:
```
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'event_id' at row 1 (SQL: insert into 
`applicant_event` (`applicant_id`, `event_id`) values (2, 6,9))
```

The issue is that `sync()` expects an array with multiple values but is being sent a string, in this example `sync("6,9")`.

### Steps To Reproduce:
Create an events model, an applicants model, and a pivot table applicant_event. 
And add a belongsToMany relationship to the applicant model
When you add several events in the admin
And add multiple events as a relationship to an applicant
Then none of the events are saved as a relationship
And the page returns an error

Grabes:
PR that allows multiple values to be saved with a belongsToMany relationship. Resolves #3213

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -125,7 +125,7 @@ abstract class Controller extends BaseController
             } else {
                 $content = $sync_data['content'];
             }
-            $data->belongsToMany($sync_data['model'], $sync_data['table'])->sync($sync_data['content']);
+            $data->belongsToMany($sync_data['model'], $sync_data['table'])->sync($content);
         }
 
         return $data;
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -120,6 +120,11 @@ abstract class Controller extends BaseController
         }
 
         foreach ($multi_select as $sync_data) {
+            if (!is_array($sync_data['content'])) {
+                $content = explode(',', $sync_data['content']);
+            } else {
+                $content = $sync_data['content'];
+            }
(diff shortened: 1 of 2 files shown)


Grabes:
17dda76 was accidentally pushed to this branch. Reverted the commit and it is now ready for merge.

Grabes:
The fix uses a fallback to the original behaviour if array content is passed. Looks like we are missing tests for relationships within Voyager and I will work on adding them as time permits. But could we merge this change into 1.1.x in the meantime?

cyrildewit:
Can you post a screenshot of the dump where the value of `$sync_data['content']` is not an array? I'm trying to reproduce this bug, but it's always an array by me.

sadfsdfsadfadsfadsfsdfsdf

I placed the following code above the foreach loop.

```php
dd($multi_select);
```

Grabes:
Here is a screenshot from my site where this error is occurring:
voyager

You can see the content of the driver_league relationship is a string and not an array

Grabes:
Could we get this merged?

MrCrayon:
I'm closing this one since I can't reproduce the but that is trying to fix.
If you can prepare a test that shows the failure you can reopen it.

PR #3233, opened 2018-05-25
Fix column not found when using additional_attributes w/ xMany relations

SteamWind:
Fix for belongsToMany and hasMany relationship display with
additional_attributes. You can now use additional_attributes with those
relations.
https://github.com/the-control-group/voyager/pull/2885

The fix get the model before plucking it. It caused error because when
you pluck the query builder, it affects the mysql query. When you use
additional attribute, it doesn’t exist in the database and result with
a `SQLSTATE[42S22]: Column not found: 1054 Unknown column
‘youradditionalattribute’ in 'field list’` error…


SteamWind:
There is also a problem with pivot table on display, it prefix the already prefixed table. Which causing this error:
```php
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'mydatabase.myprefix_myprefix_mytable' doesn't exist [...]
```

When we create the relationship, we need to put the prefix out of the name of the pivot table. I did something locally but I don't know if i'm the only one having this problem.
I changed the same line here:
```php
belongsToMany($options->model, str_replace(\DB::getTablePrefix(), '', $options->pivot_table)
```
Should I modify my pull request or create an other one and delete this one?

SteamWind:
I need to do it in a branch, I will submit it back lather.

PR #3243, opened 2018-05-28
Add Translatable do BackOffice

pedro-c-oliveira:
Add Translatable do BackOffice.
Only work on create table. Don't work on edit table. 
Made editions manually on model


SteamWind:
You shouldn't add the translation of the language you can't translate...

fletch3555:
On the contrary @SteamWind, I'd prefer new translation keys be added to all languages with English as a default.  This will keep the files up to date.  It looks better to have English text in a couple places than `new.translation.string`.

As for this PR, I'm not sure I understand what this is supposed to do...

SteamWind:
Okay, put the key is a good thing but what if we want to change the text of the english file? We need to change it everywhere then.
(I speak in general case, in this one it should be okay)

fletch3555:
Only the languages you use. And you'd likely have to do that even if it was translated appropriately.

publishable/lang/fr/database.php
```diff
@@ -62,6 +62,7 @@
     'url_slug'                  => 'URL Slug (doit être unique)',
     'url_slug_ph'               => 'URL slug (exemple : articles)',
     'visibility'                => 'Visibilité',
+    'is_translatable'           => 'Do you wish install translations?',
```
SteamWind: `'is_translatable'           => 'Voulez-vous installer les traductions?',`

publishable/lang/fr/generic.php
```diff
@@ -87,4 +87,5 @@
     'viewing'                => 'Affichage',
     'yes'                    => 'Oui',
     'yes_please'             => 'Oui SVP',
+    'is_translatable'        => 'Translatate?',
```
SteamWind: `'is_translatable'        => 'Traduire?',`

SteamWind (changes requested):

pedro-c-oliveira:
This is what this pull enhancement do in backoffice:

1. When do you wish that voyager creates model for your table, shows a button ask if you wish that model could be translatable.


2. if you wish apply translatable to your model a new column apear on table fields to choose wich fields are translatable.




3. After that you can check your new model and you can see the trait and fields translatable array on php file.




At the moment this feature only works on create new table not in edit mode.

publishable/lang/fr/generic.php
pedro-c-oliveira: i just add the necessary keys for all language files in voyager on english, after that, people will translate the key for the correct sentence in his language.

Martins-74:
This is very useful, we can make multilanguages websites in minutes with this enhancement.
Hope it get's approved  👍 ❤️

src/Commands/MakeModelCommand.php
```diff
@@ -67,6 +67,35 @@ protected function addSoftDelete(&$stub)
...
+        $traitIncl = $trait = $translatablefields = '';
+
+        if ($this->option('traitTranslatable')) {
+            $traitIncl = 'use TCG\Voyager\Traits\Translatable;';
+            $trait = 'use Translatable;';
+            $translatablefields='protected $translatable = [';
+            foreach ($this->option('traitTranslatable') as $index => $field) {
+                $translatablefields.='"'.$field.'",';
+            }
+            $translatablefields.='];';
```
emptynick:
```
$translatablefields = 'protected $translatable = ["'.implode($this->option('traitTranslatable'), '", "').'"]';
```
Saves you 4 lines.

emptynick (changes requested):

src/Commands/MakeModelCommand.php
pedro-c-oliveira: thank you

publishable/lang/fr/generic.php
SteamWind: Yes okay.

SteamWind:
Hope we could have this in edit mode soon :)

pedro-c-oliveira (commented):
improve code efficience

src/Commands/MakeModelCommand.php
pedro-c-oliveira: request changes done.

pedro-c-oliveira:
Someone reviewer to aprove the pull please

src/Commands/MakeModelCommand.php
```diff
@@ -67,6 +67,30 @@ protected function addSoftDelete(&$stub)
...
     }
 
+    /**
+     * Add Translatable to the given stub.
+     *
+     * @param string $stub
+     *
+     * @return $this
+     */
+    protected function addTraitTranslatable(&$stub)
```
fletch3555: Can you rename the method so it's consistent with the others?

fletch3555 (changes requested):

SteamWind:
Does it then use the translation table?

src/Commands/MakeModelCommand.php
pedro-c-oliveira: @fletch3555 do you prefer just "addTranslatable" ?  or suggest another name?

emptynick:
I just fixed the merge-conflicts.
You have some nasty JS scripts to show/hide the checkboxes, but the database editor is a vue-app which would make them unnecessary if used properly.

emptynick:
I'm going to close this PR because it has changes requested and wasn't touched for over half a year.
If you are still interested in fixing the requested changes, please let us know.

PR #3247, opened 2018-05-29
Fixes #3246

### LINKED ISSUE #3246: Fix front-end on the "Viewing Database" page
https://github.com/the-control-group/voyager/blob/2bf3bc072205187323bdb2552e2264862abd48a9/resources/views/tools/database/index.blade.php#L24

Should be `colspan="2"`, right?

The table header has 2 columns but the table body has 3 columns.

image

andreich1980:
The table header has 2 columns but the body has 3.

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -1034,6 +1034,14 @@ table.dataTable.no-footer{
   border-bottom:1px solid #f1f1f1;
 }
 
+table .th .no-sort::after { 
+    display: none !important; 
+}
+table .th .no-sort { 
+    pointer-events: none !important; 
+    cursor: default !important; 
+}
+
 .code-preview {
   width: 100%;
   min-height: 400px; }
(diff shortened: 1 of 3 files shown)


emptynick:
Where is the class `no-sort` actually used?
Edit: Guess I was too fast

andreich1980:
@emptynick It seems I screwed up by adding the commits to a wrong PR :/

andreich1980:
@emptynick  `no-sort` is used on `td`'s with actions for `browse` BREAD pages.
I think it's wrong, it should be on the `th` and it seems that there's no `.no-sort` class anywhere in the styles

Should I create new commits to another PR (actually, here https://github.com/andreich1980/voyager/pull/1/commits)? Or there's a way to move the commits?

andreich1980:
the 2 wrong commits should be (and the changes already are) there #3249

fletch3555:
@andreich1980, you need to be working in branches, not committing directly to `1.1` on your fork.  This is why those commits are appearing in both PRs.  I'm going to close them both, and you can open new ones when you've committed the changes to their own branches

PR #3249, opened 2018-05-29
Actions column classes: no-click, no-sort

andreich1980:
* Moved ID from all the TD's to TH
* Moved no-sort, no-click classes from the TD's to TH


fletch3555:
Closing because: https://github.com/the-control-group/voyager/pull/3247#issuecomment-393214940

PR #3256, opened 2018-05-31
Enhancement: Browse actions partial

ratt:
Ability to include extra browse actions / html without having to override the whole browse template


fletch3555:
We already have the `voyager::bread.partials.actions` partial which renders each action button. The actions are defined elsewhere and are dynamic.

I don't really understand what you expect to add with this new partial that isn't already possible.  I'll need more justification before I can consider merging this.

fletch3555:
Closing due to lack of response.

PR #3259, opened 2018-05-31
Update russian database.php

p2p-sys:
In version 1.1.2 not all words are translated into Russian


fletch3555:
I rebased changes into your 1.1 branch.  Will track all RU language updates in #3257

PR #3260, opened 2018-05-31
Fix #3246

### LINKED ISSUE #3246: Fix front-end on the "Viewing Database" page

andreich1980:
thead has 2 columns but tbody has 3

Fixes #3246


andreich1980:
Instead of my previous PR #3247 that was wrong

emptynick (approved):

PR #3263, opened 2018-05-31
Make actions column header non-sortable

andreich1980:
Instead of my previous PR #3249 that I did wrong.

It fixes `.no-sort` classe on actions column header to prevent dataTable sorting by it.
Also fixes duplicate ID for `td`s in table rows.


resources/views/tools/database/index.blade.php
```diff
@@ -21,7 +21,7 @@
                     <thead>
                         <tr>
                             <th>{{ __('voyager::database.table_name') }}</th>
-                            <th style="text-align:right">{{ __('voyager::database.table_actions') }}</th>
+                            <th style="text-align:right no-sort">{{ __('voyager::database.table_actions') }}</th>
```
emptynick: The database-tool doesn't use a Datatable, so none of your classes will kick in here

emptynick: The database-tool doesn't use a Datatable, so none of your classes will kick in here

emptynick (changes requested):

emptynick:
As changing CSS is always breaking (needs republishing) why dont you simply use datatables `orderable` options?
Like
```
"columnDefs": [
    { "orderable": false, "targets": -1 }
]
```
See [here](https://datatables.net/reference/option/columns.orderable)

resources/views/tools/database/index.blade.php
andreich1980: oh, you're right. Thanks. What should I do now? Create one more commit that removes the class in this file?

emptynick: Yes, in the end `resources/views/tools/database/index.blade.php` shouldn't be changed at all.

resources/views/tools/database/index.blade.php
```diff
@@ -61,7 +61,7 @@ class="btn-sm btn-default">
                             </div>
                         </td>
 
-                        <td class="actions">
+                        <td class="actions bread-actions">
```
emptynick: `.bread-actions` is also unneccessary as its a child of `#dataTable`

emptynick (changes requested):

andreich1980:
@emptynick  It's done.

emptynick:
That looks much better.
Can you bring the `Actions` header to the right side in the menu-browse blade?

andreich1980:
@emptynick I could add `.actions { text-aligh: right; }` but you said it's a breaking change. 
Or do you want me just add `style="text-align: right;"` on the actions column header?

emptynick (approved):

PR #3267, opened 2018-06-01
Compass security fixes

### LINKED ISSUE #3032: Exploit in Compass -> Commands
Someone with access to the commands page can run any *nix command by passing `; cat /etc/passwd` or `&& cat /etc/passwd`. Has this been talked about removing those characters or will you just say "turn off permissions for viewing compass then"? :)

### LINKED ISSUE #3227: Security breaches on Voyager Compass
Hi all,

I'm opening this new issue because during a pentest for a customer using Voyager to administrate his Laravel backend, we found two security breaches on Voyager Compass tool.

The first security breach is in Command menu, used normally to execute Laravel commands. But when we put system commands on arguments (example : && cat /etc/passwd), this one is sucessfully executed and send the return in the web page. So, an attacker can take control of server if he has an access to backend. This security breach has a major criticity and CVSS score estimate to 7,2. To mitigate this vulnerability, special characters must be sanitized.

The second security breach is in the Log review menu. It is possible to download or delete system files if the user has system rights. Normally, it is possible to download or delete the log files from the URL https://mywebiste.com/admin/compass?download=bGFyYXZlbC5sb2c=. However, if we decode base64, we get the cleartext laravel.log. If you change the download parameter by an another name file encoded in base64 (for example /etc/passwd => L2V0Yy9wYXNzd2Q=), it is possible to download the file concerned. Similarly, it is possible to delete files using the URL https://mywebiste.com/admin/compass?del=bGFyYXZlbC5sb2c=. This vulnerability has a moderate criticity and a CVSS score estimate to 6,0. To mitigate this one, we recommend restricting access to certain types of log files.

Can you take a look at these problems that we have raised ?

If you need some details, we are at your disposal.

Kind regards.

Ricardo & Margot

jannejava:
Attempt to fix the security issues in #3227 by escaping the arguments to prevent appending commands. Also, removing the ability to send in absolute paths in the log-viewer and only use the check if the file is present in the log-directory.

Fixes #3032 
Fixes #3227


fletch3555:
Does this fix all of the issues in #3227?

fletch3555 (dismissed):

jannejava:
I hope so, at least on *nix environments. Wrote some tests to prove the case. Even if testing with Orchestral/TestBench and Artisan commands are not friends. Mocking base_path() seems not to be trivial.

jannejava:
I hope so. 

Saw that I made a mistake and escaping empty arguments, fixed this now.

Wrote some tests to prove the case. Even if testing with Orchestral and Artisan commands are not that easy.

src/Http/Controllers/VoyagerCompassController.php
```diff
@@ -62,12 +63,20 @@ public function index(Request $request)
...
+            $command = escapeshellcmd($request->command);
+            $args = (isset($request->args)) ? ' '.escapeshellarg($request->args) : '';
 
             try {
-                $process = new Process('cd '.base_path().' && php artisan '.$command.$args);
+                if (App::environment('testing')) {
+                    // hard to mess with base_path from Orchestral
+                    // without causing trouble on other tests.
+                    // Stepping back to the main application
+                    $path = realpath(base_path('../../../../../../../'));
```
fletch3555, line 74: This path is a few levels too high and is causing TravisCI to fail.  What are you expecting the starting directory will be? What about the path to `artisan`?

Why not use the `Artisan` facade instead?

fletch3555 (changes requested):

src/Http/Controllers/VoyagerCompassController.php
jannejava: Using the facade would be nice and appropriate here. Using `Artisan:call()` would be better I think. Laravel will then handle illegal calls to non-Artisan commands and options also prevent shell injections. 

My guess is the challenge is how to parse and make an array of non-optional arguments from the input string, also knowing the parameters of a particular command. That's maybe the reason?  Example from the docs:

```php
    $exitCode = Artisan::call('email:send', [
        'user' => 1, '--queue' => 'default'
    ]);
```

Testing `Artisan:call()` and it works for all the simple commands like `artisan help --version` for me, but I guess more advanced arguments with parameters will not. Going with  `Artisan:call()` and provide less advanced options/arguments for added security is an option?

emptynick: I guess thats as simple as
```
$in = "--arg1=yes --arg2=no --arg3=whatever";

$parameter = [];
$arguments = explode('--', $in);

foreach ($arguments as $arg) {
    $arg = trim($arg);
    if ($arg != '') {
        $parts = explode('=', $arg);
        //Here we could test $parts[0] against all available arguments of the command
        $parameter[$parts[0]] = $parts[1];
    }
}

dd($parameter);
```

Maybe there is a nicer way with Laravel Collections.

You can get all available arguments by calling `Artisan::call('...', ['--help']);`
Then just process them like the available commands.

jannejava: That would work for most of the commands. But if you look at the signature for e.g. migrations which takes more than an option. 

```php
protected $signature = 'make:migration {name : The name of the migration.}'
```

To run this you must know that the argument's name before in order to build the array. 

```php
    $exitCode = Artisan::call('make:migration', [
        'name' => 'the_name_of_migration'
    ]);
```

You cannot just send an array like this:

```php
    $exitCode = Artisan::call('make:migration', [
        'the_name_of_migration'
    ]);
```

But if you specify this in the argument like name=the_name_of_migration, it would work.

dexcell:
this is not fixed yet?

emptynick:
Closing as superseeded by https://github.com/the-control-group/voyager/pull/3823

PR #3268, opened 2018-06-01
Fix column not found when using additional_attributes w/ xMany relations

SteamWind:
Fix for belongsToMany and hasMany relationship display with
additional_attributes. You can now use additional_attributes with those
relations.
#2885

The fix get the model before plucking it. It caused error because when
you pluck the query builder, it affects the mysql query. When you use
additional attribute, it doesn’t exist in the database and result with
a `SQLSTATE[42S22]: Column not found: 1054 Unknown column ‘youradditionalattribute’ in 'field list’` error…

This is a fresher repost from #3233

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -67,7 +67,7 @@
 				@php
 					$relationshipData = (isset($data)) ? $data : $dataTypeContent;
 					$model = app($options->model);
-            		$selected_values = $model::where($options->column, '=', $relationshipData->id)->pluck($options->label)->all();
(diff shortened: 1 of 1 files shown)


SteamWind:
How is going the QA?

diegoRivero (approved):

emptynick:
Sorry for the late response, but I think this was also fixed in #3444.
I'm going to close this for now. If it isn't fixed please let us know and we reopen this one. Thanks.

SteamWind:
Yeah, seems fixed in a different way, thanks.

PR #3269, opened 2018-06-01
French Translation Updated

SteamWind:
Completing and updating the French translation.


publishable/lang/fr/database.php
```diff
@@ -51,30 +44,31 @@
...
         'relationship'         => 'Relation (Relationship)',
         'relationships'        => 'Relations (Relationships)',
-        'has_one'              => 'En a un (Has One)',
-        'has_many'             => 'En a pluieurs (Has Many)',
-        'belongs_to'           => 'Appartient à (Belongs To)',
-        'belongs_to_many'      => 'Appartient à pluieurs (Belongs To Many)',
-        'which_column_from'    => 'Quelle colonne du',
-        'is_used_to_reference' => 'est utilisée pour référencer le',
-        'pivot_table'          => 'Tableau croisée dynamique (Pivot Table)',
+        'has_one'              => null,
```
fletch3555: Any particular reason these are null?

SteamWind: Yes, because it shouldn't be translated... 
Null = use default language
Not present = not translated

fletch3555: @emptynick tested it and it looks like null value and missing key behave the same.  Can you verify this and remove the entry? I feel the null value may get confusing and difficult to maintain long-term.  I'd rather have it be missing entirely, which indicates it needs a translation (or should be left alone)

SteamWind: For me, not having a value or value set on Null use the default translation (EN). That's weird.
Anyway I commit the translation of those values in EN.

SteamWind:
Hey, does I missed something? Is it okay to merge those translations?
I made the changes on the null values.

fletch3555 (approved):

PR #3279, opened 2018-06-06
Resizeable support for multiple images

handiwijoyo:
This will allow us to pass any image path and generate the thumbnail using `get_thumbnail`. This is useful when using it inside of multiple images array / input type. For example:
```
@foreach($posts as $post)
    @foreach($post->images as $image)
        <img src="{{Voyager::image($post->get_thumbnail($image, 'small'))}}" />
    @endforeach
@endforeach
```

If you are using multiple images input type, you can optional cast your images attribute
```

class Post extends Model
{
    use Resizable;

    protected $casts = [
        'images' => 'array',
    ];
}
```

### DIFF
--- a/src/Traits/Resizable.php
+++ b/src/Traits/Resizable.php
@@ -22,6 +22,17 @@ trait Resizable
         // We take image from posts field
         $image = $this->attributes[$attribute];
 
+        return $this->get_thumbnail($image, $type);
+    }
+
+    /**
+     * Generate thumbnail URL
+     * @param $image
+     * @param $type
+     * @return string
+     */
+    public function get_thumbnail($image, $type)
+    {
         // We need to get extension type ( .jpeg , .png ...)
         $ext = pathinfo($image, PATHINFO_EXTENSION);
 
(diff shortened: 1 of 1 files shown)


src/Traits/Resizable.php
```diff
@@ -22,6 +22,19 @@ public function thumbnail($type, $attribute = 'image')
...
+
+    /**
+     * Generate thumbnail URL.
+     *
+     * @param $image
+     * @param $type
+     *
+     * @return string
+     */
+    public function get_thumbnail($image, $type)
```
marktopper, line 36: I think it would be more clear if we used `camelCase` and went with `getThumbnail ` over `get_thumbnail`. Thoughts?

marktopper (changes requested):

src/Traits/Resizable.php
handiwijoyo: Thanks @marktopper I will change that. I will open new PR too, because my mistake that create new branch to main repo instead of my own fork

handiwijoyo:
I made new PR #3288. Close this one

PR #3280, opened 2018-06-06
Database-Tool default charset from config

### LINKED ISSUE #2745: Enhancement Request: utf8mb4 - utf8mb4_unicode_ci as default collation in MySQL
- Laravel Version: 5.6.5
- Voyager Version: dev-1.x-dev
- PHP Version: 7.2.2
- Database Driver & Version: MySQL 5.7.21 Homebrew

### Description:
When using MySQL, it seems that creating a new database table using the Voyager UI does so with a `utf8` character set and `utf8_unicode_ci` collation. This goes against the [Laravel default](https://laravel.com/docs/5.6/migrations#creating-indexes) and has [caused](https://news.ycombinator.com/item?id=14747699) [so](https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql) [many](https://www.eversql.com/mysql-utf8-vs-utf8mb4-whats-the-difference-between-utf8-and-utf8mb4/) [headaches](https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434) for MySQL users (including myself).

I'd like to propose that Voyager creates tables with a default character set of `utf8mb4` and collation of `utf8mb4_unicode_ci` when the database driver is MySQL or one of its variants (MariaDB, PerconaDB, etc.).

emptynick:
Tables created by the Database-Tool were always UTF8, so you had to go to your Database-Manager (phpmyadmin etc.) to change it.
This PR makes all newly created tables use the charset/collation setted in the default-database-connection or defaults to `utf8mb4` which is the standard since Laravel 5.4 [ref](https://laravel.com/docs/5.4/migrations#creating-indexes)
Resolves #2745


marktopper (approved):
Just tested this on my local environment, and it works like a charm.

fletch3555:
I think we should note this in the docs as well.  @emptynick, could you take care of that?

Issue #2333, opened 2017-12-10
Suggestion link of items


vtmx:
### Sugestion:
I have a suggestion to add the items of the tables as links, so when clicked go to edit them.

![post-item-with-link](https://user-images.githubusercontent.com/8462885/33807391-65d31e24-ddbd-11e7-9094-8e80480f8983.jpg)

fletch3555:
We welcome PRs for all suggestions.  I wouldn't make the text links, however, but instead just have JS link the whole <tr>

handiwijoyo:
I think we can add an option to convert that text field to link using something like:
```
{
    linkToEdit: true, // or
    linkToRead: true,
}
```

PR #3282, opened 2018-06-07
Add double click function to BREAD row

marijnz0r:
Add a way the user can navigate through the BREAD views. By double
clicking the row, the user goes to the view or edit page, depending
on their rights. To simplify the way to delete a single item, add a
delete button in the browse view if the user has rights to delete it.

Fixes https://github.com/the-control-group/voyager/issues/2333

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -88,7 +88,7 @@
                                 </thead>
                                 <tbody>
                                 @foreach($dataTypeContent as $data)
-                                    <tr class="js-row" data-id="{{ $data->getKey() }}" >
+                                    <tr class="js-bread-data-row" data-id="{{ $data->getKey() }}" >
                                         @can('delete',app($dataType->model_name))
                                             <td>
                                                 <input type="checkbox" name="row_id" id="checkbox_{{ $data->getKey() }}" value="{{ $data->getKey() }}">
@@ -287,7 +287,7 @@
         $('#delete_modal').modal('show');
     });
 
-    $('tr.js-row').on('dblclick', function (e) {
+    $('tr.js-bread-data-row').on('dblclick', function (e) {
         var doubleClickLocation = "{{lcfirst($dataType->display_name_plural)}}" + "/" + $(this).data('id');
         @can('edit', app($dataType->model_name))
             doubleClickLocation += "/edit";
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -16,11 +16,11 @@
             @include('voyager::partials.bulk-delete')
         @endcan
         @can('edit',app($dataType->model_name))
-        @if(isset($dataType->order_column) && isset($dataType->order_display_column))
-            <a href="{{ route('voyager.'.$dataType->slug.'.order') }}" class="btn btn-primary">
-                <i class="voyager-list"></i> <span>{{ __('voyager::bread.order') }}</span>
-            </a>
-        @endif
+            @if(isset($dataType->order_column) && isset($dataType->order_display_column))
+                <a href="{{ route('voyager.'.$dataType->slug.'.order') }}" class="btn btn-primary">
+                    <i class="voyager-list"></i> <span>{{ __('voyager::bread.order') }}</span>
+                </a>
+            @endif
         @endcan
         @include('voyager::multilingual.language-selector')
     </div>
@@ -38,7 +38,7 @@
                                 <div id="search-input">
                                     <select id="search_key" name="key">
                                         @foreach($searchable as $key)
-                                                <option value="{{ $key }}" @if($search->key == $key){{ 'selected' }}@endif>{{ ucwords(str_replace('_', ' ', $key)) }}</option>
+                                            <option value="{{ $key }}" @if($search->key == $key){{ 'selected' }}@endif>{{ ucwords(str_replace('_', ' ', $key)) }}</option>
                                         @endforeach
                                     </select>
                                     <select id="filter" name="filter">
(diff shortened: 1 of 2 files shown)


SteamWind:
Could be cool, but you modified a lot of lines by reindenting them. You shouldn't commit spaces changes I think.

resources/views/bread/browse.blade.php
```diff
@@ -241,50 +241,58 @@
...
-        <script src="{{ voyager_asset('lib/js/dataTables.responsive.min.js') }}"></script>
+        <link rel="stylesheet" href="{{ voyager_asset('lib/css/responsive.dataTables.min.css') }}">
     @endif
-    <script>
-        $(document).ready(function () {
-            @if (!$dataType->server_side)
-                var table = $('#dataTable').DataTable({!! json_encode(
+@stop
+
+@section('javascript')e
```
emptynick: `e` ?

emptynick (changes requested):

marijnz0r:
@SteamWind I tried to improve the indenting to make lots of lines more readable. Do you think I should undo the improvements or leave it as it is right now?

SteamWind:
StyleCI should do it automatically. But it have some trouble doing it with blade. Maybe your changes won't affect the merge... StyleCI will undo it I think.

marijnz0r:
Hey @emptynick,
I corrected my mistake, could you give it another review?
Kind regards

marktopper (dismissed):
Looks good for me

resources/views/bread/browse.blade.php
```diff
@@ -66,29 +66,29 @@
...
+                                            </th>
                                         @endforeach
                                         <th class="actions text-right">{{ __('voyager::generic.actions') }}</th>
                                     </tr>
                                 </thead>
                                 <tbody>
-                                    @foreach($dataTypeContent as $data)
-                                    <tr>
+                                @foreach($dataTypeContent as $data)
+                                    <tr class="js-row" data-id="{{ $data->getKey() }}" >
```
fletch3555: Is `js-row` a name you came up with? If so, it's far too generic.  Please come up with a more specific name.

resources/views/bread/browse.blade.php
```diff
@@ -241,50 +241,58 @@
...
+    var deleteFormAction;
+    $('td').on('click', '.delete', function (e) {
+        $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['id' => '__id']) }}'.replace('__id', $(this).data('id'));
+        $('#delete_modal').modal('show');
+    });
+
+    $('tr.js-row').on('dblclick', function (e) {
+        var doubleClickLocation = "{{lcfirst($dataType->display_name_plural)}}" + "/" + $(this).data('id');
+        @can('edit', app($dataType->model_name))
+            doubleClickLocation += "/edit";
```
fletch3555, line 293: I think varying the click action like this might be confusing to end-users.

I believe the `read` view also has an edit button, can we simply link them to the read page and they can edit if desired?

I'm open to suggestions on this.

resources/views/bread/edit-add.blade.php
```diff
@@ -83,9 +83,31 @@ class="form-edit-add"
 
                         <div class="panel-footer">
                             <button type="submit" class="btn btn-primary save">{{ __('voyager::generic.save') }}</button>
+                            <button type="button" class="btn btn-danger js-delete">{{ __('voyager::generic.delete') }}</button>
                         </div>
                     </form>
 
+                    {{-- Single delete modal --}}
+                    <div class="modal modal-danger fade" tabindex="-1" id="delete_modal" role="dialog">
```
fletch3555, line 93: Why do you have a "delete" modal here, when there's one just a few lines below this? How are they different?

resources/views/bread/edit-add.blade.php
```diff
@@ -116,8 +138,9 @@ class="form-edit-add"
 
                 <div class="modal-footer">
                     <button type="button" class="btn btn-default" data-dismiss="modal">{{ __('voyager::generic.cancel') }}</button>
-                    <button type="button" class="btn btn-danger" id="confirm_delete">{{ __('voyager::generic.delete_confirm') }}
-                    </button>
+                    @can('delete', $dataTypeContent)
```
fletch3555: What's the point of this permissions check? You're already in the modal.  Might as well put the check around the entire modal at that point.

resources/views/bread/edit-add.blade.php
```diff
@@ -166,6 +189,11 @@ class="form-edit-add"
                 $('#confirm_delete_modal').modal('show');
             });
 
+            $('.js-delete').on('click', function (e) {
+                $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['id' => '__id']) }}'.replace('__id', $(this).data('id'));
```
fletch3555: I think I'd rather see this use `.get(0)` instead of `[0]`.  I realize this code was likely copied from somewhere, but if you could update that, I'd appreciate it.

resources/views/bread/edit-add.blade.php
```diff
@@ -83,9 +83,31 @@ class="form-edit-add"
 
                         <div class="panel-footer">
                             <button type="submit" class="btn btn-primary save">{{ __('voyager::generic.save') }}</button>
+                            <button type="button" class="btn btn-danger js-delete">{{ __('voyager::generic.delete') }}</button>
```
fletch3555: Again, not really a fan of the generic class naming `js-delete`.  Please come up with something more specific.

fletch3555 (changes requested):

resources/views/bread/browse.blade.php
marijnz0r: I thought about this as well. The thing that made me choose to do it this way is the speed of usage of the admin panel for the end user. If I'm double clicking a row and I do have edit rights, I'm almost never interested in just showing the record, but almost always in directly editing it. If we always redirect the user to the `read` view, the user who wants to edit the record always has to perform another click. The way it's made right now is faster.

resources/views/bread/edit-add.blade.php
marijnz0r: The difference is that the previously existing modal is only for deleting (multiple) media/images.
I didn't want to change too much there, as I don't know the inner workings of that flow.

resources/views/bread/edit-add.blade.php
marijnz0r: Well spotted! I made a mistake here, wanted to put the permissions check around the delete button down in the footer. So after the update for this pull request it doesn't even show the delete button if you don't have permission to delete the record.

resources/views/bread/browse.blade.php
SteamWind: This kind of dilemma could be an option from the user and we could implement both...? Just a simple suggestion.
My usage would be to have a better global view of a record in table with lot of column. I don't specially want to edit it. But as the view mode isn't really different as the edit, it doesn't matter much to me...

emptynick:
The delete part was token care of elsewhere. I suggest you remove it completely.

emptynick:
Closing as this PR wasn't touched for nearly a year and there are still requested changes.
If someone else want's to implement this functionality please open a new PR.

PR #3286, opened 2018-06-08
Use slug instead of slugified table name

SteamWind:
The `relationship.blade.php` view was using _slugified table name_ instead of the stored _URL Slug_ to access the store view.
This caused error when personalising BREAD url slug and having a belongsToMany relation pointing to it in an other BREAD.

### If you want to reproduce:

1. Create a BREAD with custom URL Slug (B1)
1. Create a BREAD (B2) with a belongs to many relation to the table of B1
1. Go to B2 and try to edit one row of data

### This throws this error:
`Route [voyager.slugified-table-name.store] not defined. (View: vendor/tcg/voyager/resources/views/formfields/relationship.blade.php) (View: vendor/tcg/voyager/resources/views/formfields/relationship.blade.php)`

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -148,7 +148,7 @@
 					class="form-control @if(isset($options->taggable) && $options->taggable == 'on') select2-taggable @else select2 @endif" 
 					name="{{ $relationshipField }}[]" multiple
 					@if(isset($options->taggable) && $options->taggable == 'on')
-                        data-route="{{ route('voyager.' . $dataType->slug . '.store') }}"
+						data-route="{{ route('voyager.' . $dataType->slug . '.store') }}"
 						data-label="{{$options->label}}"
 						data-error-message="{{__('voyager::bread.error_tagging')}}"
 					@endif
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -148,7 +148,7 @@
 					class="form-control @if(isset($options->taggable) && $options->taggable == 'on') select2-taggable @else select2 @endif" 
 					name="{{ $relationshipField }}[]" multiple
 					@if(isset($options->taggable) && $options->taggable == 'on')
-						data-route="{{ route('voyager.'.str_slug($options->table).'.store') }}"
+                        data-route="{{ route('voyager.' . $dataType->slug . '.store') }}"
 						data-label="{{$options->label}}"
 						data-error-message="{{__('voyager::bread.error_tagging')}}"
 					@endif


resources/views/formfields/relationship.blade.php
```diff
@@ -148,7 +148,7 @@
 					class="form-control @if(isset($options->taggable) && $options->taggable == 'on') select2-taggable @else select2 @endif" 
 					name="{{ $relationshipField }}[]" multiple
 					@if(isset($options->taggable) && $options->taggable == 'on')
-						data-route="{{ route('voyager.'.str_slug($options->table).'.store') }}"
+						data-route="{{ route('voyager.' . $dataType->slug . '.store') }}"
```
cyrildewit, line 151: I think this is not right.

My setup:
* B1: slug='events-test'
* B2: slug='applicants'
* B2 Belongs to Many Events (taggable=true)

When I dump the content of `$dataType->slug`, the output is: `applicants`. This should be `events-test`. 

```text
// Editing an applicant

@dump(route('voyager.'.str_slug($options->table).'.store'))) // => resulted in an expected error
@dump(route('voyager.'.$dataType->slug.'.store')) // => "http://sb-voyager-app.test/admin/applicants"
```

emptynick: Take a look here:
https://github.com/the-control-group/voyager/blob/1.1/src/Actions/ViewAction.php#L31

cyrildewit: I digged a little bit more into the code. The `data-route` value is used when the taggable option is enabled in a relationship. This feature allows us to create an new item instantly within the select box instead of manually creating a new event. This route must go to the `events-test.store` route because we're creating new `events` in this example. Otherwise it will create new applicants.

cyrildewit: I was able to solve it with: https://github.com/cyrildewit/voyager/blob/c558ede4f5d9a1f04f84277a91d527f2d1499472/resources/views/formfields/relationship.blade.php#L152

SteamWind: Hum, I don't understand everything but when we tag an event in B2, we need to store the value in B1 (Slug: `event-test`), you don't get this value?
I'll try it again

SteamWind: So if I put `Voyager::model('DataType')->where('name', $options->table)->first()->slug;` instead, it should work...

SteamWind: I properly retested my solution and it works well... I don't understand what's the problem.

SteamWind: Okay, both `Voyager::model('DataType')->where('name', $options->table)->first()->slug` and `$dataType->slug` works 🤔 
@cyrildewit 's solution use the same logic as `$option->table`. Mine use the opposite logic.

SteamWind:
If you didn't made a BREAD for the relation you are pointing at, `Voyager::model('DataType')->where('name', $options->table)->first()->slug` trows an error. (property of non-object)

PR #3287, opened 2018-06-08
Replace tabs with spaces and restructured some HTML in blade files to gain more readability

cyrildewit:
I've done the following things:

* Replaced tabs with spaces in some files
* Fixed indentation of some blade files
* Did some code style changes

It's hard to review this pull request using the unified option under diff settings. I suggest you to use the split option.


cyrildewit:
Some more things that I want to discuss:

1. I wonder if I can remove the out commented line at https://github.com/the-control-group/voyager/pull/3287/files?utf8=%E2%9C%93&diff=unified#diff-cf377d287d89be3fc7d2a9c234b13774L115

2. Can I replace `<?php echo .... ; ?>` with blade mustache symbols? https://github.com/the-control-group/voyager/pull/3287/files#diff-ae7663e73e47fb2ae5a55b94af230685R68

3. Can I remove the `console.log` at https://github.com/the-control-group/voyager/pull/3287/files#diff-ae7663e73e47fb2ae5a55b94af230685R180

4. I couldn't find any usage of the `bread/partials/relationship-edit-add.blade.php` file. @emptynick already told me it was a mistake. Can I remove this?

Thanks!

---

- [x] Point 1 / _Removed the line_
- [x] Point 2 / _Skipped this improvement`_
- [x] Point 3 / _Removed the line_
- [x] Point 4 / _Removed the file_

fletch3555:
1. Go ahead and remove the commented line.
2. I'm not sure that line should even be there, since it's between two `@elseif` lines, so it _should_ be invalid code.
3. Yes, console.log can be removed
4. I swear I saw a usage of that just the other day, but I can't seem to find it right now.  I'd say it's probably a mistake and fine to delete as well.

Thanks!

emptynick:
I deleted this file for 1.1 already, but it slipped back in when I splitted the language files.

resources/views/bread/browse.blade.php
```diff
@@ -223,13 +222,14 @@
         <div class="modal-dialog">
             <div class="modal-content">
                 <div class="modal-header">
-                    <button type="button" class="close" data-dismiss="modal" aria-label="{{ __('voyager::generic.close') }}"><span
-                                aria-hidden="true">&times;</span></button>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="{{ __('voyager::generic.close') }}">
+                        <span aria-hidden="true">&times;</span>
```
fletch3555: This (and the couple others like it) are probably fine to be on one line.  I believe whitespace within a `<button>` element is respected, so it surrounding the span could potentially cause issues, primarily in older browsers.  There's no certainty that it will, but I'm okay with following that pattern.

resources/views/bread/read.blade.php
```diff
@@ -130,18 +130,20 @@
...
+
     {{-- Single delete modal --}}
     <div class="modal modal-danger fade" tabindex="-1" id="delete_modal" role="dialog">
         <div class="modal-dialog">
             <div class="modal-content">
                 <div class="modal-header">
-                    <button type="button" class="close" data-dismiss="modal" aria-label="{{ __('voyager::generic.close') }}"><span
-                                aria-hidden="true">&times;</span></button>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="{{ __('voyager::generic.close') }}">
+                        <span aria-hidden="true">&times;</span>
```
fletch3555: Same here

fletch3555 (changes requested):
I'm sure this would trigger a tabs v. spaces debate, but I honestly prefer tabs.  For the framework, however, I'm fine with forcing everything to spaces.  Having a mix of both is where it gets ugly.

Just be cognizant of that debate when making these kinds of whitespace changes elsewhere.  Others may not be so accepting. ;)

cyrildewit:
@fletch3555 thanks for the review. I'm aware of that, but since this project doesn't have a code rules guide, I have to guess which rules this project is following. Most of the code is indented with spaces, so that's the reason why I choosed spaces over tabs.

About point 2: ~I think that the line is correct. I have updated it to use `{{` & `}}`.~

fletch3555 (approved):

PR #3288, opened 2018-06-09
Resizeable support for multiple images

handiwijoyo:
This will allow us to pass any image path and generate the thumbnail using `getThumbnail`. This is useful when using it inside of multiple images array / input type. For example:
```
@foreach($posts as $post)
    @foreach($post->images as $image)
        <img src="{{Voyager::image($post->getThumbnail($image, 'small'))}}" />
    @endforeach
@endforeach
```

If you are using multiple images input type, you can optional cast your images attribute
```

class Post extends Model
{
    use Resizable;

    protected $casts = [
        'images' => 'array',
    ];
}
```


marktopper (approved):

fletch3555:
Just a word of caution against using your `1.1` branch for PRs.  We squash-merge PRs, which changes the git history, leading to you potentially getting into some odd places.  Always try to use feature branches when submitting PRs.

fletch3555:
We'll also need docs for this.  Would you mind writing up something and suggesting it on readme.io?

fletch3555 (approved):

handiwijoyo:
@fletch3555 I've sent edit suggestion on readme.io. Mind to check it? Thanks

PR #3292, opened 2018-06-11
[WIP] Add the ability to configure a custom base direction per user

cyrildewit:
This pull request will make it possible to configure a custom base direction per user. Each user can change the direction inside his profile.

Instead of calling it `dir` (like in HTML), I decided to call it `base direction` to be more descriptive.

It's a work in progress, so I'm interested in some feedback from the maintainers and of course other contributors.

Some things where I want to think about:
- Make this feature optional (should a user be able to change his base direction)
- Create a hook instead of adding it to the core
- A cleaner solution for: `config()->set('voyager.multilingual.rtl', $user->base_direction === 'rtl');`

I also had an idea about determining the direction based upon the user's locale. I'll keep you informed!

If you're not interested in this feature, please let me know!
Thanks!

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -13,6 +13,8 @@ return [
     'user' => [
         'add_default_role_on_register' => true,
         'default_role'                 => 'user',
+        // available directions: `ltr` or `rtl`
+        'default_base_direction'       => 'ltr',
         // Set `namespace` to `null` to use `config('auth.providers.users.model')` value
         // Set `namespace` to a class to override auth user model.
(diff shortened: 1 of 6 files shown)


emptynick:
> I also had an idea about determining the direction based upon the user's locale. I'll keep you informed!

Definitely a good point. This ultimately defines if the user wants/needs RTL or not.
I also think its way easier.
I dont know if its "valid" to just add a string into the language-file, something like:
`'is_rtl' => true`
But then you'd just need to do something like:
`<html dir="{{ __('voyager.is_rtl') ? 'rtl' : '' }}">`

PR #3305, opened 2018-06-13
now you are not forced to make table name in plural

simoebenhida:
You always need to specify the table name as plural but now you are not forced to do that.


fletch3555:
Apologies for the very long delay in response.... but what does this solve that isn't solved by adding `protected $table = 'singularName'` to the model class?

fletch3555:
Please feel free to reopen this once you've answered my question above

PR #3306, opened 2018-06-13
Improve RTL version of login page

amirmasoud:


emptynick:
I just changed your base-branch from `1.1` to `1.x` because the css file needs publishing. (And I by mistake merged 1.1 into your master, sorry for that)

By the way, cant you just [flip the image](https://www.w3schools.com/howto/howto_css_flip_image.asp)?

amirmasoud:
@emptynick I have tried it, it cause whole body to flip, background-color is set on body tag, for this solution I needed more changes to be done, I just did the easiest way.

resources/views/login.blade.php
```diff
@@ -14,7 +14,11 @@
     @endif
     <style>
         body {
+            @if (config('voyager.multilingual.rtl'))
```
emptynick, line 17: This config doesn't exist in 1.x anymore.
Please refer to this line:
https://github.com/the-control-group/voyager/blob/bb9646169ac30e2b230495c82bab42d89cb4f717/resources/views/master.blade.php#L2

emptynick (changes requested):

resources/views/login.blade.php
emptynick, line 17: Thats how we decide wether to use RTL or not.
```suggestion
            @if (__('voyager::generic.is_rtl') == 'true')
```

emptynick (changes requested):

emptynick (changes requested):
I'm not really happy with having another image just so its mirrored.
In my opinion this can be done through CSS.

emptynick:
Closing as the fork was deleted.

PR #3322, opened 2018-06-17
Fix migration rollback error because of wrong type of field

taida957789:
Fix the error occurred when rollback the migration


mnogokotin:
add please

emptynick:
Already fixed in https://github.com/the-control-group/voyager/pull/3974

PR #3323, opened 2018-06-18
Support multiply databases

### LINKED ISSUE #1703: Multiple Database
- Laravel Version: 5.4.*
- Voyager Version: ^0.11.14
- PHP Version: 7.1.8
- Database Driver & Version:  MariaDB 5.5.52

### Description:

I do have multiple databases which works quite well within a  Laravel App. 

How can I configure Voyager to use them?

Is it possible to switch between connections?

Kind Regards

### Steps To Reproduce:
 - new laravel project and installation of voyager with all required steps
 - create a new databases
 - setup new database connections in database.php config
 - just default one is present

### LINKED ISSUE #2055: Voyager is great ! Do you have a plan to support multi database?
- Laravel Version: 5.5.13
- Voyager Version: 1.0.5
- PHP Version: 7.1
- Database Driver & Version: mysql5.7

### Description:

I use voyager to create a admin. but my other database can't use bread builder. 

i hope voyager to support multi database and mongodb.

freedom-shelby:
Fixes #2055
Fixes #1703

Adding new config set to the config.php file that was mean to add remote databases

todo:: On database panel when add new table need to add select where user can set the database
todo:: FIX problem when databases have tables with same names


Valentyn-Kubrak:
Hi, I need this feature. When will you add this?

chorton:
This is now a year old.  Is this planned to merge and include?

emptynick:
There are todos open. Not until they are solved.

emptynick:
Closing due to inactivity.

Please feel free to reopen or grab this fork and continue developing.
But note that this will likely be unnecessary for Voyager 2.

wildtv:
Is your solution for V2 much different than what he's attempted above? 
I'm getting to the point really wanting this functionality as well.

What are you thoughts @emptynick?
If his changes are more or less still relevant to todays code base, would you rather hack this in or migrate the datatable?
I'm not sure what your thoughts are in terms of implementation in a 1.x world.

IMO he's missing a 'connection' field in the bread data table, to store which database connection to use.
Could avoid this migration potentially and store the extra database tables with the database name as a prefix?
Any current tables could lack such prefix and operate as normal. "mytable1", "database2.mytable1"

If the bread methods see a prefix in the table name it can explode and make the proper $connection and strip the prefix from the table name. 
It's a slightly hacky solution, but would suffice until you develop policy for a cleaner solution foe V2.

Beyond that it looks like he might need a select box of connections on the database page, which would switch between connections/pass to the bread pages

Do you foresee any major other problems than the above todo's?

I'd really prefer not to fork out,  or patch controllers along the way to have the best of both worlds :) 
Continuing this might be an option.

I know you all prefer Slack, but it's hard to maintain any development history in a chat room.

PR #3335, opened 2018-06-19
Suggestion: Setting options to datetimepicker

jannejava:
Browsers like Safari that does not support date and time input types, the fallback is to Bootstrap datetimepicker.

By default, English localization is used which is quite different than eg Swedish (the week starts a Monday, YYYY-MM-DD and 24 hours format). Added an option to set the locale on the options field.

This field can take all the options Bootstrap DateTime Picker takes. 
https://eonasdan.github.io/bootstrap-datetimepicker/Options/

Add an option in the BREAD-editor
locale

Datepicker displayed in Swedish locale.
swedish


package.json
```diff
@@ -23,7 +23,6 @@
     "animate.css": "^3.5.2",
     "bootstrap": "^3.3.7",
     "bootstrap-datepicker": "^1.6.4",
-    "bootstrap-datetimepicker": "0.0.7",
```
fletch3555: Duplicate changes from #3334.

emptynick:
Why dont we determine the locale based on `app()->getLocale()` instead of a "hardcoded" single, non-translatable value?

`json_encode(collect($options->datetimepicker)->merge('locale', app()->getLocale()))`

package.json
jannejava: They are dependent of these. I was not sure who to do a merge request on a merge request

jannejava:
I agree about the locales, but it's nice to customize the picker setting max-min dates, disabled dates and so forth.

trungtnm (dismissed):

emptynick:
I changed your base-branch to 1.x and fixed the merge conflicts.

PR #3377, opened 2018-07-06
Fix storage config to respect app config

BetaFold3:
As subject, we would better to respect application config for storage.
Default value still is `public`.


fletch3555:
Bumping this to `1.x`.  This change requires republishing the config file, and would break existing sites if they use a different "disk" for Voyager vs the rest of their site.

If anyone wants this in the mean time, feel free to manually edit your config file to include this change.

fletch3555 (approved):

tnguyenns1:
Woot woot!

PR #3378, opened 2018-07-06
Add a choice to keep original filename after uploading

BetaFold3:
Default behavior still is application will introduce a random name for uploaded file.
We add a choice/config so that users could have their application try to keep original uploaded file name, unless that name is duplicated.

This code works well in my system for a while, even with S3 Adapter.

Thanks.

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -223,4 +223,8 @@ return [
          'zoom' => env('GOOGLE_MAPS_DEFAULT_ZOOM', 11),
      ],
 
+     'media' => [
+         'keep_filename' => false,
+     ],
+
 ];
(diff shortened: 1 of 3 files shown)


fletch3555:
This only applies to the Media manager page, not to files/images uploaded through the form fields.

Form fields [use a JSON option](https://github.com/the-control-group/voyager/blob/1.1/src/Http/Controllers/ContentTypes/File.php#L55), but that won't work with this page.  I'd rather not have different means of achieving the same result.  @emptynick, what do you think?

dolcy:
@fletch3555 This is what I was referrring to in Slack; devs need a choice to retain the original filename with the media manager. The default random hash is great for full-blown apps, but confusing for general users i.e. a cms editor.

Moreover, a pdf file does not render a preview  (unlike an image) so it is hard to identify by random hash.

emptynick:
Closing in favor of #3860 (merged into 1.x, to be released in a future 1.2 release)

PR #3390, opened 2018-07-11
Update builder.blade.php

KarimQaderi:
fix show lang modal


fletch3555:
We need more than this as justification for the change.  Why are you requesting it?  What does it fix?

KarimQaderi:
fix1




Now after putting the code

fix2

fletch3555 (approved):

PR #3417, opened 2018-07-24
Where translation

### LINKED ISSUE #3675: Translated Slug from post
Hello, 

I have a few posts and just added yours multilanguage that is very easy to activate and use.
The issue is that I'm showing on the frontend the translated slug too, and when trying to open a post with the translated slug I get 404 even having that translated slug on the post.

Could you help me here, or at least guide me on the right direction.

Best Regards

reageek:
Adds functionality to allow searching for translated values in fields. 

The primary use case I had was for this functionality was for finding translated content via slugs. However it could be used for any kind of search for value.

The two primary uses are a static call & QueryScope

`$class::whereTranslation($args)`

or

`$class->whereTranslation($args)`

The result is a query builder instance.

Example useage

```
class PageController extends Controller
{   
    ...

    /**
     * Display the specified page or 404.
     *
     * @param string $slug
     * @return \Illuminate\Http\Response
     */
    public function show($slug)
    {   
        $page = Page::whereTranslation( $this->locales, 'slug', $slug );
        abort_unless($page, 404);
        $page = $page->withTranslations($this->locales)->active()->first();
        return $this->RenderView($this->_pageData($page);
    }
```
Closes #3675


src/Traits/Translatable.php
```diff
@@ -244,6 +245,81 @@ public function setAttributeTranslations($attribute, array $translations, $save
...
+     * @param string|array  $locale     {required} locale(s) you are looking for the field.
+     * @param string        $field      {required} the field your looking to find a value in.
+     * @param string        $operator   {required} value you are looking for or a relation modifier.
+     * @param string        $value      {optional} value you are looking for. Only use if you supplied an operator.
+     * @param bool          $default    {optional} fi true checks for $value is in default language before checking translations.
+     *
+     * @return Builder|null
+     * 
+     */
+    public function scopeWhereTranslation(Builder $query, $locales, string $field, string $operator, string $value = null, bool $default = false){
```
emptynick: Argument type declarations `string` and `bool` were introduced with PHP 7 but we still support 5.4, ref:
http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration
```suggestion
    public function scopeWhereTranslation(Builder $query, $locales, $field, $operator, $value = null, $default = false){
```

src/Traits/Translatable.php
```diff
@@ -244,6 +245,81 @@ public function setAttributeTranslations($attribute, array $translations, $save
...
+     * @param string|array  $locale     {required} locale(s) you are looking for the field.
+     * @param string        $field      {required} the field your looking to find a value in.
+     * @param string        $operator   {required} value you are looking for or a relation modifier such as LIKE, =, etc.
+     * @param string        $value      {optional} value you are looking for. Only use if you supplied an operator.
+     * @param bool          $default    {optional} if true checks for $value is in default database before checking translations.
+     *
+     * @return Builder|null
+     * 
+     */
+    public static function whereTranslation( $locales, string $field, string $operator, string $value = null, bool $default = false )
```
emptynick: Same as above
```suggestion
    public static function whereTranslation( $locales, $field, $operator, $value = null, $default = false )
```

emptynick (changes requested):

src/Traits/Translatable.php
```diff
@@ -244,6 +245,81 @@ public function setAttributeTranslations($attribute, array $translations, $save
...
+        }
+        if($default){
+            $default = self::where( $field, $value );
+            if( $default->first() ){
+                //Found in default database model
+                return $default;
+            }
+        }
+        $self = new static;
+        $table = ( isset($self->table) ) ? $self->table : studly_case( str_plural ( class_basename( $self ) ) ); //discern table name
```
emptynick: `$self->getTable()` does this for you already
```suggestion
        $table = $self->getTable();
```

emptynick (dismissed):

emptynick:
Sorry for being late. A few more things:
I think the locales-array should not be necessary, so if no locales are given, all are searched.
That would require to put `$locales` somewhere behind `$value`
Next, I think `$default` should be `true` by default.
So, in the end, something like
`whereTranslation($field, $operator, $value = null, $locales = null, $default = true)`
And for looping through the locales, you could simply use [whereIn](https://laravel.com/docs/collections#method-wherein) I think.

reageek:
> 
> 
> Sorry for being late. A few more things:
> I think the locales-array should not be necessary, so if no locales are given, all are searched.
> That would require to put `$locales` somewhere behind `$value`
> Next, I think `$default` should be `true` by default.
> So, in the end, something like
> `whereTranslation($field, $operator, $value = null, $locales = null, $default = true)`
> And for looping through the locales, you could simply use [whereIn](https://laravel.com/docs/collections#method-wherein) I think.

All sound like good ideas. I will look into them. When I have some downtime I will look at updating this.

emptynick:
I commited the stuff I suggested above and added some documentation.
@fletch3555 wanna take a look at the wording?

reageek:
Good changes/refactor. 

Sorry I just really have not had the time (end of the year crunch) to really finish this pr.

One thing that might be noting is in the documentation is that this function does currently override the scope of a query builder due to its static call nature on self. Might be worth putting in. 

To fix this line 304 needs to have a check to see if its statically called in a query builder and use self or the query builder instance.

emptynick:
Laravel does all this for you.
Tested both static and normal calls. Also I simplified the default-clause.
Now it looks good in my eyes. What do you think?

emptynick (dismissed):

emptynick (approved):

fletch3555 (approved):

PR #3112, opened 2018-05-01
Add fix for removing many to many relationship

marijnz0r:
Add a hidden field above the many to many relationship select, to make sure the field is always submitted.
This fixes the problem if you would want to delete all relations to the other object and this wouldn't work.

PR #3424, opened 2018-07-27
Fix to delete last BelongsToMany relation

vvvkor:
Fixes issue: https://github.com/the-control-group/voyager/issues/2476

There was alternative fix (adds hidden dummy input to view): https://github.com/the-control-group/voyager/pull/3112

But the issue should be fixed in controller - which is offered here.


tommykaira (changes requested):
Fixes applied on the view and controller addressed the issue when deleting the last many-to-many value of a record.

vvvkor:
@tommykaira Your review is a "Request changes", but I don't see what changes do you suggest. Am I missing something?

marijnz0r:
Looks like a good solution, thanks for following up on my pull request.

emptynick:
We have to merge one of the new profile PRs before we can merge this one, otherwise those issues will come back: https://github.com/the-control-group/voyager/issues/3010 or https://github.com/the-control-group/voyager/issues/2451
Just so you know why it takes a bit.

emptynick:
Fixed in #3016.
Merged into 1.x

Issue #3432, opened 2018-07-31
"Belongs to Many" relationship throws "column is ambiguous" error—The Return


SturmB:
- Laravel: v5.6.28
 - Voyager: v1.1.3
 - PHP: 7.2.8
 - Database: MySQL 8.0.12

## Description
Looks like the same problem that plagued me before is back. I had reported it months ago at #2847 . I didn't have a chance to verify that it worked at that time, but I'm finally in the position to do so now. Here it is:

Attempting to browse a BREAD of a table for which a "Belongs to Many" relationship has been created results in a "column 'id' in field list is ambiguous" error message. (See screenshot below)

## Steps To Reproduce
1. Create two tables, such as a products table and an invoices table, as well as their respective Models. Both tables should have an id field for their respective Primary Keys.
2. Create a pivot (or joining) table for them. Going by Laravel convention, with the example table names in the previous step, should be invoice_product.
3. Create BREAD for both tables (none is needed for the pivot table, AFAIK).
4. Edit one of the table's BREAD to add a "Belongs to Many" relationship on the other table.
5. Attempt to browse that table's BREAD.

## Expected behavior
I expected to see the usual DataTable of the table's BREAD.

## Screenshots
<img width="1452" alt="screen shot 2018-07-31 at 3 24 02 pm" src="https://user-images.githubusercontent.com/6216224/43483358-722b2e7c-94d9-11e8-8065-b5aa7e35180e.png">

## Additional context
The #2847 issue was fixed before, which is why I find that it's back again with this pivot table so confusing.
Also, when I remove that recently-added "Belongs to Many" relationship, then I can browse the BREAD again just fine—which still has another "Belongs to Many" relationship already defined in it!

SturmB:
As @emptynick pointed out to me in Slack, a minor change was made due to #2847 that fixed the issue before: https://github.com/the-control-group/voyager/blob/1.1/resources/views/formfields/relationship.blade.php#L158

Turns out that a similar fix can be done to line 121 of the exact same file.

Simply exchange `->pluck($options->label)` for `->pluck($options->table.'.'.$options->label)` on that line.

I know that the right thing to do is submit a PR, but I'm still way too much of a n00b with Git/GitHub to figure out how to do that. Thus, I've outlined above what needs to be changed to fix this.

Thanks again, @emptynick!

coke00:
if is for some help you can hidden from the view, for this time. it's not a fix just a help

Issue #2847, opened 2018-03-19
"Belongs to Many" relationship throws "column is ambiguous" error


SturmB:
- Laravel Version: 5.6.11
- Voyager Version: dev-1.x-dev
- PHP Version: 7.2.3
- Database Driver & Version: MySQL 5.7.21

### Description:
Attempting to browse a BREAD of a table for which a "Belongs to Many" relationship has been created results in a "column 'id' in field list is ambiguous" error message:

<img width="1496" alt="screen shot 2018-03-19 at 8 43 47 am" src="https://user-images.githubusercontent.com/6216224/37596989-5e768d02-2b54-11e8-9da5-503a473aeb21.png">

This appears to be a holdover bug from the release version of Voyager, as well, [as noted here](https://github.com/the-control-group/voyager/issues/2383). I don't wish to try the manual fix listed there as that would require editing the **vendor** file, which I'd rather not do. (Plus, it'll just get overwritten the next time I update.)

### Steps To Reproduce:

1. Create two tables, such as a `products` table and an `invoices` table, as well as their respective Models. Both tables should have an `id` field for their respective Primary Keys.
2. Create a pivot (or joining) table for them. Going by Laravel convention, with the example table names in the previous step, should be `invoice_product`.
3. Create BREAD for both tables (none is needed for the pivot table, AFAIK).
4. Edit one of the table's BREAD to add a "Belongs to Many" relationship on the other table.
5. Attempt to browse that table's BREAD.

emptynick:
It should have been fixed in #2579
Can you update and try again?

SturmB:
I will do so a few days from now; unfortunately, the boss has given me a higher-priority task to complete in the meantime. :~/

emptynick:
If it wasn't fixed in #2579 it was in https://github.com/the-control-group/voyager/pull/2436 (Please dont ask haha)

PR #3433, opened 2018-08-01
# Column is No Longer Ambiguous #

SturmB:
* Fixes the `column is ambiguous` error when browsing a BREAD after adding a `Belongs to Many` relationship to it.
  * Changes `->pluck($options->label)` on line 121 to `->pluck($options->table.'.'.$options->label)`.
* See issues #3432 and #2847.


SturmB:
This fix may be already addressed in a better patch at PR #3444. If so, then this one may be disregarded (or deleted or whatever). Further testing is recommended.

emptynick:
Closing in favor of #3444

PR #3444, opened 2018-08-03
# Browsing Accessors Fix #

### LINKED ISSUE #3432: "Belongs to Many" relationship throws "column is ambiguous" error—The Return

### LINKED ISSUE #3457: `Belongs To Many` select box requires `id`
## Version information
 - Laravel: v5.6.29
 - Voyager: v1.1.3
 - PHP: 7.2.8
 - Database: MySQL 8.0.12

## Description
The `select` dropdown when editing a record in a `Belongs To Many` BREAD relationship only shows entries if it is displaying the `id` field in the relationship. Yeah, it's a bit complicated; see the steps and screenshots below for clarification.

## Steps To Reproduce
1. Create a couple of many-to-many related database tables and their associated pivot table, then populate them with some data.
2. Create BREAD for them both, as well as the `Belongs To Many` relationships.
3. Set one of the `Belongs To Many` relationships to `display` a column other than `id`.
4. Save changes, then browse that table's BREAD.
5. Pick a record where the `Belongs To Many` column has some data showing. It should be showing it properly so far. Now click the Edit button for that record.
6. Notice that the data is now missing in the dropdown list for that `Belongs To Many` field.

## Expected behavior
In the Edit BREAD operation, I expect it to show all entries that already exist for a `Belongs To Many` field, regardless of which column was chosen to be displayed for it.

## Screenshots
Hopefully this clears things up.
I have a table called `products` with two `Belongs To Many` relationships. One is set to `display` the `id` field of its related table, `print_methods`; the other is set to `display` a field called `short_name` from its related table, `colors`:
screen shot 2018-08-08 at 9 51 35 am

When browsing the BREAD for `products`, I can see the data that already exists, joining the two related tables to it:
screen shot 2018-08-08 at 9 52 09 am

Working so far.
However, when I click the Edit button for the _first_ entry and look down at the two relationship dropdowns:
screen shot 2018-08-08 at 9 52 26 am
The `print_methods` data shows up because it is `display`ing the `id` field from that table.
The `colors` data is _not_ showing up because it is `display`ing the `short_name` field from its table.

## Additional context
* I have tried setting `product`'s `Belongs To Many` relationship with `colors` to `display` its `id` field, and it _does_ show up. This leads me to believe that it is somehow biased to _only_ show data if it is set to `display` the `id` field.
* I have also tried creating an accessor in `colors` and `display` that accessor instead, but got the same result.
* I have a gut feeling that this is somehow related to issue #3456, but no proof of it.
* I have not tried selecting some items in the dropdown to see what happens, because I do not want to disturb my current database data. (I might try it at some point later, especially if it is requested from someone here trying to help fix this issue.)

SturmB:
* Thanks to @emptynick, this will use accessors on a Model if they exist when browsing/viewing its BREAD (with relationships) instead of assuming that it's always a column name.

Fixes #3432.
Also now fixes #3457.


SturmB:
As a side effect, this fix appears to also address #3432. I had previously created PR #3433 to address that (again, with @emptynick's help), but it may not be necessary anymore. Further testing is recommended.

SturmB:
Bug found already.

With this patch, the `id` field/column does not appear when browsing any BREAD view for a table to which the `Belongs To Many` relationship has been applied. I do not know if it just ignores the first column or the one with the name of `id`. I am not sophisticated enough to debug this.

SturmB:
Okay, the bug I mentioned above is most likely _not_ due to this patch, but rather one that already exists in v1.1. I will open a separate issue about it.

resources/views/formfields/relationship.blade.php
```diff
@@ -155,7 +159,9 @@ class="form-control @if(isset($options->taggable) && $options->taggable == 'on')
                 >
 
                         @php
-                            $selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->pluck($options->table.'.'.$options->key)->all() : array();
+                            $selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->get()->map(function ($item, $key) use ($options) {
+                                return $item->{$options->label};
```
ThisIsJustARandomGuy: This needs to be `return $item->{$options->key}` because it is used in the check below, in line 173:
`@if(in_array($relationshipOption->{$options->key}, $selected_values)){{ 'selected="selected"' }}@endif`

ThisIsJustARandomGuy: The way it's currently implemented, belongsToMany select2 fields always stay empty, because it tries to look for the keys in `$selected_values`, but `$selected_values` stores the labels.

SturmB: Good spot. It will take care of #3457, too. I'll push another commit.

resources/views/formfields/relationship.blade.php
```diff
@@ -160,7 +160,7 @@ class="form-control @if(isset($options->taggable) && $options->taggable == 'on')
 
                         @php
                             $selected_values = isset($dataTypeContent) ? $dataTypeContent->belongsToMany($options->model, $options->pivot_table)->get()->map(function ($item, $key) use ($options) {
-                                return $item->{$options->label};
+                                return $item->{$options->key};
```
SturmB: Fixes #3457.

SturmB, line 163: Fixes #3457.

SturmB (commented):
Fixes #3457.

resources/views/formfields/relationship.blade.php
emptynick: As a side note: those keywords (fixes, resolves, ...) only have an effect when used in the first comment.

SturmB: Thank you. n00b is as n00b does.

fletch3555 (approved):

PR #3450, opened 2018-08-06
Database Test fix

emptynick:
This fixes the `DatabaseTest` failing when using [DBAL >= 2.8.0](https://github.com/doctrine/dbal/releases/tag/v2.8.0).
I honestly dont know what they changed, but changing back to `DBAL 2.7.2` does not throw any errors.

What happens (and why it failed) is:

- The primary key was dropped
- The table-data is sended to Voyagers-UI and saved
- The primary-key is back (This is where the problem starts)
- `assertFalse()` fails

All I can say is that its *not* Voyagers fault as there was no commit at the time the tests started failing.

~The outcome/functionality of the `can_change_index()` test should still be the same.~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "require": {
         "illuminate/support": "~5.4.0|~5.5.0|~5.6.0",
         "intervention/image": "^2.4",
-        "doctrine/dbal": "2.7.*",
+        "doctrine/dbal": "~2.5.0|~2.6.0|~2.7.0",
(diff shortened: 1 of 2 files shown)


fletch3555:
This actually breaks the test.  The `assert` lines are checking for the state AFTER that line runs.  By moving the line after the asserts, you made it so it's only testing that previous state, which is before any change was made.

fletch3555 (approved):
Can't say I love it, but it does work.  Thanks!

PR #3451, opened 2018-08-06
Hidden ID Field Fix

SturmB:
* Comments-out the lines responsible for hiding the `id` field in all BREAD operation views.
  * This is done in response to the fact that any `Belongs To Many` relationship causes the `id` field to of its BREAD to disappear entirely, even when its checkboxes are marked as visible in all operation views.
* Adds TODO notes about creating "config vars" instead of just removing these lines. This was suggested by @emptynick on Slack, but I'm afraid I don't know what those are, nor am I likely to have the skills to implement them.


SturmB:
Never mind. It turns out that this PR is faulty. It causes _other_ fields (foreign key fields, to be specific) to become visible as well when they are meant to be hidden. So I don't know how to make just the `id` field visible again without affecting the other ones.

SturmB:
All credit for finding the cause goes to @MrCrayon. If I understand the underlying cause correctly, the default id field is being saved with a belongsToMany relationship, which makes it appear as though the id field is connected to it, when it shouldn't be. Thus, we stop it from making such a 'connection' by only saving the field if the relationship is a belongsTo relationship. This should fix #3456. Further details can be seen there. This PR supersedes #3451.

* Restores the commented-out lines that were clearly not a part of the problem.
* Adds a check to see if a given relationship is belongsTo before hiding any fields associated with it.

fletch3555 (approved):

Issue #3456, opened 2018-08-08
ID field not shown in BREAD with Belongs To Many relationship


SturmB:
## Version information
 - Laravel: v5.6.29
 - Voyager: v1.1.3
 - PHP: 7.2.8
 - Database: MySQL 8.0.12

## Description
The `id` field does not appear in any BREAD view if that BREAD has a `Belongs To Many` relationship. This is best seen with Voyager's own built-in `users` table & BREAD, since it has a `Belongs To Many` relationship with `roles`.

## Steps To Reproduce
1. Install Laravel, Voyager, database, etc. The usual.
2. Login to Voyager's admin backend and navigate to the BREAD list.
4. **Edit** the BREAD for `users` and tick all of the checkboxes for the `id` field, so that they will be visible in all BREAD operation views. Make sure to save changes, natch.
5. **Browse** the BREAD for `users`.
6. Be astonished that the `id` field is not visible.

## Expected behavior
I expect the `id` field (or _any_ field, for that matter) to be visible whenever it is set as such in the table's BREAD.

## Screenshots
Easy. Here's a shot of the `users` BREAD after I have enabled all checkboxes:
<img width="1251" alt="screen shot 2018-08-06 at 12 41 54 pm" src="https://user-images.githubusercontent.com/6216224/43839924-5a7a5f40-9aed-11e8-8ce2-0f876825ee72.png">

And here's the result of browsing that very same BREAD after saving those changes:
<img width="1248" alt="screen shot 2018-08-06 at 12 42 30 pm" src="https://user-images.githubusercontent.com/6216224/43839957-6da58a4a-9aed-11e8-827d-3c72e2eac755.png">

## Additional context
[...]

PR #3461, opened 2018-08-09
# Hidden ID Field Fix, Attempt Deux #

SturmB:
All credit for finding the cause goes to @MrCrayon. If I understand the underlying cause correctly, the default `id` field is being saved with a `belongsToMany` relationship, which makes it appear as though the `id` field is connected to it, when it shouldn't be. Thus, we stop it from making such a 'connection' by only saving the field if the relationship is a `belongsTo` relationship. Further details can be seen in #3456. This PR supersedes #3451.

* Restores the commented-out lines that were clearly not a part of the problem.
* Adds a check to see if a given relationship is `belongsTo` before hiding any fields associated with it.


SturmB:
Wait, never mind. I did this incorrectly.

PR #3474, opened 2018-08-15
Fixes checkbox validation bug

### LINKED ISSUE #3475: validation bug with required_if and checkboxes
## Version information
 - Laravel: v5.5.*
 - Voyager: v1.1.3
 - PHP: 7.2
 - Database: [type] [version] (e.g. MySQL 8.0)

## Description
In a validation case where e.g. a required_if condition is based on whether a checkbox field is set or not, the validation doesn't work correctly because the checkbox value is always on. This happens because ALL fields are passed with the XHR validation request, even not checked checkboxes.

## Steps To Reproduce
1. Add the following validation to the title field of the posts table:

`{
    "validation": {
        "rule": "required_if:featured,on"
    }
}`

2. Try to save a post with an empty title field. Make sure that the featured checkbox is unchecked.
3. Even though the featured checkbox is off the validation fails.

## Expected behavior
The validation should pass for an empty title field and an unchecked featured checkbox.

## Additional context
See the proposed solution at PR #3474

webber-ch:
Fixes a bug that occurs if validation is associated with checkbox values (e.g. required_if:featured,on).

Fixes #3475


resources/assets/js/app.js
```diff
@@ -173,7 +173,10 @@ $(document).ready(function () {
                     continue;
                 }
             }
-            data.append(this.elements[i].name, this.elements[i].value)
+            // Add checkboxes only if they are checked
+            if(e.currentTarget.elements[i].type != 'checkbox' || e.currentTarget.elements[i].checked) {
```
fletch3555, line 178: I think I prefer `== &&` as opposed to `!= ||` as you have currently.

fletch3555 (dismissed):

resources/assets/js/app.js
webber-ch: I commited the desired changes.

resources/assets/js/app.js
```diff
@@ -173,7 +173,10 @@ $(document).ready(function () {
                     continue;
                 }
             }
-            data.append(this.elements[i].name, this.elements[i].value)
+            // Add checkboxes only if they are checked
+            if(e.currentTarget.elements[i].type != 'checkbox' || e.currentTarget.elements[i].type == 'checkbox' && e.currentTarget.elements[i].checked) {
```
emptynick: Doesn't it need to be
`if(e.currentTarget.elements[i].type != 'checkbox' || (e.currentTarget.elements[i].type == 'checkbox' && e.currentTarget.elements[i].checked))`
?
Or remove the second `e.currentTarget.elements[i].type == 'checkbox' `
So it gets
`if(e.currentTarget.elements[i].type != 'checkbox' || e.currentTarget.elements[i].checked)`

emptynick (dismissed):

resources/assets/js/app.js
fletch3555: `&&` has higher precedence than `||`, so parentheses aren't actually necessary.  Though @emptynick is right that `type == 'checkbox'` is implied in the right side of the `||`.

webber-ch: The "compact" version was applied in my first commit: 74534a3b68a0242c5a535161c29691999d5c6400
The decision is up to you, @fletch3555 !

fletch3555: Apologies for the confusion @tschmidiger.  When I suggested `== &&` before, I meant the existing `!=` could be converted, not that you should add what you added.  Looking at it now, I see that would've been incorrect logic, so I'm glad you didn't do that.  Please switch back to the "compact" version.  I'm good with that.

webber-ch: Ok, no worries...I just switched back to the first commit.

fletch3555:
Bumping JS changes to 1.x

fletch3555 (approved):

Issue #2441, opened 2018-01-06
ID field disappear when create BelongsToMany relationships


doatta:
- Laravel Version: 5.5
- Voyager Version: 1.0.11
- PHP Version:
- Database Driver & Version:

### Description:

Dear developers,
I have 3 tables - 
1. table1s
-id
-name
2. table2s
-id
-name
3. table1_table2
-table1_id
-table2_id

When I create BelongsToMany relationships in table1s, ID field disappear for the table1s.

In Voyager version 1.0.9 ID field also disappeared for edit and view mod, but I was able to see ID field in browse mod, but for the Voyager 1.0,11 it also disappeared from browse mod.

anbulanov:
The same problem for me!

varbark:
I also have same problem, after digging into the code, I found there is a removeRelationshipField function will be called in the controller, which will remove the row field correspond to the column in the relationship 


```
protected function removeRelationshipField(DataType $dataType, $bread_type = 'browse')
    {
        $forget_keys = [];
        foreach ($dataType->{$bread_type.'Rows'} as $key => $row) {
            if ($row->type == 'relationship') {
                $options = json_decode($row->details);
                $relationshipField = @$options->column;
                $keyInCollection = key($dataType->{$bread_type.'Rows'}->where('field', '=', $relationshipField)->toArray());
                array_push($forget_keys, $keyInCollection);
            }
        }

        foreach ($forget_keys as $forget_key) {
            $dataType->{$bread_type.'Rows'}->forget($forget_key);
        }
    }

```

emptynick:
Fixed in https://github.com/the-control-group/voyager/pull/3451

PR #3480, opened 2018-08-22
 FixBug: Remove relationship field only with 'belongsTo' relations

bluemmb:
**Issue:** Relationship field disappears when creating relationships

Related to #2441

It logically seems that among the supported relations : belongsTo, belongsToMany, hasMany, hasOne,
we only need to remove the key field when the relation is of type 'belongsTo'.


MrCrayon:
There is already #3451 that does the same.

PR #3481, opened 2018-08-22
Check if 'file' field is array while cleanup

bluemmb:
When 'file' field is null ( if it is nullable ), the cleanup function raises error in second foreach loop.

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -401,8 +401,7 @@ class VoyagerBaseController extends Controller
 
         // Delete Files
         foreach ($dataType->deleteRows->where('type', 'file') as $row) {
-            $field = json_decode($data->{$row->field});
-            if (is_array($field)) {
+            if (is_array($field = json_decode($data->{$row->field}))) {
                 foreach ($field as $file) {
                     $this->deleteFileIfExists($file->download_link);
                 }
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -401,8 +401,11 @@ protected function cleanup($dataType, $data)
 
         // Delete Files
         foreach ($dataType->deleteRows->where('type', 'file') as $row) {
-            foreach (json_decode($data->{$row->field}) as $file) {
-                $this->deleteFileIfExists($file->download_link);
+            $field = json_decode($data->{$row->field});
+            if (is_array($field)) {
```
fletch3555: You can simplify this a bit with something like:
```
if (is_array($field = json_decode($data->{$row->field}))) {
```

But beyond that, is there any reason passing "true" into `json_decode` as the 2nd argument wouldn't work?

fletch3555 (commented):
Please share some background on this PR.  What prompted it? What's the error? What is it fixing?

src/Http/Controllers/VoyagerBaseController.php
bluemmb: Yes thanks, it is better.

Passing true to json_decode still returns null and raises the same error.

bluemmb:
@fletch3555 I have made a field of type 'file' nullable in my project, and when I try to delete a row that it's file field is null the error below raises.

error

Following the delete operation codes, led me to cleanup() function that tries to delete files but when it is null the `foreach (json_decode($data->{$row->field}) as $file)` error emerges because json_decode returns null.

emptynick:
This should have been fixed in https://github.com/the-control-group/voyager/pull/3580
Please correct me if I'm wrong.

Issue #3497, opened 2018-08-25
voyager:install --with-dummy fails on [zh_CN] locale


dhb52:
## Version information

 - Laravel: v5.6.34
 - Voyager: v1.1.3
 - PHP: 7.2.9
 - Database: Mariadb 10.3.9

## Description

voyager:install --with-dummy fails on [zh_CN] locale.

## Steps To Reproduce
1. set config/app.php ```locale => 'zh_CN'```.
2. ```php artisan voyager:install --with-dummy```

```
Seeding dummy data

   Illuminate\Database\Eloquent\ModelNotFoundException  : No query results for model [TCG\Voyager\Models\DataType].

  at /Users/barry/Code/laravoyage/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:414
    410|         if (! is_null($model = $this->first($columns))) {
    411|             return $model;
    412|         }
    413|
  > 414|         throw (new ModelNotFoundException)->setModel(get_class($this->model));
    415|     }
    416|
    417|     /**
    418|      * Execute the query and get the first result or call a callback.
  Exception trace:

  1   Illuminate\Database\Eloquent\Builder::firstOrFail()
      /Users/barry/Code/laravoyage/database/seeds/TranslationsTableSeeder.php:57

  2   TranslationsTableSeeder::dataTypesTranslations()
      /Users/barry/Code/laravoyage/database/seeds/TranslationsTableSeeder.php:19
```

PR #3498, opened 2018-08-25
voyager:install --with-dummy fails on [zh_CN] locale

dhb52:
voyager:install --with-dummy fails on [zh_CN] locale #3497


emptynick:
This does not really fix the issue, instead it just skips all those items making the seeder useless.
The solution would be to use the translation function with the proper key instead of hardcoded english values.

dhb52:
yah, this is just only a workaround for now to finish the voyager:install command a little more elegant, not a patch.

fletch3555:
I haven't seen any other complaints about this, nor am I overly fond of this "solution".  I'm going to close this for now.  If it's still a problem and there's enough interest, we can revisit it then

Issue #3388, opened 2018-07-11
Cannot add multiselect option


anandmainali:
## Version information

 - Laravel: v5.6.*
 - Voyager: v1.1.*
 - PHP: 7.1.3
 - Database: mysql 
## Description

I am trying to add product size option using multiselect. But i am getting a issue of,

Invalid argument supplied for foreach() (View: /home/helpme/testing/vendor/tcg/voyager/resources/views/bread/browse.blade.php)

## Screenshots
![error](https://user-images.githubusercontent.com/34227482/42550131-811feb74-84ef-11e8-9047-e3b26cd96c32.png)
![select](https://user-images.githubusercontent.com/34227482/42550134-88f16288-84ef-11e8-86c4-b5bbb649c588.png)

fletch3555:
Your `options` should be an array (`[ ]`), not an object (`{ }`)

gRoberts84:
@fletch3555 - This is still an issue.

I have setup an `Select Multiple` field and set the additional details to contain:

```
{"options":{"online":"Online","sms":"SMS","phone":"Phone","postal":"Postal"}}
```

When going to the Add form, the options are available in the dropdown and upon selection, are added to the underlying table as:

```
["online"]
```

When browsing the table, it's expecting an array, but it's returning the raw value. 

The only way I can get around this is to use a mutator, i.e. 

```
    public function getEntryMethodsAttribute() {
        return json_decode($this->attributes['entry_methods'], true);
    }
```

I have also noticed that the default values do not appear to work either (although I cannot find any documentation to support it beyond an empty string)

HTH?

gRoberts84:
Well.... that's annoying.

Seems that issue only affects the browse view. The Edit view expects it to be a string and complains that it is already an array.

tommykaira:
This is an issue for me as well.

![voyager-select-multiple](https://user-images.githubusercontent.com/1387152/44628010-adfe5d00-a97b-11e8-81ea-b2a51fe5c289.png)

![voyager-select-multiple2](https://user-images.githubusercontent.com/1387152/44628060-83f96a80-a97c-11e8-821f-0486fda6f7e0.png)

The value stored in the table is a serialized JSON string.

What I did was I had to cast it to an array type in the model

e.g.

`protected $casts = [
        'options' => 'array',
    ];`

PR #3500, opened 2018-08-26
Select multiple throwing Invalid argument

emptynick:
The browse-view didn't `json_decode` the data from the DB, leading to a
`Invalid argument supplied for foreach() (View: vendor/tcg/voyager/resources/views/bread/browse.blade.php)`

Because it was trying to execute (for example)
```
foreach ('["1","3"]' as $item) {  }
```
This PR also fixes the read-view when the value is `null`
```
foreach (json_decode(NULL) as $item) {  }
```

Fixes https://github.com/the-control-group/voyager/issues/3388


fletch3555 (approved):

fletch3555:
@emptynick we should consider pulling the json_decode logic into the controller (or model..?) to avoid having to call it 6 times in the view.  That can be done separately though

PR #3504, opened 2018-08-27
Allow setting field descriptions via a translation file

ThisIsJustARandomGuy:
In order to simplify updating field descriptions I added a few lines of code to the DescriptionHandler class. Default behavior is not changed, so if a DataRow has $details->description set, that description will be displayed. However, if a translation `voyager.help.{$dataType->name}.{$row->field}` exists, that will be used.

So suppose we have a DataType `posts` which has one DataRow `title`. The DescriptionHandler as it is now will just check whether `DataRow->options->description` exists. If it doesn't exist, the description isn't visible. The updated version does an additional check (only if options->description isn't set) - it looks for the translation voyager.help.posts.title and uses that instead, if it exists.

### DIFF
--- a/src/FormFields/After/DescriptionHandler.php
+++ b/src/FormFields/After/DescriptionHandler.php
@@ -8,6 +8,11 @@ class DescriptionHandler extends AbstractHandler
 
     public function visible($row, $dataType, $dataTypeContent, $options)
     {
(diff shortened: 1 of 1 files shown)


ThisIsJustARandomGuy:
This would also need some description (get it?) in the Readme.

emptynick:
This will be available in Voyager 2 ootb.

PR #3507, opened 2018-08-28
Proposed fix for multiple image field quality setting

gwleuverink:
The multiple image content type class ignores user defined quality settings.

The class does an isset() check on the $options variable, which is not defined thus always returns false. As a result the $resize_quality variable is set to it's default of 75%.

This PR fixes the conditional check.


fletch3555:
This is not a breaking change and therefore could go into `1.1`.  Could you please redo this PR starting from the `1.1` branch instead?  We recommend always using `1.1` as your working branch unless you NEED to rely on some functionality that has already been merged into `1.x`

gwleuverink:
@fletch3555 Not a problem. Please check out this new PR over here #3508

PR #3511, opened 2018-08-29
fix hard coded menu item policy (#3501)

### LINKED ISSUE #3501: MenuItemPolicy ignores passed action
## Version information
 - Laravel: v5.6.28
 - Voyager: v1.1.3
 - PHP: 7.2

## Description
Right now regardless of passed action, the MenuItemPolicy checks only for "browse" action.
```PHP
    protected function checkPermission(User $user, $model, $action)
    {
        ...
        // If permission doesn't exist, we can't check it!
        if (!self::$permissions->contains('key', 'browse_'.$slug)) {
            return true;
        }

        return $user->hasPermission('browse_'.$slug);
    }
```
The action should not be hard coded in the policy check. That is counterintuitive and breaks the flexibility.

## Expected behavior
```PHP
    protected function checkPermission(User $user, $model, $action)
    {
        ...
        // If permission doesn't exist, we can't check it!
        if (!self::$permissions->contains('key', $action.'_'.$slug)) {
            return true;
        }

        return $user->hasPermission($action.'_'.$slug);
    }
```

sbruev:
Fixes #3501

### DIFF
--- a/src/Policies/MenuItemPolicy.php
+++ b/src/Policies/MenuItemPolicy.php
@@ -42,6 +42,10 @@ class MenuItemPolicy extends BasePolicy
             $slug = 'admin';
         }
 
+        if (empty($action)) {
+            $action = 'browse';
+        }
+
         // If permission doesn't exist, we can't check it!
         if (!self::$permissions->contains('key', $action.'_'.$slug)) {
             return true;
--- a/src/Policies/MenuItemPolicy.php
+++ b/src/Policies/MenuItemPolicy.php
@@ -43,10 +43,10 @@ class MenuItemPolicy extends BasePolicy
         }
 
         // If permission doesn't exist, we can't check it!
-        if (!self::$permissions->contains('key', 'browse_'.$slug)) {
+        if (!self::$permissions->contains('key', $action.'_'.$slug)) {
             return true;
         }
 
-        return $user->hasPermission('browse_'.$slug);
+        return $user->hasPermission($action.'_'.$slug);
     }
 }


emptynick:
We need to make sure that we pass `browse` as the action to this method and not an empty string for example.
Otherwise we could let it default to `browse`.

sbruev:
@emptynick While making the empty check or default action to 'browse' is trivial, I don't think we should encourage the bad practices. We could make the 'browse' default action for all polices by editing the Base Policy, of course, but why? Someone has to pass intentionally the null or '' as first argument. That is very strange decision and not correct (just imagine the nightmare to review that code by someone else).

On the other hand, if we make that check only for Menu Item Policy, we will violate the uniformity of working with policies.

emptynick:
I'm just saying that we need to make sure that we **don't** pass an empty string or null

src/Policies/MenuItemPolicy.php
```diff
@@ -43,10 +43,10 @@ protected function checkPermission(User $user, $model, $action)
         }
 
         // If permission doesn't exist, we can't check it!
-        if (!self::$permissions->contains('key', 'browse_'.$slug)) {
+        if (!self::$permissions->contains('key', $action.'_'.$slug)) {
```
fletch3555: As mentioned by @emptynick, we need to guarantee `$action` is never null or empty string.  Can you add a check above this similar to:
```
if ($action == null || empty($action)) {
    $action = 'browse';
}
```

fletch3555 (changes requested):

src/Policies/MenuItemPolicy.php
sbruev: - First of all `empty ($action)` function already includes check for `null` value.
- Secondly, with all due respect, I disagree with the necessity of this check. We don't do the similar checks in other policies, why we should do it here?

emptynick: Because `$action` was always `browse` (because it was hardcoded)
So someone could use `checkPermission($user, $model, '')` and it would work.
If we don't check it, it wont work anymore so its kind of a breaking change.
We always have to keep the worst-case-scenario in mind.

fletch3555: Exactly as @emptynick said.  Just because others don't do it doesn't mean this shouldn't.  If we change other policies to this dynamic approach, then they will need a default set as well.  Thanks for making the quick change.

fletch3555 (approved):

PR #3514, opened 2018-09-01
Feature add bread link to navmenu

damms005:
BREAD is one of the integral parts of Voyager. This PR puts the BREAD link into the navigation menu


fletch3555:
This MenuItem already exists due to #3085

PR #3543, opened 2018-09-14
1.1 link field

Emerica:
I don't expect this to pass :)
Until it's known how different read and browse views for bread fields can be customised going forward, this is a proof of concept.

 A field could be added within the voyager config for the field types instead of inside the main service provider.
The browse and read views can loop over those field values and @include each fields template, browse, read, edit, etc.
I figure the bread.formfields could hold the views, I'm not sure how moving the current edit views would effect other code? 

This link field allows for icons from googel s2,  font-awesome, or a static image
{
"type": "fa",
"icon": "facebook",
"hideurl":true,
"target":"_blank"
}

{
"type": "image",
"icon": "http://someimage.jpg",
"hideurl":true,
"target":"_blank"
"download":true
}
type - google, image, fa
icon: google: ignored, image: url, fa: font awesome icon, eg facebook.
download  - <a> download
hreflang -  <a>  language , en
media - <a>  media
ping - <a> ping
rel - <a> rel
atype - <a> type
target - <a> target
aclass - <a> class
imgclass - class
hideurl - will hide the url text


Emerica:
As per the mention above about doing this more dynamically in the bread view blades.

Move the edit fields into the bread/edit folder, then you can create custom views for custom form fields on each type of bread view.

```blade
@foreach($dataType->browseRows as $row)
    <td>
           <?php $options = json_decode($row->details); ?>
           @foreach($voyager_config_fields as $field) #Configure these in config/app or voyager.php?
               @if($row->type == $field)
                   @include('voyager::bread.formfields.browse.'.$field)
               @endif
           @endforeach
     </td>
@endforeach
```

emptynick:
I think this is better made as a hook.
Closing

PR #3546, opened 2018-09-15
FormData polyfill for IE / Edge

hooch:
This restores the ability to save changes in IE11 and Edge.


hooch:
If you refer to Mozilla docs on FormData, cross-browser support is pretty shaky:

https://developer.mozilla.org/en-US/docs/Web/API/FormData#Browser_compatibility

tnylea:
Thanks @hooch 

So, currently the form submission is not working in IE 11 and edge is that why the Polyfill is needed?

Just want to clarify.

Thanks.

fletch3555:
Moving to `1.x` as it requires rebuilding and republishing JS.

fletch3555 (approved):

PR #3547, opened 2018-09-15
#3350 - restore the ability to empty a multi select

hooch:
This restores the ability to empty a multi select, and works whether or not the user selects the "None" option.  It also works when the user completely empties the field of all options.

The change to Relationship.php is necessary to rid the $content array of ```[ n => null ]``` values, which trigger an SQL error.


hooch:
Fix for the styleci issue a335a42

emptynick (approved):
Working fine. Thanks!

PR #3553, opened 2018-09-16
Replace OR with ?? for Laravel 5.7 compatibility 

cotiga:
There are certainly more...


fletch3555:
What version of php was that operator added in? 7.0? 7.1? We still technically support 5.6, so you'll definitely find a few older language constructs in use

cotiga:
Ouch, I think you must think to no longer support the older version of Laravel (and therefore PHP)

Since PHP 7.0
http://php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op

https://laravel.com/docs/5.7/upgrade
Blade
The or Operator

emptynick:
L5.7 is just a few days old. we can't drop support for all other versions.
Instead we should manually test those variables with `isset` or something.

cotiga:
Use #3562 (backward compatible)

PR #3569, opened 2018-09-19
Fix migration rollback issue

shenxn:
This issue is already fixed in branch 1.x by PR [#3073](https://github.com/the-control-group/voyager/pull/3073). It still exists in branch 1.1


fletch3555:
That was intentional as we consider it a breaking change (you would need to publish the updated migration after updating Voyager).

PR #3580, opened 2018-09-25
Fix BREAD file type bug while delete

### LINKED ISSUE #1855: Invalid argument supplied for foreach()
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7.1
- Database Driver & Version: 5.7.19

### Description:

Return invalid argument for a foreach, when trying to delete a record with a file field, but no file is present.

### Steps To Reproduce:

Create a record with a file field, but don't upload any file.
Delete that record from the admin interface.

You will get an invalid argument on the foreach.
screen shot 2017-09-24 at 21 59 36

It seems that it goes to the foreach without verifying if a file is really present first?

skyArony:
fix #1855 

The BREAD file type exist a bug that if it be set NULL，when delete it，it will throw
```
ErrorException (E_WARNING)
Invalid argument supplied for foreach()
```
because there is not check the variable

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -401,8 +401,10 @@ class VoyagerBaseController extends Controller
 
         // Delete Files
         foreach ($dataType->deleteRows->where('type', 'file') as $row) {
-            foreach (json_decode($data->{$row->field}) as $file) {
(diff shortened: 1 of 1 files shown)


tnylea (approved):
Thanks man!

This looks good. Thanks for adding this bug fix.

PR #3584, opened 2018-09-25
Adding option for the voyager migrations to be autoloaded

tnylea:
In some cases users may regenerate their migrations and have them all located in the `/database/migrations` folder. If this happens, it will conflict with the migrations in the `tcg/voyager` folder.

This new addition to the config will allow the user to prevent autoloading the migrations in the `tcg/voyager` folder.


fletch3555:
That's actually not true.  Laravel enumerates migrations by filename, so its impossible to have duplicates conflict.

tnylea:
Here's what I've done in the past.

I would regenerate my migrations based on the current mysql database. This means that my DataTypes and DataRows table gets a new migration created in the `/database/migrations` folder.

This would mean that I now have 2 migrations for DataTypes and DataRows table. Adding this config will give the user the option to exclude the migrations from the `tcg/voyager` folder when running:

```
php artisan migrate
```

fletch3555:
That seems like an odd use-case to me.  I personally would put the data in seeders and just overwrite those.  Leave the migrations for structure, which likely won't change from what Voyager provides (combined with any custom migrations you may add, but that shouldn't affect the data_rows/data_types tables).

tnylea:
I understand it may be a smaller use case than others, but I feel that more options couldn't hurt anything. This would be an additional line in the config. I could see many times where a user might need this option.

The `loadMigrations` method is optional via the Package developer, but in this case we could add a single line to the config to allow users to exclude them.

This is not going to change the default functionality, it will only allow for an additional configuration for the developer of the application.

Due to projects that I've migrated over from other frameworks I've needed to do a reverse migration and seed and have ran into this same issue a few times.

tnylea:
The migrations actually used to be added to the migrations folder but was changed via this commit: https://github.com/the-control-group/voyager/commit/e5a1464a6245925ad9c4551f77d5fa7bcc771c40#diff-1de6db76b51f87f079c5c336de8d8ecc which is totally cool, but now this will allow for more flexibility. Let me know if you don't think it should go in and we can discuss further.

Thanks :)

tnylea:
@fletch3555 Let me know if you can merge this in today :)

Thanks man!

fletch3555 (approved):

emptynick:
Please keep in mind that we need that config-var in [config/voyager-dummy.php](https://github.com/the-control-group/voyager/blob/1.1/publishable/config/voyager_dummy.php) also.

fletch3555:
Yes, fair point.  Mind submitting a PR to get that added as well?

PR #3004, opened 2018-04-19
Safari not posting forms with empty files. Issue #3001

jannejava:
There seems to be an bug in Safari 11 that is not sending forms with empty files.

https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari

Fixes #3001

PR #3589, opened 2018-09-28
Update app.js

dimetriynik:
Fixed error sending form data.

Previously, not the correct data was sent, if a radio button with the same name is used, several values are sent

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -162,19 +162,8 @@ $(document).ready(function () {
 
         var url = $(this).attr('action');
         var form = $(this);
-        var data = new FormData();
-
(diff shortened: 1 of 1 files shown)


emptynick:
This basically just reverts https://github.com/the-control-group/voyager/pull/3004.
I think we need to look for another solution for the problem you describe.
Also, we need to merge that into 1.x

fletch3555:
Agreed with @emptynick.  I'm not opposed to reverting that fix so long as another fix is provided.  Or someone finds another solution to this problem.

Either way, I'm going to close this.  Feel free to reopen or open a new PR if a solution is discovered.

PR #3601, opened 2018-10-04
Throw Exception if request does not match with translatable attributes

DrudgeRajen:
```
foreach ($transFields as $field) {
        if (!$request->input($field . '_i18n')) {
            throw new Exception("Invalid Translatable field " . $field);
        }

        ...
}
```


emptynick:
What does this fix or when does it happen?

DrudgeRajen:
@emptynick If the field name mentioned in a model for translatable doesn't get a match with the actual name of table column then it throws an error. So, It will throw the exception so that devs can know what's actually going on inside.

fletch3555 (approved):

PR #3602, opened 2018-10-04
Added fit() option (without position)

### LINKED ISSUE #3590: Add fit features for images
We've got an option to crop, resize and upsize an image. However, not always it's possible to predict in what orientation the user will upload an image so that rules `resize: 300x300` and `crop: 300x300` will result in either square image or rectangle because an image is landscape-oriented.
Therefore there is a need in `fit` like in Intervention/Image so that it will be needed size always. Please, consider to add this option to image fields.

gultyayev:
When height isn't set it will be equal to $resize_width as it works in Intervention Image itself.

Fixes #3590


yarkm13:
I have checked PR, working correct.

Issue #2067, opened 2017-10-22
Is there a way to get the current login user name in JSON. 


masumonline:
- Laravel Version: 5.5.18
- Voyager Version: 1.0
- PHP Version: 7
- Database Driver & Version:

### Description:
Is there a way to to add current loin user name on text field. so when a user fill a form there user name can add on the filed automatic. 

I know on laravel i can use Auth::user()->name to get current user but how can i use it in brade options.

This is a Question.

rppudo:
{{Auth::user()->name}}

fletch3555:
I'm not sure what this has to do with Voyager.  Are you looking to have certain fields pre-fill with the user's name? If so, That's not something Voyager currently supports, and it's very unlikely we'll merge that into Voyager, but you're free to make it work for your own project.

Also, "how do I..." questions are best asked on our slack group.  It's a lot easier to hold discussions there.

PR #3616, opened 2018-10-10
1.x

HazemHa:
Hi guys 
maybe  you solve this issue....
https://github.com/the-control-group/voyager/issues/2067

i solved it ,and i want to share it..
if there any one  still  looking  for solution..
steps are in these images
https://imgur.com/a/1ax5Est
in brief..
step 1:
add checkbox in the form of creating Relationship 
step 2:
receive this field which added in step 1 and add the field  in $relationshipDetails 
which it json format.
step 3:
replace the select menu in link {tableName}/createwith with  two hidden inputs.
**first input**: to select the column name which we will use Auth:: user () ->id 
**second input**: to select it is allowed for us to use authentication user or not 
step 4:
receive fields in the controller and check what column need to fill with authentication user 
if true, add the user id in the request
with the same name column, which we select when we create relationships
code : 
-1-
```
<p>fillable by Auth user  : <input type="checkbox" name="authUser" value="true"></p>
in relationship-new-modal.blade
```

-2-
in VoyagerBreadController  function addRelationship
```
// Build the relationship details
$relationshipDetails = added this row in array ('authUser'    => !empty($request->authUser))

```
-3-
replace the select menu with two hidden inputs if we want to use current auth user
```
   @if(!isset($options->authUser))  .
..etc
.etc
..etc
..etc
@else
 <input hidden type="text" name="requiredColumn" value="{{ $options->column }}">
<input hidden type="text" name="{{ $options->column }}" value="{{$options->authUser}}">
 @endi
```f

step 4

in the VoyagerBaseController in store function store
```
 public function store(Request $request)
    {
        $requiredColumn = $request->requiredColumn;
        $CheckIsAllow = (int)$request[$requiredColumn];
        if($CheckIsAllow == 1){
        array_merge($request->all(), [$requiredColumn => \Auth::user()->id]);
         }
}
```f


fletch3555:
This should not be a pull request.

HazemHa:
ok sorry about that

PR #3621, opened 2018-10-11
added listener for bread delete for menu and permission removal

ntuple:
**Delete** the **Menu** and **Permissions** associated with a **Bread** if the Bread is deleted.


src/Listeners/DeleteBreadMenuItem.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Delete a MenuItem for a given BREAD.
+     *
+     * @param BreadDeleted $bread
+     *
+     * @return void
+     */
+    public function handle(BreadDeleted $bread)
+    {
+        if (config('voyager.bread.add_menu_item') && file_exists(base_path('routes/web.php'))) {
```
fletch3555: What's the purpose of checking for the route file?

src/Listeners/DeleteBreadPermission.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Delete Permission for a given BREAD.
+     *
+     * @param BreadDeleted $bread
+     *
+     * @return void
+     */
+    public function handle(BreadDeleted $bread)
+    {
+        if (config('voyager.bread.add_permission') && file_exists(base_path('routes/web.php'))) {
```
fletch3555: Again with the route file

src/Listeners/DeleteBreadPermission.php
```diff
@@ -0,0 +1,42 @@
...
+            // Permission::generateFor(snake_case($bread->dataType->slug));
+            $role = Role::where('name', config('voyager.bread.default_role'))->firstOrFail();
```
fletch3555: None of this is necessary.  I'd say it's safe to assume the database is set up correctly with ONDELETE, and will delete the associated pivot table rows when the permissions themselves are deleted.

src/Listeners/DeleteBreadPermission.php
```diff
@@ -0,0 +1,42 @@
...
+
+            // Get permission for added table
+            $permissions = Permission::where(['table_name' => $bread->dataType->name])->get()->pluck('id')->all();
+
+            // Assign permission to admin
+            $role->permissions()->detach($permissions);
```
fletch3555: You're not actually deleting the permissions anywhere, just removing them from the default user.  Instead, they should be deleted and removed from all roles (ideally automatically per my comment above, or manually if soft-delete is used)

fletch3555 (changes requested):

src/Listeners/DeleteBreadMenuItem.php
```diff
@@ -0,0 +1,37 @@
...
+     * Delete a MenuItem for a given BREAD.
+     *
+     * @param BreadDeleted $bread
+     *
+     * @return void
+     */
+    public function handle(BreadDeleted $bread)
+    {
+        if (config('voyager.bread.add_menu_item')) {
+            $menuItem = MenuItem::where(['title' => $bread->dataType->display_name_plural]);
```
emptynick: I think its more unique/safe to select the menu-item by the route.
```suggestion
            $menuItem = MenuItem::where('route', 'voyager.'.$bread->dataType->slug.'.index');
```

emptynick (dismissed):

fletch3555 (approved):

PR #3624, opened 2018-10-12
Fix issue #2996 and multi-bytes language problem

### LINKED ISSUE #2996: The BelongsTo SelectBox is bugged on Edit - Doesn't Update the value
- Laravel Version: 5.5
- Voyager Version: Last
- PHP Version: 7.2
- Database Driver & Version:

### Description:
The BelongsTo SelectBox is bugged on Edit - Doesn't Update the value

### Steps To Reproduce:
Add a record to a table with a relantionship. After, edit it and select other value on the relationship. The record won't update the value after saving. I debbuged the Select and looks like the "Selected" value never changes.

ghost:
1. Fixes #2996
2. Fixes content may not display while it is multi-bytes language.


fletch3555:
Please don't solve multiple unrelated issues in a single PR.  Now I'm forced to either accept both or decline both.

Also, I have no idea what the fix is doing.  Why are you removing the `role_id` datarow entry? It's not "useless" (as your commit message claims).

emptynick:
The multi-bytes stuff was already fixed in https://github.com/the-control-group/voyager/pull/3667

emptynick:
And #2996 was fixed in https://github.com/the-control-group/voyager/pull/3016.
So I think both concerns of this PR were fixed in other PRs and ways.
Closing.

PR #3630, opened 2018-10-15
Update app.js

yaroslawww:
Fix display error for name with array.


fletch3555:
I'm not sure what you're fixing here.  The comment directly above the line you changed says this should already work without your change.

```
// This will work also for fields with brackets in the name, ie. name="image[]
```

fletch3555:
Closing due to inactivity

PR #3645, opened 2018-10-18
Use actual PK instead of assuming id.

gmedaglia:
Multiple images removal is not working when using a primary key != "id" for the corresponding table. The `data-id` attribute for the images doesn't get populated and the ajax request returns a Bad Request error.


fletch3555:
This is something we fixed a while back, but it looks like that one snuck by (or was merged in accidentally).

Good catch, thanks!

fletch3555 (approved):

PR #3660, opened 2018-10-20
fix relationship  browse.blade.php

KarimQaderi:
fix relationship

Fixes https://github.com/the-control-group/voyager/issues/3659

PR #3661, opened 2018-10-20
fix relationship edit-add.blade.php

KarimQaderi:
fix relationship


emptynick (changes requested):
I think I'd prefer to change line 70 to something like
`@include('voyager::formfields.relationship', ['options' => $row->details])`
Instead of assigning the variable for every formfield.

emptynick (approved):

fletch3555:
This and #3660 probably could've been combined into one, but that's okay.  Thanks!

PR #3674, opened 2018-10-22
Japanese language (ja) files

shibomb:
Add: Japanese language (ja) files
Update:  Can not translate 'Browse' string in admin BREAD index view.


fletch3555:
Please don't make multiple distinct changes in a single PR.  I would be forced to either accept or decline both.

fletch3555 (approved):

PR #3691, opened 2018-10-24
Relation use html

KarimQaderi:
Model user
```
public function getNameAttribute($value)
    {
        if(strpos(Route::currentRouteName(), 'voyager.') !== false && strpos(Route::currentRouteName(), '.index')!== false)
            return "<a href='/user/$this->id'>$value</a>";
        else
            return $value;

    }
```

image


fletch3555:
This would be much better if done in the view instead of the model.  I'm thinking "link field" could be a new FormField, or maybe just use that for any relationship in the browse/read view.  Lots of options to think about, but I can't accept this as-is

PR #3695, opened 2018-10-25
relation link

KarimQaderi:


emptynick:
I'm not really sure how I feel about this.
You are searching for the DataType in every single row.
2 relationships and 25 results per page results in 50 additional queries, of which 49 are not necessary.
I think this should be implemented in a controller.

fletch3555:
Agreed.  That code should live elsewhere

emptynick:
Closing as the fork was deleted.

PR #3704, opened 2018-10-27
Fix upload large files with Dropzone

belguinan:
When you try to upload larger files, you need to set timeout to more than 30 seconds.


emptynick:
I think 180.000 seconds is way too much and not really reasonable.
That's more than 2 days. Lets go with a few minutes or even an hour.

fletch3555:
@emptynick a [quick search](https://stackoverflow.com/questions/44512084/large-file-upload-request-gets-cancelled) shows that timeout is in milliseconds, not seconds, so 180k is only 3 minutes

emptynick:
Whoops, then I didnt say anything.

fletch3555 (approved):

PR #3710, opened 2018-10-30
First parameter must either be an object or the name of an existing c…

m-hcb:
…lass

First parameter must either be an object or the name of an existing class issue arises when in create timestamp field in any entity. Id we do not specify any details options in bread, it stores in db null, so at the time of browse, it gives error.
So this is my workaround.


emptynick:
Its does not store null anymore. See
https://github.com/the-control-group/voyager/pull/3662

emptynick:
Closing as this should be fixed in https://github.com/the-control-group/voyager/pull/3662

Issue #3712, opened 2018-10-30
 After composer update I got error


dyaskur:
## Version information

 - Laravel: v5.6
 - Voyager: v1.1.10
 - PHP: 7.1.5
 - Database: mysql v 8.0.12

## Description

After I upgraded voyager from v1.1.7  to v1.1.10 I got this error

> json_decode() expects parameter 1 to be string, object given

in code:
`            $details                              = json_decode($row->details);`

## Steps To Reproduce
It's happen in all BREADs page.

## Screenshots
![screen shot 1440-02-21 at 4 02 54 pm](https://user-images.githubusercontent.com/9539970/47710998-be2a0580-dc66-11e8-9cb1-ce1e03693c88.png)

DrudgeRajen:
Might be because of **cached view** try ```php artisan view:clear``` and reload.

emptynick:
He is using custom views.
We changed Voyager to directly return an array instead of `json_decode` it everywhere.
There are quite a few issues with the same "problem"

DrudgeRajen:
@emptynick 

> We changed Voyager to directly return an array instead of json_decode it everywhere.

the problem here is the view is custom which is the copy of voyager bread.blade.php. We have update voyager to directly return array instead of json_decode everywhere. So, the value here it got the object inside ```json_decode()``` . I think the custom views is not used properly.

dyaskur:
@emptynick
 > We changed Voyager to directly return an array instead of json_decode it everywhere. 

I see, now fixed. Thanks you

@DrudgeRajen 
> I think the custom views is not used properly.

I just copied from vendor, am I wrong?

emptynick:
If you didn't changed anything in your views until now, just copy them again.

dyaskur:
@emptynick 
> If you didn't changed anything in your views until now, just copy them again.

I changed something but I already copied again with my changes. So the problem is solved. But I have question, the way I use Custom Views is not wrong, is it? I just copied from tcg/vendor to  local resource folder and then edit it.

emptynick:
Thats one way to do it, so its not wrong.
But we can't guarantee that it wont break like it just did in future versions.

dyaskur:
Okay thank you

DrudgeRajen:
@emptynick It will obviously break if anyone has used view by copying it from original one since the way of fetching data is changed in newer versions.  So, if original view is copied then it should be replaced with new one.

jniyaz:
[...]

PR #3713, opened 2018-10-30
added @yield to add more submit buttons

DrudgeRajen:
As per the documentation, to overrides views

> You can override any of the BREAD views by creating a new folder in resources/views/vendor/voyager/slug-name and slug-name will be the slug that you have assigned for that table.

So, to make a small change we need to copy the whole view to the app level ```resource/vendor/voyager/slug-name``` dir. And if anything gets the change in future in the view part then It will break the app level views. Like #3712.
 As a result, devs need to copy and paste the whole thing again by maintaining their changes. 
So, instead of copying and pasting the whole view code, the developer can extend the bread views in the ```resource/views/vendor/voyager/slug-name``` and can use the benefit of ```@yield``` in extended views so that developer can add the required field and other things in the extended one. 
Example: 
Added veiw in ```resources/views/vendor/voyager/slug-name/edit-add.blade.php```
```
@extends('voyager::bread.edit-add')
@section('submit-buttons')
    @parent
    <button type="submit" class="btn btn-primary save">Save And Publish</button>
@endsection
```
And the action of the button can be handled by overriding the respective controller.


fletch3555:
I like the idea, but I don't necessarily see the benefit of it here.  How often are you adding other buttons here? What do they do?

DrudgeRajen:
@fletch3555 The benefit is to give the flexibility to add more then one button if required without  pasting the whole view content. Example: Adding save and publish button which will save and publish the content which can be implemented by overriding the respective method.

fletch3555 (approved):
I don't see significant benefit from this particular change, but I do like making the views more modular so extending them is easier.

Issue #3706, opened 2018-10-29
Multilanguage and duplicate data


Crazyweedz:
## Version information

 - Laravel: v5.7
 - Voyager: v1.1
 - PHP: 7.1
 - Database: [MySQL] [5.7.19] 

## Description

I have enabled multilanguage on my model and everythings seems work fine but when I read my model I have a duplicate data.

I use the default views and controllers, how to fix this problem?

-- I apologize for my bad english

## Steps To Reproduce
1. Go to 'Tools & Database'
2. Create a new table
3. Go to 'Tools a BREAD'
4. Click on 'Add BREAD to this table' on the table formerly created
5. Set an input to 'Rich Text Box' & Save
6. Enabled the multilinguage on this model a create a new entry
7. Read your newest entry.

## Screenshots
Green Highlight => Original data
Red Highlight => Duplicate data

![chrome_2018-10-29_17-43-49](https://user-images.githubusercontent.com/9811137/47666785-ba35af00-dba4-11e8-875e-f37e7ceaa8ed.jpg)
![chrome_2018-10-29_17-43-56](https://user-images.githubusercontent.com/9811137/47666786-ba35af00-dba4-11e8-8a2c-308f18c57d58.jpg)

yakuter:
Hello Crazyweedz,
In your database, there must be a **translations** table. Can you check it about the translation of this post. If there is no duplicate content, then it means the problem is jus about your codebase.

Crazyweedz:
Hello, thank you for your help.

Maybe my codebase is wrong, but the problem happen on fresh install of Voyager and Laravel.

When i go on the Route 'voyager.posts.show', problem still here and the duplicate data too.

I check translations table, there is no duplicate content.

I do not see what I could have done wrong

yakuter:
Ok, this is a progress. Can you show your 'table test' model file content? I guess it is the only file you edited over fresh install.

Crazyweedz:
Yes, sure. Here is it 

```
<?php

namespace App;

use Illuminate\Database\Eloquent\Model;
use TCG\Voyager\Traits\Translatable;


class TableTest extends Model
{
    use Translatable;

    protected $translatable = ['nom'];
}
```

Yes, this is only file I edited. I want to reiterate that the problem only appears for Rich Text Box input type

In fact, I just followed the indications of the documentation

yakuter:
Crazyweedz, you did everything fine, nothing is wrong with your work. 

[...]

PR #3714, opened 2018-10-30
Turkish Language File and Multilingual Issue

yakuter:
I found two mistakes in my recent merged pr about Turkish language. They were in datatable.php and dimmer.php

Just updated the files.


yakuter:
I found an issue (#3706)  about duplicate content in multilingual read blade and added the fix to PR.

fletch3555:
Please don't mix concerns in a single PR.  Right now, if I was fine with the language changes but not the div change, I wouldn't be able to just accept one.  Same logic applies if one change was backwards-compatible, but the other wasn't, I'd be forced to push them both back to the next minor release instead of a patch.

fletch3555:
Closing due to inactivity.  Please submit these two changes as separate PRs if they're still relevant, along with justification for the change.

PR #3739, opened 2018-11-09
Edit languages (Main modify zh-TW)

ycs77:
* Main modify language: `zh-TW`
* Replace the underscore in the language folder name with a hyphen. Because using [Laravel-lang](https://github.com/caouecs/Laravel-lang). (Example: `zh_TW` => `zh-TW`)
* `analytics.set_view` => `analytics.select_view`


fletch3555:
We can't just change the locale strings just to appease a 3rd-party package we don't use.  We need to make sure it still works for everyone that's NOT using that package.

Having said that, I understand the `-` version is the "standard" (https://en.wikipedia.org/wiki/IETF_language_tag), but the php-intl extension (which I believe is what Laravel uses under the hood) returns the `_` version (http://php.net/manual/en/resourcebundle.locales.php)

If it works, great! But if not, I can't accept the changes.

ycs77:
1. Laravel-lang is just a suite for collecting language pack files, This is done by copying the language pack into Laravel's language folder, Without it, the application works as well.
2. Sorry, something went wrong. The main zh-TW edits have been uploaded.

fletch3555:
I don't understand why you've renamed `set_view` to `select_view`, since that's completely unrelated to everything else in this PR.  Also, you didn't change any usages of that key, so it's almost guaranteed to break something.

Don't submit PRs with multiple distinct changes, because for this exact reason, I can't accept one and decline the other.

Also, you haven't answered my comments above.  We can't blindly change all these files on a whim.

For these reasons, I'm closing this PR.  You're welcome to separate the changes into multiple PRs and add more details around the reasoning for these changes, and we'll happily review it again.

PR #3749, opened 2018-11-13
Doc update, User Model extend / Trait Implementation

SteamWind:
Updating the doc to help user to integrate voyager in theyre app: Extending User Model or implementing User Contract and Trait.

I'm open to any typo suggestion 😋

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -72,12 +72,36 @@ And you will be prompted for the users name and password.
 
 This section is meant for users who are installing Voyager on an already existing Laravel installation or for users who want to perform a manual install. If this is not the case, you should go back to the [installation](installation.md) documentation or skip this section.
 
+### Publish assets
 The first thing you should do is publish the assets that come with Voyager. You can do that by running the following commands:
 
 ```bash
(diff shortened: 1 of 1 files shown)


fletch3555:
That's technically not true for the JWTAuth package (see my reply in slack). Let's refrain from mentioning other packages, and just say something like "In the case where you need to extend from another class (i.e. if another package requires it), then Voyager allows this instead:"

SteamWind:
I mentioned both. I think it's a common way to use JWTAuth. If we don't mention it there, where should we do it? In the JWTAuth Doc? That's less relevant...
It's not because you use JWTAuth that you use Voyager but the other side is more relevant. Using Voyager to maintain an API is a good Idea I think.
If you find we shouldn't mention it, I'll cut it.
I could just say "Note that Voyager User extends `Illuminate\Foundation\Auth\User`"?

SteamWind:
Okay, I removed the JWTAuth exemple but I just mention it to help people know that they don't need to extend the Illuminate\Foundation\Auth\User.
Hope it serves this case without saying too much about JWTAuth.

PR #3754, opened 2018-11-18
Fix bug#

alnouirah:
Fix a bug: When you change the direction  to rtl, the User Options button does not work.
issu


emptynick:
This needs to go to `1.x` because users need to re-publish assets.

publishable/assets/css/rtl.css
```diff
@@ -325,3 +325,26 @@
         float: right;
     }
 }
+
+
+[dir="rtl"] .app-container .content-container .navbar-top,
+[dir="rtl"] .app-container .side-body,
+[dir="rtl"] .app-container.expanded .content-container .navbar-top {
+    transition: all .5s cubic-bezier(.19,1,.22,1);
```
emptynick: Voyager does not use this transition when opening the user-dropdown, so I think neither should RTL-mode.

emptynick (changes requested):

publishable/assets/css/rtl.css
alnouirah: Sorry for the late response. I will look at it when I finish my work.

emptynick (approved):

PR #3758, opened 2018-11-21
- Added toggle to top of BREAD browse view if current model uses Soft…

### LINKED ISSUE #2165: Soft delete Admin
- Laravel Version: 5.5.*
- Voyager Version: the last one
- PHP Version:7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )
- Database Driver & Version: mysql

### Description:

Hi thank you it was really simple to get the soft delete working
the problem is that i cannot load the deleted items in the admin panel!!
i need to restore some items or edit them!!

as u can see the table is empty but in the DB i already have some deleted products!

mattrabe:
- Added toggle to top of BREAD browse view if current model uses SoftDeletes
- When toggle is clicked, redirect the user to ?showSoftDeleted={bool}, preserving existing query params
- In VoyagerBaseController->index() apply ->withTrashed() only if model uses SoftDeletes and showSoftDeleted query param is true

Fixes #2165


resources/views/bread/browse.blade.php
```diff
@@ -289,5 +293,28 @@
...
+                    'filter' => $search->filter,
+                    'key' => $search->key,
+                    'order_by' => $orderBy,
+                    'sort_order' => $sortOrder,
+                ];
+            @endphp
+            $(function() {
+                $('#show_soft_deletes').change(function() {
+                    if ($(this).prop('checked')) {
+                        $('#dataTable').before('<a id="redir" href="{{ (route('voyager.'.$dataType->slug.'.index', array_merge($params, ['showSoftDeleted' => 1]), true)) }}"></a>');
```
mattrabe, line 313: I am creating and then auto-clicking an anchor here, because `window.location.href = ''` failed due to the fact that Laravel's `route()` function returns a urlencoded string. ie, `?param1=1&amp;param2=2`. Those `&amp;`s cause `location.href =` to fail, redirecting the user to an invalid URL.

mattrabe:
I just realized that this also needs to include an Undelete feature - will add next chance I get.

mattrabe:
Okay - undelete/restore feature added

emptynick:
Please dont commit the compiled `app.css`.
Github will do this automatically after merging your PR.
Otherwise it will conflict with other PRs.

mattrabe:
My bad - is it worth going back and changing the commit? Also - maybe app.css should be .gitignore'd? (I'm sure someone already thought of that and there's a reason it's not...)

emptynick:
Just delete the file from your fork and commit again.
I'm not really sure, but my guess is that it would be ignored even when pulling from Github, therefore it would be missing when using a VCS in composer.

mattrabe:
Done. I deleted publishable/assets/css/app.css and re-committed.

Re: gitignore - yea, compiled CSS files are a pain in VCS. Wish there was a better way

emptynick:
Sorry, false information from me.
It needs to be the current app.css

mattrabe:
Okay - history rewritten to include app.css verbatim from 1.1 branch (still shows in commit, but content is unchanged), and my two smaller clean-up commits all in the one commit

fletch3555:
@mattrabe hate to be a pain, but it's still there and needs to be purged.  Simply having the "same content" isn't enough, since it likely varies by some hidden character (i.e. whitespace).  Any change you commit will cause merge conflicts with our build pipeline, causing it to fail to commit.

I can fix it for you if needed, so long as I have write access to your fork (not always the case)

mattrabe:
@fletch3555 Go for it - I just gave you and @emptynick collaborator invites on https://github.com/mattrabe/voyager - branch is `ticket-2165` - do whatever you need to do to get it to jam through the pipe!

emptynick:
I fixed the merge conflicts and restored the app.css.
You dont need to give us write access to your fork, just make sure `Allow edits from maintainers.` is ticked in your PR (it is, no worries)
f52e6cd9-d3f0-4c97-acd3-018edcbcf8e4

emptynick:
I merged the conflicts again and tested it.
Its working fine, only thing is that for me it shows `Sorry it appears there was a problem restoring this ...` but restores properly.

mattrabe:
Hmm this was not happening previously. I would have to take a look.

emptynick:
As per the documentation, `restore()` can return a bool or null.
Not sure why and when it'd do that tho.

mattrabe:
@emptynick I'm not able to reproduce the incorrect message.

PHP 7.2.8, Laravel 5.7.14, Voyager ticket-2165 branch

I have a model, `Project`:

```
<?php

namespace App;

use Illuminate\Database\Eloquent\SoftDeletes;

class Project extends Model
{
    use SoftDeletes;

    public function technologies()
    {
        return $this->belongsToMany('App\Technology');
    }
}
```

and the `projects` database table does have a column `deleted_at`.

I have a `Admin/ProjectController` that is used for the BREAD for this model, and it simply extends `\TCG\Voyager\Http\Controllers\VoyagerBaseController`.

```
<?php

namespace App\Http\Controllers\Admin;

class ProjectController extends \TCG\Voyager\Http\Controllers\VoyagerBaseController
{
}
```

When I go into Voyager admin -> Projects, I get the "Show Deleted" toggle at top. When I turn it on, I see soft deleted records, with the "Restore" button instead of the "Delete" button, as expected.
screen shot 2018-12-28 at 7 23 36 am

When I click "Restore" next one of the soft deleted records, the browser navigates to /admin/projects/{project_id}/restore which responds with a 302 to /admin/projects. The browser then loads /admin/projects, and I see the success message at top, as expected.
screen shot 2018-12-28 at 7 24 36 am

The Project that I "Restored" now appears as no longer soft deleted, as expected.

emptynick:
Yup, I figured out what happened.
`$model->restore()` just returns `$model->save()` after unsettings `deleted_at`.
I used Voyagers post model which overrides the `save()` method and doesn't return `parent::save()`.

Edit: but we need to put this to 1.x because you changed CSS.

emptynick:
I just commited 2 additional permission checks so people cant inject `?showSoftDeleted=1` without permission.

emptynick (approved):

fletch3555 (approved):

CupOfTea696:
Wouldn't it make more sense to just have a "trash" view you can switch to, that only shows deleted item? And the actions for deleted items should just be "delete permanently" and "restore"

mattrabe:
I’m agnostic on whether the list view is the “Trashed” filter or the “Include Deleted” filter.

I would like to vote that the actions for deleted items not be restricted, though. I value the ability to edit data for records that are not active on the front end (because they are soft-deleted), and I feel that restricting the capability serves no purpose currently. If we had an additional status for “Published” vs “Unpublished” I could see the value here - but lacking that, soft-deleted records are the only way to maintain “unpublished” items in the BREAD that are editable.

lucascnunes:
Really was looking for it, could I ask if you guys can add now a **Force** delete kind of option/button?

> Wouldn't it make more sense to just have a "trash" view you can switch to, that only shows deleted item? And the actions for deleted items should just be "delete permanently" and "restore"

I like being able to see they mixed on one list instead of 2(deleted and not deleted) but would like to have the force delete option with the restore.

PR #3759, opened 2018-11-23
Update relationship.blade.php

bxmas13:
Changed incorrect spelling; None results  -->  No results


emptynick:
Can't we use `__('voyager::generic.none')` so it's still translatable?

fletch3555:
@emptynick is correct.  This should use a translatable value, not a hard-coded string.

Also, you should always cut your branch off of the latest release branch(currently 1.1), not one of the dev branches (1.x).  This is such a simple change, it can easily go in as a patch for 1.1, but we can't merge it like this.  Please make your change again on a new branch off of 1.1 and open a new pull request.

PR #3760, opened 2018-11-23
Prevent MySQL error about foreign keys after running PHPUnit or Laravel Dusk tests

Martenvanurk:
After running PHPUnit test or Laravel Dusk test, a MySQL error is thrown when rollback the migrations. By refactoring the down method a bit, so it's compatible with the up method (unsignedInteger vs integer), I were able to tackle this MySQL error.


emptynick:
Closing as a duplicate of https://github.com/the-control-group/voyager/pull/3322

PR #3778, opened 2018-12-01
Update overriding-files.md

jetrodn:


emptynick:
Thats not completely correct.
`Voyager::useModel(...)` replaces any of [Voyagers models](https://github.com/the-control-group/voyager/tree/1.1/src/Models).
If you "only" want to replace a BREAD-model (Post, Page, Category), you dont need to use `Voyager::useModel(...)` at all.

PR #3803, opened 2018-12-13
Add new content type "Key-Value to JSON"

bnku:
In BREAD configuration:
default

Add new item / Edit item:
default

Final Data:
default


fletch3555:
Not sure how I feel about this.  At first glance, your data format is far too verbose.  Why not `{$key: $value, $key2: $value2}`?

bnku:
> Not sure how I feel about this. At first glance, your data format is far too verbose. Why not `{$key: $value, $key2: $value2}`?

I suggested above decision. But in this case, when converting JSON to PHP array, duplicate keys will be overwritten. Therefore, I rolled back the changes.

emptynick:
I think you should move this into a hook because it's very specific to your needs.

bnku:
> I think you should move this into a hook because it's very specific to your needs.

I agree. If I found good documentation on hooks, I would immediately did so.
Could you explain how to register new ContentTypes in a hook?

emptynick:
`php artisan hook:make myhook`
`php artisan hook:install myhook`
`php artisan hook:enable myhook`
Then https://docs.laravelvoyager.com/customization/adding-custom-formfields
Instead of the AppServiceProvider you'd use your HookServiceProvider of course.

bnku:
Transferred this decision into a hook:
https://github.com/bnku/key-value-json

emptynick:
I'm going to close this PR.
Feel free to add your hook to [awesome-voyager](https://github.com/larapack/awesome-voyager) and [larapack](https://larapack.io/).

PR #3805, opened 2018-12-13
Verify if the thumb resize is a string of null

migueldevpt:
In the SingleImage ContentType it's verified if the thumb resize is a string of null but not on the MultipleImage Class.
This code places the same verification on the MultipleImage Class.


emptynick:
Not really sure why one would use null as a string, but since we check the same for a single Image I guess this is fine?!

migueldevpt:
I do agree that it makes sense to set it to null only when it is null not when it is a string of null.

But if the SingleImage Class is set to only accept null instead of a string of null the default dummy seeder for the posts needs to also be changed because it uses for the image.details a string of null in the resize height option.

This being how if found this diference in the Single and MultipleImage Classes

emptynick:
You mean this?
https://github.com/the-control-group/voyager/blob/ada5e4abaf69588059fb01a874392c1bd4cabc7a/publishable/database/dummy_seeds/PostsTableSeeder.php#L140
Thats the height.

migueldevpt:
That's exactly the line. 
https://github.com/the-control-group/voyager/blob/ada5e4abaf69588059fb01a874392c1bd4cabc7a/src/Http/Controllers/ContentTypes/Image.php#L73-L75
If the previous line would be changed to only accept null and not 'null' would it not be required to also change that seeder?

fletch3555 (approved):

PR #3814, opened 2018-12-17
Show images on the ordering page

demiurge-ash:
If order's field is image type then we show the image, not the url.

order-images


emptynick (approved):
Very nice. Thanks!

PR #3827, opened 2018-12-21
Fix media-manager permissions

emptynick:
Autoload loads the constructor before the db-connection is made and therefore throws errors when using artisan.
Sorry for that, I could swear I tested this.


fletch3555:
I was wondering if this was going to be a problem... oh well

fletch3555 (approved):

cyrildewit:
Isn't the `Voyager::canOrFail()` method deprecated?

https://github.com/the-control-group/voyager/blob/1.1/src/Voyager.php#L262

PR #3828, opened 2018-12-21
If Model has json type in casts, I have error

mcmc4519:
Hi!
I think, json_encode should be remove, because json_encode has been saved into database as string: http://prntscr.com/lxyo3g

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -68,12 +68,12 @@
                                         @foreach($dataType->browseRows as $row)
                                         <th>
                                             @if ($isServerSide)
-                                                <a href="{{ $row->sortByUrl($orderBy, $sortOrder) }}">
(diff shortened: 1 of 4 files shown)


emptynick:
You reverted the changes from [this PR](https://github.com/the-control-group/voyager/pull/3781/files#diff-cf377d287d89be3fc7d2a9c234b13774), is there any reason for this?
Also, this only works when a cast is used. So it will fail for everyone.

mcmc4519:
> You reverted the changes from [this PR](https://github.com/the-control-group/voyager/pull/3781/files#diff-cf377d287d89be3fc7d2a9c234b13774), is there any reason for this?
> Also, this only works when a cast is used. So it will fail for everyone.

1. Ok, can I delete changes from browse.blade.php?
2. Yes, you are right, it works only, if cast is used. Can I check about used cast on model or no?

mcmc4519:
I want to create new, clear pullrequest

PR #3830, opened 2018-12-22
Add homepage link in login.blade.php

cotiga:
My Christmas present, a link to homepage!
Yes, he sorely lacking me when I disconnect from Voyager.
I don't know about you, but for me it's very helpful.

### DIFF
--- a/resources/views/login.blade.php
+++ b/resources/views/login.blade.php
@@ -18,19 +18,19 @@
             background-color: {{ Voyager::setting("admin.bg_color", "#FFFFFF" ) }};
         }
         body.login .login-sidebar {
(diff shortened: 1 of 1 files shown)


fletch3555:
I can barely tell what you changed since you also made all those whitespace changes.

Having said that, Voyager should not presume your site is available at `/`.  So I don't think this we'll be able to add this into Voyager.  I'd like to get @emptynick's opinion though.

cotiga:
The addition of the link is from line 93 to 96
The rest is only small syntax corrections

fletch3555:
I'm going to decline this per my previous comment.

Also, in the future, please refrain from committing those whitespace/formatting changes.  All they do is pollute the pull request.

cotiga:
OK it's noted. Thank's

Issue #3655, opened 2018-10-20
Rtl ScrollY Problem


MmKargar:
- Laravel: v5.6
 - Voyager: v1.1
 - PHP: 7.2
 - Database: MySQl
hey guys . 
in rtl version voyager have css issue.
can anyone fix this problem?

![image](https://user-images.githubusercontent.com/11253027/47253035-ad091980-d45a-11e8-94bd-10fc3359182c.png)

in rtl version when my menus are too many and needs scroll , scrolling not showing very well and show the scroll bar on my menues.
i tried to fix it but i couldn't.
if anyone can , please help.

emptynick:
I can see what you describe, sometimes its working, sometimes its not.
Looks like the position is calculated somewhere in Javascript and it misses the changes to the menu-width sometimes.

MmKargar:
exactly this is javascript calculation , sometimes its work fine in chrome and IE but always showing it in mozilla.
and i try so much but it didn't solved. :neutral_face:

mohamed4shim:
i have the same problem when page first load, and it's hidden after clicking any link,any one can help me?

MmKargar:
the problem didn't solved and i just remove perfect scrollbar from voyager and use [simplebar](https://github.com/Grsmto/simplebar) its not rtl but its not so bad as perfect scrollbar rtl.
![image](https://user-images.githubusercontent.com/11253027/47964636-12abf700-e052-11e8-929e-fa0030a34e88.png)

i try so many rtl javascript scrollbars libraries but all of them have conflict.
perfect scrollbar and other scrollbars has no problem , the problem is conflict with voyager.

deva7mad:
hay @mitismirza how are you.
 can you explain how you fixed it ?

MmKargar:
@deva7mad  you need to remove perfectScroll from voyager . for that you have to clone the voyager and remove 
```
   window.perfectScrollbar = require('perfect-scrollbar/jquery')($);

    $('.side-menu').perfectScrollbar();
```
from voyager/resources/assets/js/ app.js 

 and remove 

```
@import "node_modules/perfect-scrollbar/dist/css/perfect-scrollbar";

/********** Add z-index to perfect-scrollbar **********/

.ps.ps--active-x>.ps__scrollbar-x-rail, .ps.ps--active-y>.ps__scrollbar-y-rail {
    z-index: 999;
}

/********** END perfect-scrollbar style **********/
```

[...]

Issue #3036, opened 2018-04-22
Uncaught TypeError: Cropper is not a constructor


lucentlogix:
- Laravel Version: 5.6.0
- Voyager Version: 1.1.0
- PHP Version: 7.2.4
- Database Driver & Version: Mysql 5.7

### Description:
cropperjs 1.3.1+ brakes cropping functionality
Uncaught TypeError: Cropper is not a constructor
cropperjs 1.3.0 is working fine for me, might be a good idea to lock this version until a fix can be produced. 

### Steps To Reproduce:
cd vendor/tcg/voyager
npm install
npm run dev
cd ../../../
php artisan vendor:publish --tag=voyager_assets --force
Navigate to and select an image in media
Click Crop
Inspect console output

emptynick:
Closing as #3832 was merged (into 1.x, to be released in a future 1.2 release)

Issue #2858, opened 2018-03-20
"SCRIPT445: Object doesn't support this action" error with TinyMCE on IE/Edge


netdelight:
- Laravel Version: 5.4.36
- Voyager Version: 1.0.7
- PHP Version: 7.1
- Database Driver & Version:MySQL 5.1

### Description:
On IE (Edge and 11), TinyMCE is not loading and the error "SCRIPT445: Object doesn't support this action" appears in the console.

### Steps To Reproduce:
Display Voyager admin and edit a page in Edge.

A workaround is to add this before the app.js is loaded :
```
<script type="text/javascript">
        window.tinymce = { suffix: "", base: "{{ voyager_asset('js') }}" }
</script>
```
(based on this solution : 
https://stackoverflow.com/questions/23732815/tinymce-does-not-discover-correct-plug-in-path-in-ie10-and-chrome )

emptynick:
Judging by [this issue](https://github.com/tinymce/tinymce/issues/1096) it was resolved in a newer version and will then be fixed by https://github.com/the-control-group/voyager/pull/3832

emptynick:
Closing as #3832 was merged (into 1.x, to be released in a future 1.2 release)

PR #3832, opened 2018-12-22
Remove unnecessary libraries, fix vulnerabilities

emptynick:
- Update Laravel-Mix from 0.x to 4.x
- Removed Axios (no reference/usage)
- Removed jquery-sortable (no reference/usage. Also required an insecure version of jQuery, thus `npm audit` complains)
- Removed doubled Vue from `devDependencies`
- Removed slugify package - we use our own
- Removed toastr static file and use dep instead
- Remove perfect-scrollbar static file - not used at all
- Remove match-height static file - not used at all
- Remove bootstrap-toggle static file - use from npm
- Remove jquery-cookie (not used, sidebar state is still stored)
- Remove bootstrap-tooltip - not used, bootstrap has a native tooltip
- Use TinyMCE from npm. Plugins/Themes/Skins get published automatically
- Use Datatables from NPM
- Delete jquery-disable-selection.js - not used at all (built-in to jQuery)
- Replaced nestable with a newer version from npm - everything still working  
- Fixed "Cropper is not a constructor" in versions > 1.31 (cropperjs exports `default` instead of `main`
 or `module`)  

**Important:**
Webpack now copies TinyMCEs plugins/themes/skins from the npm package, so updating should be a breeze.
It is possible that our `compile assets script` does not care about those files and doesn't commit them.
They can be included in the `publishable` dir with every PR tho.

Fixes https://github.com/the-control-group/voyager/issues/3036, fixes https://github.com/the-control-group/voyager/issues/2858 fixes https://github.com/the-control-group/voyager/issues/3655


fletch3555:
You good here, or have more to add?

fletch3555 (dismissed):

fletch3555 (approved):

PR #3834, opened 2018-12-22
Dynamically load assets without publishing

emptynick:
This loads all assets directly from `publishable/assets` instead of publishing them.
This has various advantages, for example changing js/css in a patch version.
Also developing js and css is easier with it.


Braiani (approved):
I just tested and worked like a charm!

fletch3555 (approved):

fletch3555:
This is a MUCH better solution! great job

Issue #3604, opened 2018-10-05
Voyager doesn't fetch data with relationship 


DrudgeRajen:
## Version information

 - Laravel: v.5.1
 - Voyager: v1.1
 - PHP: 7.2
 - Database: postgres 9.5.3

## Description

In **BreadRelationshipParser** Trait, **getRelationships()** method always return empty array since it is checking if the **details** column of **Data Rows Table**   has **relationship**   . But while saving bread with relationship, bread doesn't save relationship field on **details** column on Data Rows table.
```
  $dataType->browseRows->each(function ($item) use (&$relationships) {
            $details = json_decode($item->details);
            if (isset($details->relationship) && isset($item->field)) {
           ....
```

## Steps To Reproduce
1.Create a Bread with any relationship 
2. Check detials column of Data Rows Table on Database

emptynick:
Simply looking at the code/database I can confirm that.
The condition should be
`if ($details->type == 'relationship')`

DrudgeRajen:
@emptynick  Exactly. :+1:

abhinav-rabbit:
> Simply looking at the code/database I can confirm that.
> The condition should be
> `if ($details->type == 'relationship')`

I just re-confirmed on the database that, if the datatype is "relationship" we have data in $item->type as 'relationship', but in $details->type we have value as "belongsTo".

![datarow](https://user-images.githubusercontent.com/35209172/46590834-c1efbf80-cae0-11e8-85fe-424da14ed475.png)
![detail_colunm](https://user-images.githubusercontent.com/35209172/46590835-c1efbf80-cae0-11e8-9d16-39cf0f0fd9ea.png)

ntuple:
> Simply looking at the code/database I can confirm that.
> The condition should be
> `if ($details->type == 'relationship')`

I just re-confirmed on the database that, if the datatype is "relationship" we have data in **$item->type** as **'relationship'**, but in **$details->type** we have value as **'belongsTo'**.

DataRows Table's fields having relationship data :
![datarow](https://user-images.githubusercontent.com/380662/46590859-e51a6f00-cae0-11e8-90f7-46e813daa950.png)

Same **DataRows** table's **detail** field data
![detail_colunm](https://user-images.githubusercontent.com/380662/46590860-e5b30580-cae0-11e8-9e87-39e52f847072.png) 

But is Code we have the above.

DrudgeRajen:
[...]

Issue #2561, opened 2018-01-29
trouble with model relationship using $with


faghihi:
- Laravel Version: 5.5
- Voyager Version: latest
- PHP Version: 7.1
- Database Driver & Version:

### Description:

I have a model which i have added the protected $with = ["skills"] in it 
now when i make the voyager bread for it I keep seeing errors on every page.

fletch3555:
1. `latest` is not a valid Voyager version.  Please share the ACTUAL version reported by composer
2. It would be helpful if you shared the actual error you're seeing.  We may be quite skilled in the mystical arts, but I assure you we're not mind-readers.

ravilushqa:
Hello! I have same problem.
Laravel Version: 5.4
Voyager Version: 1.0.14
PHP Version: 7.0
Database Driver & Version: MariaDB-10.2.1
More information:

when model has parameter $with, $item->getRelations() returns that contains in $with, in my option it's array of four elements.

Than it's try to find this relations in data_rows,
$dataType->browseRows->where('field', $field)->first();

it's returns null and on next line it's failed on $bread_data->details because $bread_data is null
**File where it's fails**
```
/**
     * Create the URL for relationship's anchors in BROWSE and READ views.
     *
     * @param Model    $item     Object to modify
     * @param DataType $dataType
     *
     * @return Model $item
     */
    protected function relationToLink(Model $item, DataType $dataType)
    {
        $relations = $item->getRelations();

        if (!empty($relations) && array_filter($relations)) {
            foreach ($relations as $field => $relation) {
                if (isset($this->relation_field[$field])) {
                    $field = $this->relation_field[$field];
                } else {
                    $field = snake_case($field);
                }

                $bread_data = $dataType->browseRows->where('field', $field)->first(); //<-- Null !!!

                $relationData = json_decode($bread_data->details)->relationship; //<-- Fail

                if ($bread_data->type == 'select_multiple') {
                    $relationItems = [];
                    foreach ($relation as $model) {
                        $relationItem = new \stdClass();
                        $relationItem->{$field} = $model[$relationData->label];
                        if (isset($relationData->page_slug)) {
                            $id = $model->id;
                            $relationItem->{$field.'_page_slug'} = url($relationData->page_slug, $id);
                        }
                        $relationItems[] = $relationItem;
                    }
                    $item[$field] = $relationItems;
                    continue; // Go to the next relation
                }

                if (!is_object($item[$field])) {
[...]

PR #3836, opened 2018-12-25
Remove unnecessary relationship-code

emptynick:
Method `getRelationships` in `BreadRelationshipParse` and the whole `HasRelationship` Trait relies on relationship-method-names which we dont use/know.
Therefore it is not usable/useful and I removed it completely.
Because of that `relationToLink()` doesn't do anything also and I removed everything related to it.

Sideeffect: `$with` in models is working again.
Fixes https://github.com/the-control-group/voyager/issues/3604 and fixes https://github.com/the-control-group/voyager/issues/2561


DrudgeRajen:
> $with in models is working again. 

Can anybody explain ?

abhinav-rabbit:
@emptynick **I think this is just code clean up**. This PR just remove all the remaining code related to relationship with the use of Eager loading capabilities, which was not working (which had been deprecated i guess).

As I had stated it the comment is the issue:
https://github.com/the-control-group/voyager/issues/3604#issuecomment-427711651
**"We have data in $item->type as 'relationship', but in $details->type we have value as 'belongsTo'."**

This PR removes feature related to eager loading of the relation once and for all. 

@DrudgeRajen "$with in Model" will not work for sure, Since we don't have any relationship defined on Model level and cant be eager loaded in any was possible.

ntuple:
@emptynick **I think this is just code clean up**. This PR just remove all the remaining code related to the relationship with the use of Eager loading capabilities, which was not working (which had been deprecated I guess).

As I had stated it the comment is the issue:
https://github.com/the-control-group/voyager/issues/3604#issuecomment-427711651
**"We have data in $item->type as 'relationship', but in $details->type we have value as 'belongsTo'."**

This PR removes feature related to eager loading of the relationship once and for all. 

@DrudgeRajen "$with in Model" will not work for sure, Since we don't have any relationship defined on the Model level and cant be Eager loaded in any way possible.

emptynick:
> This PR removes feature related to eager loading of the relationship once and for all.  

Because all those features solely rely on relationship-method names. So it can't and will never work with the way we currently define relationships. Why keep it then?!

> @DrudgeRajen "$with in Model" will not work for sure, Since we don't have any relationship defined on the Model level and cant be Eager loaded in any way possible.  

If you define `$with` in your model it will work again. I can guarantee you that.
See issue #2561

DrudgeRajen:
@emptynick I understand `$with` will work but what if dev will not keep `protected $with=['methodname'] `? Eager Load will not work ? I think the method that are removed were  somehow handling these things, might be in older version.

emptynick:
`$with` is a feature of Laravel/Eloquent.
And because we query relationships directly through the foreign model (not through the method) we don't need to eager-load the relationship before.

fletch3555 (approved):

yueyanzhao:
when updated to v1.1.11, 
in my project,there is a error,
"Method App\Http\Controllers\Voyager\CouponController::getRelationships does not exist.",
I only composer update, what I need to do

yueyanzhao:
@emptynick

yueyanzhao:
@emptynick

yueyanzhao:
@emptynick 
when updated to v1.1.11,
in my project,there is an error,
"Method App\Http\Controllers\Voyager\CouponController::getRelationships does not exist.",
I only composer update, what I need to do after this update? 
waiting for your reply, thank you!

emptynick:
Just remove the method call. It's unnecessary

yueyanzhao:
@emptynick ok, thank you

PR #3837, opened 2018-12-25
Fix media-manager permissions

emptynick:
Counterpart to #3827 for 1.1


fletch3555:
CanOrFail had been deprecated.  Maybe a better way to handle it?

fletch3555:
This should be broken out to a Gate (since Policies require a model class), and implemented with `$this->authorize('browse_media')` instead.

Since it's already here, I'll merge it, but the above change should be submitted as a new PR.

fletch3555 (approved):

PR #3856, opened 2018-12-30
Bread accessors

emptynick:
Allows to define accessors for every single BREAD action.
For example, to modify a prop `name` only for browse, you'd use
```php
public function getNameBrowseAttribute()
{
    return $this->name ? $this->name : 'Empty';
}
```


resources/views/bread/edit-add.blade.php
```diff
@@ -54,6 +54,9 @@ class="form-edit-add"
                                 <!-- GET THE DISPLAY OPTIONS -->
                                 @php
                                     $display_options = isset($row->details->display) ? $row->details->display : NULL;
+                                    if ($dataTypeContent->{$row->field.'_'.(!is_null($dataTypeContent->getKey()) ? 'edit' : 'add')}) {
```
fletch3555, line 57: This check for edit/add is done in a number of places... perhaps it should be pulled out to a single call at the top, or in a controller (i.e. `$mode = .....`)

Can probably be done in a separate PR though

fletch3555 (approved):

PR #3857, opened 2018-12-30
Add links and beautify relation in bread views

vince83110:
This simple PR just improve the visibility of the relations. I used "badge" system for the browse view, and I add links to the related entity for each relation.


emptynick:
The link part is very similar to https://github.com/the-control-group/voyager/pull/3695 where I already commented that I'm not a friend of.

vince83110:
Yes I just saw, I will change the way to get the slug and update it. My PR is more complete by the way on the view part.

emptynick:
If you find a way to do this in the controller, then please do it.
Also your PR reverts the changes from https://github.com/the-control-group/voyager/pull/3444. Please don't do that.

vince83110:
I purpose to store the slug in the details of the relation. It's possible to update the slug when you update the bread. I think the simplest solution.

resources/views/formfields/relationship.blade.php
```diff
@@ -15,7 +17,15 @@
                 @endphp
 
                 @if(isset($query))
-                    <p>{{ $query->{$options->label} }}</p>
+                    @if (isset($options->slug) and isset($query->id))
```
emptynick: Please don't use hardcoded `id` instead use `->getKey()`

resources/views/formfields/relationship.blade.php
```diff
@@ -64,29 +82,41 @@
                     $relationshipData = (isset($data)) ? $data : $dataTypeContent;
                     $model = app($options->model);
             		$selected_values = $model::where($options->column, '=', $relationshipData->id)->get()->map(function ($item, $key) use ($options) {
-            			return $item->{$options->label};
+            			return [
+            			    'label' => $item->{$options->label},
+                            'id' => $item->id ?? null
```
emptynick: Null coalesce is PHP > 7 only and we still support PHP 5

resources/views/formfields/relationship.blade.php
```diff
@@ -15,7 +17,15 @@
                 @endphp
 
                 @if(isset($query))
-                    <p>{{ $query->{$options->label} }}</p>
+                    @if (isset($options->slug) and isset($query->id))
+                        <p>
+                            <a href="{{ route('voyager.'. $options->slug .'.show', ['id' => $query->id]) }}" target="_blank">
```
emptynick: Same

resources/views/formfields/relationship.blade.php
```diff
@@ -117,57 +147,69 @@
                 @php
                     $relationshipData = (isset($data)) ? $data : $dataTypeContent;
                     $selected_values = isset($relationshipData) ? $relationshipData->belongsToMany($options->model, $options->pivot_table)->get()->map(function ($item, $key) use ($options) {
-            			return $item->{$options->label};
+            			return [
+                            'label' => $item->{$options->label},
+                            'id' => $item->id ?? null
```
emptynick: Same as above

src/Models/DataType.php
```diff
@@ -263,24 +268,4 @@ public function setOrderDisplayColumnAttribute($value)
...
-
-    public function getOrderDirectionAttribute()
-    {
-        return isset($this->details->order_direction) ? $this->details->order_direction : 'desc';
-    }
-
-    public function setOrderDirectionAttribute($value)
-    {
-        $this->attributes['details'] = collect($this->details)->merge(['order_direction' => $value]);
-    }
```
emptynick: Please don't remove existing code without a reason

emptynick (changes requested):

juliusvdijk:
I second this feature.
It is mind blowing that it currently is not possible.

@emptynick any update on the status of this?

fletch3555 (dismissed):

fletch3555 (approved):

Issue #2283, opened 2017-11-26
Audio files upload via media manager


ghost:
- Laravel Version: 5.4
- Voyager Version: 1.0
- PHP Version: 7.0
- Database Driver & Version: mysql

### Description:
How to enable audio files upload via media manager?

emptynick:
The problem is that Laravel (or PHP?) sometimes is not able to get the correct extension because there's no mimetype. It then stores it as a binary-file.

emptynick:
Closing as #3860 was merged (into 1.x, to be released in a future 1.2 release)

PR #3860, opened 2018-12-30
Keep media names, add number if exists

### LINKED ISSUE #1655: Media Upload Random Name
- Laravel Version: 5.4.38
- Voyager Version: 0.11

Hi,
I am developing an website that will need to upload images to the server, and the media, out-of-the-box, is awesome. But one requirement is to keep the names of the images the same, not the 20 chars random.

Can anyone help me?

Thanks in advance.

emptynick:
Uploading a file named `test.txt` into an empty dir:
1. Uploaded as `test.txt`
2. Uploaded as `test_1.txt`
3. Uploaded as `test_2.txt`

Uploading a file named `test_1.txt` into an empty dir:
1. Uploaded as `test_1.txt`
2. Uploaded as `test_2.txt`
3. Uploaded as `test_3.txt`

And so on.
Tested in multiple sub-directories.
Also fixes a problem uploading files with unknown mimetypes, because Laravel generates the extension based on the mimetype, which is undefined in some cases.
Fixes #1655 fixes https://github.com/the-control-group/voyager/issues/2283 closes https://github.com/the-control-group/voyager/pull/3378


src/Helpers/helpers.php
```diff
@@ -20,3 +20,15 @@ function voyager_asset($path, $secure = null)
         return route('voyager.assets', $path);
     }
 }
+
+if (!function_exists('get_file_name')) {
+    function get_file_name($name)
```
fletch3555, line 25: Not sure I like this as a helper function... in fact, I think we should move existing helper functions out into service classes or something anyway.  We can do that in another PR

fletch3555 (approved):

maxkopych:
Do we have this option to keep image names for image field in BREAD?

PR #3867, opened 2019-01-03
RTL fix, add Arabic translation and config fix

mrkindy:
RTL fix, add Arabic translation and config fix


resources/views/master.blade.php
```diff
@@ -37,6 +37,11 @@
         .voyager .breadcrumb a{
             color:{{ config('voyager.primary_color','#22A7F0') }};
         }
+        @if(config('voyager.multilingual.rtl'))
```
emptynick: This is not necessary anymore. It was taken care of in https://github.com/the-control-group/voyager/pull/3754

emptynick (dismissed):

resources/views/master.blade.php
mrkindy: Done

mrkindy:
Done

emptynick:
I think this is fine then?
It will fall-back to the given language-string if it doesn't exist anyway.

publishable/config/voyager.php
```diff
@@ -132,17 +132,17 @@
     'dashboard' => [
         // Add custom list items to navbar's dropdown
         'navbar_items' => [
-            'Profile' => [
+            'voyager::generic.profile' => [
```
emptynick, line 135: The same changes you made to this file also need to be done in `voyager_dummy.php`

emptynick (changes requested):

publishable/config/voyager.php
mrkindy: Done

emptynick (approved):

emptynick:
Documented in https://github.com/the-control-group/voyager/pull/3909/commits/aba010917cd0d2e037db5e0f97e0c40415420744

Issue #3873, opened 2019-01-06
[BUG] Field 'form_id' doesn't have a default value


andrey-helldar:
## Version information

 - Laravel: v5.7.19
 - Voyager: v1.1.11
 - PHP: 7.2.11
 - Database: MySQL 8.0.13

## Description
When you try to create a record with a relationship, the system gives an error.

## Steps To Reproduce
1. In `Bread` set up a relationship with `Belongs To`.
2. When saving changes, the system throws an error:
```sql
General error: 1364 Field 'form_id' doesn't have a default value
SQL: insert into `forms_fields`
(`slug`, `type`, `title_en`, `title_ru`, `title_zh_cn`, `sort`, `updated_at`, `created_at`)
values
(qwdqwd, text, qwdqwd, rgerthrth, 3445h4h, 34, 2019-01-07 06:00:03, 2019-01-07 06:00:03)
```

#### 1
![2019-01-06_23-53-33](https://user-images.githubusercontent.com/10347617/50741636-cf683300-1210-11e9-8320-97a35fef225a.png)
#### 2
![2019-01-06_23-53-41](https://user-images.githubusercontent.com/10347617/50741637-d131f680-1210-11e9-8647-12088f70989f.png)
#### 3
![2019-01-06_23-54-01](https://user-images.githubusercontent.com/10347617/50741638-d2632380-1210-11e9-9c7e-5b4a5b27bf56.png)
#### 4
![2019-01-06_23-59-16](https://user-images.githubusercontent.com/10347617/50741639-d42ce700-1210-11e9-8df6-11307e0ac724.png)
#### 5
![2019-01-06_23-59-35](https://user-images.githubusercontent.com/10347617/50741641-d68f4100-1210-11e9-9083-74e1b03401b2.png)
#### 6
![2019-01-07_00-00-40](https://user-images.githubusercontent.com/10347617/50741642-d8590480-1210-11e9-9e20-9cc2c6dbdf84.png)
#### 7
![2019-01-06_23-59-53](https://user-images.githubusercontent.com/10347617/50741643-da22c800-1210-11e9-9c7d-5b4d605ff376.png)
#### 8
![2019-01-07_00-00-17](https://user-images.githubusercontent.com/10347617/50741644-dc852200-1210-11e9-970d-87c7c612d2ed.png)
#### 9
[...]

PR #3875, opened 2019-01-06
Fixed `field 'form_id' doesn't have a default value` (#3873)

andrey-helldar:
See #3873 PR


emptynick:
Closing as this is desired behaviour.
Disallowing a field to be written will end up in a field not being written - makes sense in my eyes.

See discussion in #3873

PR #3893, opened 2019-01-16
update relationship.blade.php

avenCN:
fixed relationship query use key  (user setting through admin panel) replace default id will return null.


emptynick:
Makes sense. Thanks!

emptynick (approved):

PR #3894, opened 2019-01-16
Add the feature to serialize the Translator model using json_encode

marijnz0r:
Add the JsonSerializable interface to the TCG\Voyager\Translator class to be able to use the json_encode feature. See https://secure.php.net/manual/en/class.jsonserializable.php


vblinden (approved):
Very nice!

emptynick (approved):

PR #3898, opened 2019-01-18
Correcting code impact of DB

Epistol:
If $row->detail is not well rendered as the DB side, it will generate a text content, thus it can't read the inside of it


emptynick:
Its is already decoded here:
https://github.com/the-control-group/voyager/blob/82c8d67479afbe088c30f23969c6f1aa0ab227e3/src/Models/DataRow.php#L64-L67
There's no reason to do it again, it will even fail (see your tests)

PR #3910, opened 2019-01-23
fixed submit quickly click create many records bug.

avenCN:
If you quickly click create button will appear many records, so here must disabled button submit form.


emptynick:
This has been fixed in https://github.com/the-control-group/voyager/pull/3890 for 1.x already.
It completely removes Ajax validation.

PR #3912, opened 2019-01-24
Issue 1695

### LINKED ISSUE #1695: support array field in BREAD system
- Laravel Version: 5.4.30
- Voyager Version: 0.11.14
- PHP Version: 7.1.19
- Database Driver & Version: mysql 5.7

### Description:
I used the [attribute casting](https://laravel.com/docs/5.4/eloquent-mutators#attribute-casting) to automatically convert an `information` field to array.
When I try to use BREAD to read or edit the field (as `text_area`), an exception `htmlspecialchars() expects parameter 1 to be string, array given` is thrown.
Now I override the template `bread/read.blade.php` and `formfields/text_area.blade.php` to bypass.
But I'd like to know whether there is a better way or could be added as an feature later.

shaunthegeek:
Fix #1695


fletch3555:
Pull requests should have descriptive titles and ideally a description of what is being fixed and why (the why can include a link back to the original issue). What you've created here is useless to us without more info

Issue #3871, opened 2019-01-04
"Trying to get property 'type' of non-object"


priyag597:
when try to edit admin profile i am facing this issues 

I checked into database as well in edit column 1 was already  there . 
 
Fixed by steps bewlow:

Go to phpmyadmin
find "user_belongsto_role_relationship" && "user_belongstomany_role_relationship" at field column
Chang from "0" to "1" in Edit column

emptynick:
Please use the issue template. Otherwise we are not able to help you.

priyag597:
Laravel Version: 5.7.14
Voyager Version: 1.1.1
PHP Version: 7.2

"Trying to get property 'type' of non-object"   when I trying to edit admin user I am getting this issues , i had gone through the already raised https://github.com/the-control-group/voyager/issues/3159 solution but it is same as it was .

find "user_belongsto_role_relationship" && "user_belongstomany_role_relationship" at field column
Chang from "0" to "1" in Edit column 

I found this solution but it dosent work for me.

priyag597:
https://github.com/the-control-group/voyager/issues/2431#issuecomment-451375112

I got this issues as well but i didnt get any response on it .please help me out

priyag597:
please reply somebody

priyag597:
why dont you reply @emptynick 
i have just got an stuck .

emptynick:
You still did not provide the necessary informations from the issue-template.
For example the steps to reproduce.
Because you linked to an issue which was about upgrading...did you upgrade?
Are you sure you are on 1.1.1?

vaggelis2018:
Updated to 1.1.1 would fix the issue i had the same problem. Just check if you have override the controller

priyag597:
## Version information

 - Laravel: v5.7.14
 - Voyager: v1.1.11
 - PHP: 7.2
 - Database: 10.1.37-MariaDB

## Description

![image](https://user-images.githubusercontent.com/46341574/50757931-bd28dc00-1286-11e9-8c8c-c6ea66a0eab0.png)

1. When I am trying to view users on click users sidebar nav 
2. And admin user profile when I try to edit (path= "right side click on avtar ->click on profile ->Edit my profile -> error  ")
 I am getting following this error

## Steps To Reproduce
Steps to reproduce the behavior:
1. Go to 'login into admin'
2. Click on 'users nav sidebar'
4. See error  Trying to get property 'type' of non-object

## Screenshots
[...]

PR #3924, opened 2019-01-30
Trying to get property 'type' of non-object issue.

mhmelshaaer:


resources/views/tools/bread/relationship-partial.blade.php
```diff
@@ -1,4 +1,4 @@
-@php $relationshipDetails = $relationship['details']; @endphp
+@php $relationshipDetails = json_decode($relationship['details']); @endphp
```
emptynick, line 1: This attribute is already casted here. I don't see a reason why we should `json_decode`
 it again
https://github.com/the-control-group/voyager/blob/6c9d6289bedd7e8db22b020aabd5789aa20d7dbc/src/Models/DataType.php#L244

emptynick (changes requested):

resources/views/tools/bread/relationship-partial.blade.php
mhmelshaaer: When browsing or editing User BREAD after a fresh voyager installation, for some reason it throws "try to get property of non-object" error, on editing line 54 in relationship-partial.blade.php "$relationshipDetails->type" throws an error and on browsing line 16 in BreadRelationshipParser.php "$row->details->type"  throws an error. This issue happens a lot, I tried solving it by editing those lines and it worked, and now, magically, after a while, it throws an error for those edits and works just fine after undoing the edits.

emptynick: See https://github.com/the-control-group/voyager/issues/3871
Turns out there was invalid (not installed by default) json in his database

MouhammedElshaaer: yes, I've checked that, but it was already set to 1. Maybe I'll try another fresh voyager installation and will see if this happens.

mhmelshaaer: Oh, I see. Thanks, I will close this after you see this comment.

PR #3926, opened 2019-01-31
Included Laravel version support

guizoxxv:


fletch3555:
I'm not sure this is the right place for this information.

PR #3586, opened 2018-09-26
Add Ajax relation support - from #3448

vince83110:
Using #3448 , I made new PR with more clean code and commit. I also corrected some behaviors in the controller.

I didn't make any test since it depends on custom models and custom relations, and it's only code added.

Fixes #3581 closes #3448

PR #3927, opened 2019-02-01
Easier customization of BREAD features

DODMax:
Hi there Cap'tain(s),
I've been using Voyager for quite some time and upgrading between major version has always been a bit difficult since I have rather customized BREAD views. Here is a PR that I hope solves part of this issue (in my case at least, I recognize it's probably biased).

**Customized BREAD views**
*Problem:* Custom layouts for BREAD views are hard to maintain between upgrades.
*Solution:* Split BREAD to put the magic into partials so those bits are easy to include when overriding a Voyager view. For example `bread.browse` becomes:
```blade
@extends('voyager::master')

@section('page_title', __('voyager::generic.viewing').' '.$dataType->display_name_plural)

@section('page_header')
    <div class="container-fluid">
        @section('browse-page-title')
            <h1 class="page-title">
                <i class="{{ $dataType->icon }}"></i> {{ $dataType->display_name_plural }}
            </h1>
        @show

        @include('voyager::bread.partials.browse_actions')
        @include('voyager::multilingual.language-selector')
    </div>
@stop

@section('content')
    <div class="page-content browse container-fluid">
        @include('voyager::alerts')
        <div class="row">
            <div class="col-md-12">
                <div class="panel panel-bordered">
                    <div class="panel-body">
                        @include('voyager::bread.partials.browse_search')
                        <div class="table-responsive">
                            @include('voyager::bread.partials.browse_table')
                        </div>
                        @include('voyager::bread.partials.browse_footer')
                    </div>
                </div>
            </div>
        </div>
    </div>

    @include('voyager::bread.partials.browse_modal-delete')
@stop

@include('voyager::bread.partials.browse_scripts')
```
**Custom Relationship Dropdown**
*Problem:* Relationship select in edit forms rely on the `all()` method to display the list of available options, but sometime you need to narrow down that list.
*Solution:* Look for a method on `$dataTypeContent` that returns the available options, otherwise fallback to all. For example, if you want to customize the list of users for your Task model:
```php
Task extends Model {
    public function memberIdList()
    {
        return Member::where('active', 1)->orderBy('created_at')->get();
    }
}
``` 
Note this was initially introduced in #1052 but never ported to the new relationship system.

**Ability to customize VoyagerBaseController**
Problem: VoyagerBaseController is hard to extend, you have to override the whole method in your own controller.
Solution: Added `CustomizableBread` trait that makes it possible to override key parts of the processing.

*The cases I identified so far*
**A/** Need to redirect somewhere else than the index after store/update/delete.
Added method in charge of providing redirect url:
```php
protected function getRedirectRoute(Request $request, $dataType, $data = null)
{
        return route("voyager.{$dataType->slug}.index");
}
```

**B/** Need to customize the query builder instance or the model before it gets fed to the BREAD processing.
For example if you want your index to only display active users you could override the following mehod:
```php
protected function getQueryForIndex(Request $request, $dataType)
{
        return app($dataType->model_name)->select('*');
}
```

Let me know you thoughts!


emptynick:
I think there are some more easy approaches:

**Customized BREAD views**
Instead of splitting all bread-files, you could use:
```
@yield('section')
My default content
@show
```

To override you then just call
```
@extends('bread....')
@section('section')
I will override the above content
@endsection
```

**Custom Relationship Dropdown**
Has been replaced by Ajax calls for 1.2 https://github.com/the-control-group/voyager/pull/3586

**Ability to customize VoyagerBaseController**
I think we just should introduce a parameter `$redirect = true` to all methods, so an overriden method could do:
```php
public function edit(...)
{
    parent::edit(..., false);
    //Do whatever you want to do afterwards

    return redirect(...);
}
```
This will give you the ability to do changes before AND afterwards

DODMax:
Great, let me reply in reverse order :-)

**3/** Indeed as long as we don't return the call the parent the `redirect` is not even needed. The only advantage of having the function is that we don't need to query `$dataType` and `$data` from the DB again but it does not really justify complicating the code.

**2/** To the 1.2!

**1/** Yes I do extend the bread views in this way, then override `content`, but I end up needing to copy most of the code generating the table, search & order fields, pages... from the bread view.
I would rather include those part as partials in my layout where I need them.

emptynick:
Closing as this shouldn't be necessary in v1.2 anymore.
Please let me know if you think it is necessary.

PR #3909, opened 2019-01-22
Small improvements/Upgrade guide

emptynick:
Added the upgrade-guide for 1.2:
- New configurations
- Asset-loading (remove old assets)
- User-BREAD controller

Fixed some small things:
- Simplified edit/add condition in bread-view. See https://github.com/the-control-group/voyager/pull/3856#discussion_r244538249
- Radio buttons didn't use the display-text, but the key instead. Fixes #3907 
- Use `empty()` instead of `count()` for multiple-select. Fixes #3879 
- Replaced hardcoded "BREAD". Fixes #3548 
- Lost order-params when searching and lost search params when ordering. Fixes #3923
- Added `default` prop to number formfield. Fixes #3503
- Renamed `Model namespace` to `Fully qualified Model name`. Closes #1908

Enhancements:
- Added customizable formfield-views. Fixes #3301 Closes #3738
- Added documentation for custom actions
- Action specific validation rules. Fixes #2034
- Added BREAD-scopes. Fixes #3941 
- Added advanced installation docs. Fixes #3719 
- Replaced user-models settings cast with an accessor/mutator-pair. Closes #3950

PR #3934, opened 2019-02-04
1.1 checkbox default filter

Emerica:
These commits attempt to make the checkbox more user friendly.
There is currently a documentation issue on the website which shows that a checkbox can have a default value set with 
`
{
"checked":"true"
}
`

This is not the case as the form field is explicitly looking for a Boolean.
The first commit changes this to use filter_var with a a Boolean filter, this should cast values properly, so json ints, strings, or bools could be used in the admin panel.

I personally copy pasted the documentation and not having things work was a bummer, learning a couple quotes could seemingly waste a few hours of my time... here we are.

Looking at the field it also became apparent it might be nice to have a custom class defined and thus commit 2, adds the ability for custom classes on checkboxes.

Commit 3, fixes some indentation and keeps things within an 80 col limit for us terminal folks.


Emerica:
`
{
    "on" : "On Text",
    "off" : "Off Text",
    "checked" : "true",
    "class" : "toggleswitch",
}
`

In Voyager by default,  a Check Box is converted into a toggle switch. 
The on key will contain the value when the toggle switch is on, and the off will contain the value that is set when the switch is off. 
The checked value is type insensitive and may be  true, "true" or 1 to be enabled by default.
The default will be off.
To override the toggleswitch functionality, the "class" option may be used.
On and off values are stored in the "data-on" and "data-off" attributes of the selected input field.

emptynick:
Thanks.
Nonetheless it does make sense to cover most possibilities.
Just mentioning that I changed the docs about this in https://github.com/the-control-group/voyager/pull/3909/commits/87a33225565534ed09eba71dcdad31a16876a94d

emptynick (approved):

PR #3939, opened 2019-02-07
Fix image create error on media upload

ElliottLandsborough:
realPath.file on line 227 doesn't exist yet because it has not been written yet - it gets written to the file system at line 230 with the copy function.

An exception occurs but the file gets written anyway. This patch stops the exception.

Steps to reproduce:
1) Navigate to media section
2) Open network tab in console
3) Upload an image to root folder (I was uploading a png ~ 30kb)
4) 500 error happens on /admin/media/upload

(laravel/framework 5.7.25, php 7.2.14, intervention/image 2.4.2, filesystem driver: s3)

### DIFF
--- a/src/Http/Controllers/VoyagerMediaController.php
+++ b/src/Http/Controllers/VoyagerMediaController.php
@@ -224,7 +224,7 @@ class VoyagerMediaController extends Controller
                 'image/svg+xml',
             ];
             if (in_array($request->file->getMimeType(), $imageMimeTypes)) {
-                $image = Image::make($realPath.$file);
+                $image = Image::make($request->file->getRealPath());
 
                 if ($request->file->getClientOriginalExtension() == 'gif') {
                     copy($request->file->getRealPath(), $realPath.$file);
--- a/src/Http/Controllers/VoyagerMediaController.php
+++ b/src/Http/Controllers/VoyagerMediaController.php
@@ -4,6 +4,7 @@ namespace TCG\Voyager\Http\Controllers;
 
 use Exception;
 use Illuminate\Http\Request;
+use Illuminate\Http\UploadedFile;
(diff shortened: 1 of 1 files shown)


emptynick:
`$realPath` is define here:
https://github.com/the-control-group/voyager/blob/1badfc9b3107cb9051f127ab4e1ebaba3193f6fd/src/Http/Controllers/VoyagerMediaController.php#L206
And `$file` here:
https://github.com/the-control-group/voyager/blob/1badfc9b3107cb9051f127ab4e1ebaba3193f6fd/src/Http/Controllers/VoyagerMediaController.php#L217

There is something else at play.

ElliottLandsborough:
Wow that was an extremely quick reply - yeah I agree re-looking at it. I think I've found out what it is so will submit another patch soon.

emptynick:
Please take a look at the 1.x branch.
 There were some changes to the file already.

ElliottLandsborough:
I have double checked the old code and made sure it fits in with the 1.x branch.

Why do we save the image to the laravel filesystem on line 217 but then re-save it on line 230 and 232 without using the filesystem?

This breaks uploads that need to be rotated for anyone using s3 because 217 correctly uploads to s3 and then 230/232 upload to storage/public and completely ignore the file system driver.

I have submitted a patch that fixes it to always use the storage driver - along with stopping it from writing the same file more than once.

ElliottLandsborough:
As a side note, the "$realPath.$file" variable is created but if the file has been uploaded to s3 the path doesn't actually exist on the local file system.

By 'doesn't exist yet' I meant that the file had not been written to the file system yet so we can't create a new image object with it.

My original solution was definitely wrong though, ha.

emptynick:
Closing. If you want/can fix the other problem, please open a new PR.

PR #3943, opened 2019-02-09
Add featured scope to the post model

hishamhadraoui:
Add featured scope to the post model in order to keep things simple and easy

### DIFF
--- a/src/Models/Post.php
+++ b/src/Models/Post.php
@@ -48,6 +48,18 @@ class Post extends Model
     }
 
     /**
+     * Scope a query to only Featured scopes.
+     *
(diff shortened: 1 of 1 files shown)


fletch3555:
This is something you can easily add to your own project. The Post model that ships with Voyager is just an example on how Voyager could be used

hishamhadraoui:
yes of course but that applies to the published scope too. for the sake of simplicity and the ability to use the voyager post model without adding or remaking it for a simple blog project or anything that can use the post model ... and needs a simple scope like this 👍

fletch3555:
I don't want to change these "example" models right now.  Closing

PR #3950, opened 2019-02-14
Illegal string offset 'locale'

workschan:
Create a collection from `$this->settings`

### DIFF
--- a/src/Models/User.php
+++ b/src/Models/User.php
@@ -38,6 +38,6 @@ class User extends Authenticatable implements UserContract
 
     public function getLocaleAttribute()
     {
-        return $this->settings['locale'];
+        return collect($this->settings)->get('locale');
     }
 }


emptynick:
That's effectively what the cast just a few lines above does.

CupOfTea696:
You're probably getting that error because the User model in your app overwrites the casts attribute. Just add an `array` cast for the `settings` attribute in your `app/User.php`.

```php
    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'email_verified_at' => 'datetime',
        'settings' => 'array',
    ];
```

The real solution would be for the installer to check if the User model has a `casts` property defined, and add this automatically.

emptynick:
Superseeded by #3909 commit https://github.com/the-control-group/voyager/pull/3909/commits/56e72e668a7c124f69172dbac1f12ba3bdad7272.

PR #3952, opened 2019-02-15
Let developers use a separate guard for auth users

alkhachatryan:
Most of us doesn't use the same guard (user guard) for both of admins and front users.

I'm not exception. I have two guards - users and admins. And I couldn't use your awesome package because of hardcoded auth guards.

In this PR you can see 12 changed  files, where I decided to separate the front and admin users guards.

### DIFF
--- a/resources/views/dashboard/navbar.blade.php
+++ b/resources/views/dashboard/navbar.blade.php
@@ -40,8 +40,8 @@
                     <li class="profile-img">
                         <img src="{{ $user_avatar }}" class="profile-img">
                         <div class="profile-body">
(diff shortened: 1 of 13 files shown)


emptynick:
But that guard doesn't exist by default. That's why your tests are failing

alkhachatryan:
So you can add in voyager installation and it will install it automatically

emptynick:
But we can't (technically) and won't merge a faulty PR.

emptynick:
We can't merge your PR as you deleted your fork. Closing.

PR #3953, opened 2019-02-16
Fixing Bug: Delete Menu Item

### LINKED ISSUE #2503: Menu builder wrong items being deleted
- Laravel Version: 5.5.28
- Voyager Version: 1.0.13
- PHP Version: 7.0.22-0ubuntu0.16.04.1
- Database Driver & Version: NA

### Description:
Deleting items in the menu builder doesn't work properly if a delete operation was canceled by the user.

### Steps To Reproduce:
- Go to the menu builder
- Hit delete on one of the menu items (ITEM_A)
- Cancel the delete modal by clicking outside of it or clicking cancel
- Hit delete on another item (ITEM_B)
- Confirm your action on ITEM_B
- The first item (ITEM_A) will be deleted instead

### Cause:

https://github.com/the-control-group/voyager/blob/97df19285b2b3c8f50bc3a3840702295477e366a/resources/views/menus/builder.blade.php#L274

Once the __id is replaced by an actual id (number), the replacement will no longer work.

### Possible fix:

Change the way the id is injected into the action.

### Finally:

Thank you very much for all the great work you've done with Voyager.

P.S. I _may_ submit a PR to solve this if I had enough time over the weekend.

prakhargvp:
Delete menu item is not changing the form action url.

onpaste 20190217-071907

Fixes #2503


emptynick:
Not working for me.
It doesn't replace anything and thus redirects to `/admin/menus/X/item/__id`

resources/views/menus/builder.blade.php
```diff
@@ -277,7 +277,7 @@
              */
             $('.item_actions').on('click', '.delete', function (e) {
                 id = $(e.currentTarget).data('id');
-                $('#delete_form')[0].action = $('#delete_form')[0].action.replace("__id",id);
+                $('#delete_form')[0].action = $('#delete_form')[0].action.replace("/[^\/]*$/",id);
```
emptynick: Not working for me, it doesn't replace anything.
Why not instead use:
```suggestion
                $('#delete_form')[0].action = '{{ route('voyager.menus.item.destroy', ['menu' => $menu->id, 'id' => '']) }}/' + id;
```

emptynick (changes requested):

resources/views/menus/builder.blade.php
prakhargvp: But we need to change the menu id in modal popup form action. 
It's dynamic because there is not a single menu they are so many with different menu id.

emptynick: We do that. It just attaches the dynamic ID to the "static" route `admin/menus/X/item/ + id`.
Try it, its working.

prakhargvp:
Ya that's fine too.. :-)

emptynick (approved):

Issue #3962, opened 2019-02-20
[BUG] User role is null sometimes


alkhachatryan:
## Version information
 - Laravel: v5.7.*
 - Voyager: v1.1.11
 - PHP: 7.2
 - Database: PostgreSQL 9

## Description
I installed the voyager on ready project where I have +800K records. The problem of the installation is: when I browse dashboard or users everything is ok. But when I go to Database I get the error.

Sometimes user role is null. When you open the Database page, this function works multiple times, as I understood - 2 times. When it's the first call, the user role is normal Role object, but when it's the second call, user role is null.

## Steps To Reproduce
1. Install the Voyager on ready project with multiple not empty tables (maybe this is the problem)
2. Go to Database
3. See error

## Expected behavior
I excepted to see the page of database. To do it, I wrote a code, which is checking, if $this-role is null - get the role of current logged in user.

## Screenshots
![image](https://user-images.githubusercontent.com/22774727/53113649-815a8500-355b-11e9-8963-987b97c10c0b.png)

## Additional context
Again. See the screenshot. If  you run this code, you will get an error, but if you uncomment the code everything will be ok. So I think you have to implement this code into your core and make a new release, or tell us how we can override this function, so we can do it in our source, not in vendor. 

Thanks

emptynick:
You use the Trait in your model, so you can just duplicate it to your namespace and include it instead of voyagers.

alkhachatryan:
Will it hurt the installation of the voyager on new machines?

alkhachatryan:
And as I understood, you will not fix it in new release?

alkhachatryan:
I'll do a PR now

emptynick:
I dont know what the issue is.
Effectively your workaround is the same as calling `$this->load('role')` again.
But to really fix this issue, we'd need to find out why its not loading the relationship at all, which is not easy because I can't replicate the issue (you are the first to report this)

alkhachatryan:
https://github.com/the-control-group/voyager/pull/3963

emptynick:
Closing as the PR was closed aswell.

PR #3963, opened 2019-02-20
Fixed the problem when user role is NULL

alkhachatryan:
Fixed the unknown issue related with NULL user role

### DIFF
--- a/src/Traits/VoyagerUser.php
+++ b/src/Traits/VoyagerUser.php
@@ -117,6 +117,9 @@ trait VoyagerUser
     private function loadPermissionsRelations()
     {
         $this->loadRolesRelations();
+        
+        if(!$this->role)
+            $this->role = Role::find( Auth::user()->role_id );
 
         if (!$this->role->relationLoaded('permissions')) {
             $this->role->load('permissions');


emptynick:
Same as in #3962.
This is basically just calling `$this->load('role')` again, which might fail again because it failed doing the same before.

fletch3555:
I don't really like this solution.  I'd rather see `->load('role')`, but as @emptynick said, it still may not solve the problem.

alkhachatryan:
That's why I prefer this way to solve it.

emptynick:
Overly dramatic visualization of whats happening and what you are doing:
```php
$value = null;
$value = 1; //aka $this->load('role'); < fails

if (!$value) {
    $value = 1; //aka $this->load('role'); < might fail again
}
```
Instead of just manually trying to assign the role (again) you/we should find out why its failing in first place.
Calling `Role::find( Auth::user()->role_id );` is effectively the same as `load('role')`.
It is failing the first time, and might fail the second time as well, and maybe even a third, fourth...time.

alkhachatryan:
@emptynick it's not the same.

If I remove this "dramatic" code, the error appears. It's really not the same

emptynick:
Take a look at the source.
`load(...)` does nothing else than `$this->x = Model::find(y)`.
It's more helpful to find the cause of the problem instead of "fixing" it by doing the same again - believe me.

wishbone-productions:
Same here, after updating the admin user it deletes the user_roles and nulls the role_id. looks like the bug is reintroduced. Maybe it has something to do with upgrading to laravel 5.8. Bug started to appear right after that.

php: 7.1.2
laravel: 5.8
voyager: 1.2.1

emptynick:
Did you upgrade from 1.1 and followed the upgrade guide?

wishbone-productions:
Yes done all off the steps as described.

Think this is part of the solution: https://github.com/the-control-group/voyager/issues/2451#issuecomment-470569735
But its a bit to dirty.

I'm not seeing rights the rights fields like, Default Role / Additional roles. Although i'm logged in as admin.

wishbone-productions:
line 63 in edit-add.blade.php is passing the user model instead of the roles model.

`@can('editRoles', $dataTypeContent)`

Should be: 

` @can('edit',  Voyager::model('Role'))`

emptynick:
Please show us a screenshot of your user bread settings

wishbone-productions:
https://github.com/the-control-group/voyager/issues/2451#issuecomment-470938098

Sineld solution above worked for me.

PR #3965, opened 2019-02-21
Added Modals, Added ability to add full detail of a model when has relationship

Xoshbin:
Voyager by default when a model have a relationship it adds only the field that's been added when making the bread. But this PR adds the ability to show a button beside those fields that have relationship and when clicked showing the complete form of that related bread. It's been tested all working well when adding filling the form it saves the data without leaving the page. what is missing is a javascript or ajax function to update the current row that have the button beside.


emptynick:
Please don't submit chmod changed (see [here](https://github.com/Gizra/KnowledgeBase/wiki/How-to-set-Git-to-ignore-chmod-changes)).
It makes it utterly hard for us to review your PR.

Xoshbin:
I did that, Do I have to do the PR again?

emptynick:
Its probably the easiest way, yes.
You could change the chmod back, but you'd need to change the value of `git config core.filemode` to true again

Xoshbin:
I'm afraid it's to much for me to find the default chmods so I may not be able to PR again. Sorry

Xoshbin:
I did another PR, But I'm not sure if I did the right thing, could you please check it out. Thanks

PR #3974, opened 2019-02-27
Laravel 5.8 support

emptynick:
- Replaced `str_*` and `array_*` helper with the corresponding facades `Str::*` and `Arr::*`
- Had to change the migrations to use `bigint` for everything that points to the user-id (roles and permissions) / Rollback works fine for me now.
- I didn't drop Laravel 5.4 support in this PR as it needs much more changes to our testbench

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ View the Voyager Cheat Sheet: https://voyager-cheatsheet.ulties.com/
 
 <hr>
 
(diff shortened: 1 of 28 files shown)


fletch3555 (approved):

CyberPunkCodes:
Any idea when this will have a release for 5.8?

Just started a fresh project, needing an admin dashboard :) Though it can wait a couple of days if it's to be expected soon.

Thanks!

PR #3983, opened 2019-03-04
update menu children by setRelation() method

chanr1th:
The `$item->children = static::processItems($item->children);` doesn't work.
It still showing the menu item in sidebar, so in this change will use setRelation method instead.
 `$item->setRelation('children', static::processItems($item->children));`

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -120,7 +120,7 @@ class Menu extends Model
             }
 
             if ($item->children->count() > 0) {
-                $item->children = static::processItems($item->children);
+                $item->setRelation('children', static::processItems($item->children));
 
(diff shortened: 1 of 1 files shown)


emptynick:
Do you have some steps to reproduce what you are describing?

chanr1th:
I have another user role which doesn't have permission to access Tools->BREAD. I already unchecked permission "Browse BREAD", but the menu item of BREAD still showing in sidebar. By using the $item->children is not override the relationship in MenuItem, but it create new property name "children". That why I still see the sub-menu item.

emptynick:
Yep, reproducible and fixes the problem. Thanks!

emptynick (approved):

PR #3992, opened 2019-03-09
Hotfix menu doesnt collapse

FlickTheManual:
come from slack #help channel, thanks to @emptynick 

then I looked up all the vendore views file in order to matching up all the files, then it's ok.
I see that the menu has a vue component,  and so it required a json file now, sound's good :smiley: 
but i got a problem, and I've search on all the gitub and slack before talk, but didn't find a relatedmy menu doesn't work, the js debuger call me that event doesn't defined in this portion of code:

```javascript
$('.panel-collapse').on('hide.bs.collapse', function(e) {
        var target = $(event.target);
        if (!target.is('a')) {
            target = target.parent();
        }
        if (!target.hasClass('collapsed')) {
            return;
        }
        e.stopPropagation();
        e.preventDefault();
    });
```

in upper I wrote the source code come from `tcg/voyager/ressource/assets/js/app.js` vendor package.


emptynick:
Please don't commit the compiled assets as they are compiled automatically once its merged.
I reverted those changed.

emptynick (approved):

FlickTheManual:
ho my bad, I'll know for next time. Thanks for the advice. 

I thought that public assets were also to be published compiled

PR #4000, opened 2019-03-12
Fix Multiple Issues in Media Manager

suxur:
Issues Fixed:
- When creating folders in Firefox the name is "undefined"
- When moving files in Firefox the selected destination returns "undefined"
- Unable to use "command" or "meta" key for multi-selection of files
- Using "meta key" brought to surface another issue where you could select the same file / folder multiple times.


resources/views/media/manager.blade.php
```diff
@@ -293,7 +293,8 @@
 
                 <div class="modal-body">
                     <h4>{{ __('voyager::media.destination_folder') }}</h4>
-                    <select class="form-control">
+                    <select class="form-control" v-model="modals.move_files.destination">
+                        <option value="">Select Destination</option>
```
emptynick: Any specific reason for this option?
What happens if I use it?

emptynick (changes requested):

resources/views/media/manager.blade.php
suxur: Mainly for placeholder text. When using v-model on the select input we have a default value, currently set to `''` and if this option is not there the select input appears empty, this prevents that and gives the user instruction.

Currently, if used it tries to move the file or folder to the same location it already exists which returns an error "File already exists at path." If you're okay with keeping this as a "placeholder" option, I'd like to update this PR by adding a check the value before submitting the post request to move the file.

resources/views/media/manager.blade.php
```diff
@@ -293,7 +293,8 @@
 
                 <div class="modal-body">
                     <h4>{{ __('voyager::media.destination_folder') }}</h4>
-                    <select class="form-control">
+                    <select class="form-control" v-model="modals.move_files.destination">
+                        <option value="" disabled>Select Destination</option>
```
emptynick: This hardcoded string needs to be translated.
For simplicity I suggest 
```suggestion
                        <option value="" disabled>{{ __('voyager::media.destination_folder') }}</option>
```

emptynick (changes requested):

emptynick (approved):

PR #4004, opened 2019-03-14
Make assets-path absolute

### LINKED ISSUE #3994: Voyager can't find app.css, app.js, and images
## Version information

 - Laravel: latest
 - Voyager: latest
 - PHP: 7.2
 - Database: MySQL-5.6

## Description:
Voyager can't find app.css, app.js, and images. It tries to find it in this directory: 
http://projectname/admin/assets/css/app.css. But I have some previous projects with voyager and  there is another path: http://projectname/vendor/tcg/voyager/assets/js/app.css. Also in my current project there is no such a folder: domains\projectname\public\vendor\tcg\voyager\assets, but it should be there, because it is in my previous projects. I reinstall laravel project and voyager several times, but it didn't help. My env: APP_URL=http://shop.project

## Additional context
Please help me

emptynick:
Makes the assets-loading path absolute instead of relative to the controller.
Also, instead of loading `admin/assets/xyz.css` which makes Nginx think there should be a physical file, assets are now loaded as `admin/assets?path=xyz.css`

Fixes #3994 
Notice: The TinyMCE plugins need to be included directly in the JS file, so they are not necessary in `publishable/assets/js/plugins` anymore.
Because those are many many files I didn't delete them in this PR.
This needs to be done separately later


fletch3555 (dismissed):

innerdev:
There is another solution: https://github.com/the-control-group/voyager/issues/3994#issuecomment-473149434

May be just add this to documentation?

fletch3555 (approved):

PR #4006, opened 2019-03-14
Update adding-custom-formfields.md

ghost:


emptynick:
Why would you use a built-in view in a custom formfield?
The view is created directly below that.

PR #4009, opened 2019-03-15
Rollback not working when settings table has data

ElieSauveterre:
I have an error when doing a rollback if the settings table have some data.
```
  Illuminate\Database\QueryException  : SQLSTATE[22004]: Null value not allowed: 1138 Invalid use of NULL value (SQL: ALTER TABLE settings CHANGE value value TEXT NOT NULL COLLATE utf8mb4_unicode_ci)
```

### DIFF
--- a/migrations/2018_03_16_000000_make_settings_value_nullable.php
+++ b/migrations/2018_03_16_000000_make_settings_value_nullable.php
@@ -26,7 +26,7 @@ class MakeSettingsValueNullable extends Migration
     public function down()
     {
         Schema::table('settings', function (Blueprint $table) {
-            $table->text('value')->nullable(false)->change();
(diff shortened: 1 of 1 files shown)


migrations/2018_03_16_000000_make_settings_value_nullable.php
```diff
@@ -26,7 +26,7 @@ public function up()
     public function down()
     {
         Schema::table('settings', function (Blueprint $table) {
-            $table->text('value')->nullable(false)->change();
+            $table->text('value')->nullable()->change();
```
emptynick, line 29: This does not do anything because the field is nullable already at this point.
Also, the down method should exactly do what it already does, to revert the field being nullable

emptynick (changes requested):

PR #4017, opened 2019-03-20
Relationship edit-add

### LINKED ISSUE #4003: [BUG] If you untick edit for the relationships in BREAD, the row does not exists
## Version information

 - Laravel: v5.8.4
 - Voyager: v1.2.2
 - PHP: 7.2.15
 - Database: MYSQL Ver 14.14 Distrib 5.7.25, for Linux (x86_64)

## Description

If you untick edit for the relationships in BREAD, the row does not exist.
If "edit" is unticked adding a record isn't working either. (`Results could not be loaded`)

## Steps To Reproduce
1. Go to 'BREAD'
2.  Edit a model BREAD
3. Untick 'Edit' on a relation (I tried on a belongsTo reltionship type)
4. Try adding a record using that model BREAD and you will get : Results could not be loaded and an error in google chrome dev tools console.

## Expected behavior
With relation'edit' field unticked , adding a bread should work.

## Screenshots
Annotation 2019-03-12 130443
Annotation 2019-03-12 130401

## Additional context
As emptynick pointed on the slack #help channel, the problem seems to be on [this line](https://github.com/the-control-group/voyager/blob/1.2/src/Http/Controllers/VoyagerBaseController.php#L645)

> If you untick edit for the relationships, the row does not exist
> ```
> src/Http/Controllers/VoyagerBaseController.php:645
>         ```foreach ($dataType->editRows as $key => $row) {```
> the-control-group/voyager
> ```
> Just need to pass add or edit to that method somehow

grimxd:
Fixes https://github.com/the-control-group/voyager/issues/4003


fletch3555:
@grimxd, we need more information when opening pull requests.  If it's related to an open issue, a link to the issue is fine, but ideally you should describe what you're changing and the reasoning/justification behind the change

emptynick:
If you untick `edit` for a relationship and go to the add-page for that BREAD the relationship does not show any results, because the controller only looks for the editRows.
This PR changes this to be depending on the current action.

emptynick (approved):

PR #4027, opened 2019-03-24
return null instead of json_encode -- fixes issue with other column t…

ryanm352:
…ypes such as varchar, text for mysql < 5.7


resources/views/master.blade.php
```diff
@@ -43,6 +43,7 @@
     @if(!empty(config('voyager.additional_css')))<!-- Additional CSS -->
         @foreach(config('voyager.additional_css') as $css)<link rel="stylesheet" type="text/css" href="{{ asset($css) }}">@endforeach
     @endif
+	@stack('css')
```
fletch3555, line 46: This change is unrelated to the PullRequest title.  Please remove it

DaLooK:
Issue already fixed by pull request #4158

PR #4033, opened 2019-03-25
Several ? 4 ?

Epistol:
I mean, it's a strong word for 4 methods isn't it ?


emptynick:
Several, in my eyes, means more than 1

PR #4038, opened 2019-03-26
Fix the route problem after adding/updating BREAD.

alkhachatryan:
When we add new BREAD the new slug appends to route list. In new Laravel version we have to manually cache the list of routes to have it updated after modifications.

In our case, when we add new BREAD and it's routes append to the route list we have to cache the list again. Because of it we have an exception every time we create new BREAD (update old one slug).

For two cases (store and update) I call here Artisan command to optimize the config files.

### DIFF
--- a/src/Http/Controllers/VoyagerBreadController.php
+++ b/src/Http/Controllers/VoyagerBreadController.php
@@ -4,6 +4,7 @@ namespace TCG\Voyager\Http\Controllers;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


emptynick:
`optimize` is definitely the wrong command to use.
It will only generate the file if it doesn't already exists, so storing a BREAD once is fine, storing another BREAD won't do anything.
I'm not even a hundred percent sure that it optimizes something with routes.

Also, I dont see a reason in creating BREADs in a production environment. That should be dev-stuff.
And I didn't hear about this problem before.
Why don't you hook up to the `BreadChanged`/`BreadAdded` events and clear your route-cache there?

PR #4043, opened 2019-03-27
TranslationTableSeeder fixes

Braiani:
- With this PR I fix the error when installing Voyager with a locale different from `en`.
- Improved the way TranslationTableSeeder seeds the Translation contents. So, with this PR, Translation Seeder isn't hardcoded anymore.
- From now on, the seeds' dummy content use Laravel lang files to fill the categories and pages tables.

***Important***: As far as I changed some dummy seeder content, it's needed to run:
 ```php
php artisan vendor:publish --force --tag="dummy_seeds"
```

### CODE REVIEW

**Comment by emptynick** [comment]
You need to update the multilanguage-test to use the translations as well:

https://github.com/the-control-group/voyager/blob/9101761b2f821d239c1e05015a52d5982c53a5cd/tests/MultilingualTest.php#L243-L266

Also, I'd like to see a test where Voyager is installed with a different local.
Something like
```
$this->app->setLocale('pt_br');
$this->install();
```
should do the trick

File: publishable/lang/en/seeders.php [hunk 1]
```diff
@@ -1,6 +1,16 @@
 <?php
 
 return [
+    'categories' => [
```
emptynick: I think I'd prefer an array with "objects" of categories. Like
```
[
    [
        'name' => '...',
        'slug' => '...'
    ],
    [
        'name' => '...',
        'slug' => '...'
    ]
]
```

So one could simply change the language-files to seed another category.
This of course applies to posts/pages as well

emptynick (approved):

publishable/lang/en/seeders.php
emptynick, line 4: I think I'd prefer an array with "objects" of categories. Like
```
[
    [
        'name' => '...',
        'slug' => '...'
    ],
    [
        'name' => '...',
        'slug' => '...'
    ]
]
```

So one could simply change the language-files to seed another category.
This of course applies to posts/pages as well

emptynick (changes requested):

PR #4049, opened 2019-03-29
Remove empty translations from $translations array

### LINKED ISSUE #4048: [Feature] Remove empty translate inputs if none provided
- Laravel: v5.7.28
 - Voyager: v1.2.2
 - PHP: 7.2
 - Database: mysql  Ver 15.1 Distrib 10.3.13-MariaDB, for osx10.14 (x86_64) using readline 5.1

## Description of problem
Let's say i have a blog Post model which uses voyager Translatable trait.  I set 2 languages as "en" and "zh-tw". "en" is default language. In model i set "name" and "body" as translatable fields.

If i create new post with "en" only, without providing any input data for the "zh-tw", by default voyager bread will add empty values for each translatable fields for the post model. 
So i will end up with empty name and body fields in translations table. 

In view when i use getTranslatedAttributes('name') or getTranslatedAttributes('body') method which returns app current locale value of the attribute and fallback to app default locale value, i end up with empty "name" and "body" because these entries are exists in translations table. So fallback locale can't be used. Which makes me think this could be a bug. I think voyager should not query to DB if no any translated inputs provided for the other languages. But by default voyager sets empty strings for the translatable attributes. And in the server there is no any check if value is empty or not.

## Proposed solution
i dig in voyager code and found a solution. I need to override TCG\Voyager\Traits/Translatable trait to achieve this. 

I could check translatable attribute values if they are empty or not and remove them from array if they are empty by this:

```
 public function setAttributeTranslations($attribute, array $translations, $save = false)
    {
        $response = [];

        if (!$this->relationLoaded('translations')) {
            $this->load('translations');
        }
       
        //Changes here
        foreach ($translations as $key => $value) {
           if(empty($value)){
                unset($translations[$key]);
           }
        }
        //To here

        $default = config('voyager.multilingual.default', 'en');
        $locales = config('voyager.multilingual.locales', [$default]);

        foreach ($locales as $locale) {
            if (!isset($translations[$locale])) {
                continue;
            }

            if ($locale == $default) {
                $this->$attribute = $translations[$locale];
                continue;
            }

            $tranlator = $this->translate($locale, false);
            $tranlator->$attribute = $translations[$locale];

            if ($save) {
                $tranlator->save();
            }

            $response[] = $tranlator;
        }

        return $response;
    }
```
As you can see, there i check posted translatable attributes and remove them if they are empty. By this, voyager getTranslatedAttributes method can fallback to default locale since there will not be any entry in translations table.

## Alternatives considered
I could also set the translatable attributes to default language value by 
```
foreach ($translations as $key => $value) {
           if(empty($value)){
                $translations[$key] = $translations[config('voyager.multilingual.default', 'en')];
           }
}
```
This way there will be english values saved to translations table in case first solution cause any issue.

## Additional context
I applied this logic by making my own Translatable trait same as voyager's but with the modifications and using them in my models. Everything seems fine for now. I would like to know your thoughts. Thanks.

ghost:
Check if $translations array has empty locale value and remove them if any. Fixes #4048


emptynick:
[array_filter](https://php.net/manual/en/function.array-filter.php) only takes one additional lines and does the same.

MrCrayon (changes requested):
I think it would be better to change instead https://github.com/the-control-group/voyager/blob/9101761b2f821d239c1e05015a52d5982c53a5cd/src/Traits/Translatable.php#L227
```diff
-           if (!isset($translations[$locale])) {
+           if (empty($translations[$locale])) {
```

emptynick:
I think `array_filter` is fine.
Instead he can completely remove the condition you mentioned.

MrCrayon:
@emptynick I was thinking that you were right but I checked and array_filter will work only if we leave also:
https://github.com/the-control-group/voyager/blob/9101761b2f821d239c1e05015a52d5982c53a5cd/src/Traits/Translatable.php#L227
that's the check that is actually removing the addition.

Otherwise we will still have:
https://github.com/the-control-group/voyager/blob/9101761b2f821d239c1e05015a52d5982c53a5cd/src/Traits/Translatable.php#L237
and at that point that's going to be undefined.

nemrutco:
@MrCrayon ```if (!isset($translations[$locale])) { ``` will return false because $translations[$locale] will have empty string by default because of the way implemented in voyager frontend. So  '' string is considered empty but isset will return true. array_filter checks empty strings and removes  those from array so i guess it's better way to implement. 



```
$var = '';

// This will evaluate to TRUE so the text will be printed.
if (isset($var)) {
    echo "This var is set so I will print.";
}
```

I have already to simple way using array_filter.

ghost (commented):
Used array_filter to make it more simple.

MrCrayon:
@nemrutco yes I know how it works.
I was replying to @emptynick that was suggesting to leave filter_array and remove the check with !isset() telling him that if we remove that part it's not going to work.

I think the only two options are add the filter_array as you did or just change !isset with empty, I think the latter is better because it has the same result and only changes one word.

nemrutco:
> @nemrutco yes I know how it works.
> I was replying to @emptynick that was suggesting to leave filter_array and remove the check with !isset() telling him that if we remove that part it's not going to work.
> 
> I think the only two options are add the filter_array as you did or just change !isset with empty, I think the latter is better because it has the same result and only changes one word.

right, i missed that part. I think checking  $translations array should be re considered  because i face some weird behaviors like at creating of post, there will be no any translations entry to the table, but if you update existing post without providing translation, english values will be applied to the translations table. So my implementing is only for  creation of the post. Because after creation of the post, fallback language will be default value for all other languages, so other language inputs will not be an empty string at update.

emptynick:
Guess we can agree on using @MrCrayon solution.

reageek:
Was experiencing this issue as well. 

For those of us that have database that do already have the blank translator fields rather than having to hunt those entries down and remove them, you can apply a simple fix to line 183 of Translatable.php ... that fixes the results from using getTranslatedAttribute so that it actually falls back correctly. 

> if ($localeTranslation && $localeTranslation->getAttribute('value') != '') {

in place of 

> if ($localeTranslation)

It might be an idea for this pr to apply the above change as well... so that upgrades don't replicate the issue.

emptynick:
@MrCrayon thats what you suggested right?

MrCrayon:
@emptynick yes

emptynick (approved):

PR #4051, opened 2019-03-29
Customize admin favicon using admin.icon_image setting

olegts239:
Adding possibility to change favicon using admin.icon_image setting


MrCrayon (changes requested):
I think in this case a one line solution might be enough:
```diff
-<link rel="shortcut icon" href="{{ voyager_asset('images/logo-icon.png') }}" type="image/x-icon">
+<link rel="shortcut icon" href="{{ Voyager::setting('admin.icon_image') ? Voyager::image(Voyager::setting('admin.icon_image')) : voyager_asset('images/logo-icon.png') }}" type="image/x-icon">
```
But if you want to keep the if else for clarity, to not repeat html, I would do something like this:
```blade
@php
if (Voyager::setting('admin.icon_image')) {
    $admin_favicon = Voyager::image(Voyager::setting('admin.icon_image'));
} else {
    $admin_favicon = voyager_asset('images/logo-icon.png');
}
@endphp
<link rel="shortcut icon" href="{{ $admin_favicon }}" type="image/x-icon">
```

olegts239:
@MrCrayon thanks for your comment, i used the same code in my commit as [there](https://github.com/the-control-group/voyager/blob/e3ab444f365eaf6ef140004737a4fe6aa24c6b3e/resources/views/master.blade.php#L58-L63) for unified code styling

MrCrayon (dismissed):
@olegtsoy I saw about your choice for code styling, it makes sense.

Works for me.

emptynick:
I'm not a hundred percent sure that this is compatible with
https://github.com/the-control-group/voyager/blob/317ce0f2014b08ea4f0f3ead0f973f1c952f642c/resources/views/login.blade.php#L50-L55
What happens if you upload a .ico file? Also, I think some browser don't directly accept a PNG as a favicon (IE for example)

MrCrayon:
According to https://en.wikipedia.org/wiki/Favicon PNG and ICO are supported on all modern browsers, including IE11.
If someone is still using older IE versions not seeing bookmark icon might be the last of their problems :laughing:

emptynick:
Its more about ` type="image/x-icon"`.
Uploading a PNG might (not sure) result in errors/notices in the console.
Can't we just change it to `image/png` and only use `image/x-icon` for the default .ico file?

resources/views/master.blade.php
```diff
@@ -10,7 +10,12 @@
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,700" rel="stylesheet">
 
     <!-- Favicon -->
-    <link rel="shortcut icon" href="{{ voyager_asset('images/logo-icon.png') }}" type="image/x-icon">
+    <?php $admin_favicon = Voyager::setting('admin.icon_image', ''); ?>
+    @if($admin_favicon == '')
+        <link rel="shortcut icon" href="{{ voyager_asset('images/logo-icon.png') }}" type="image/x-icon">
```
fletch3555: I agree with @emptynick.  Please change to `type="image/png"`

fletch3555 (changes requested):

emptynick (approved):

emptynick (approved):

PR #4057, opened 2019-04-04
Bread translatable

MrCrayon:
This pull request make use of the already available translation of display_name_singular and display_name_plural in Voyager interface and adds translatable property for display_name in DataRow so that labels in forms can also be easily translated.


resources/views/tools/bread/edit-add.blade.php
```diff
@@ -2,6 +2,10 @@
 
 @if (isset($dataType->id))
     @section('page_title', __('voyager::bread.edit_bread_for_table', ['table' => $dataType->name]))
+    @php
+        $display_name = $dataType->display_name_singular;
```
fletch3555: Not sure if a mistake or intentional, but shouldn't this be wrapped in a call to `getTranslatedAttribute()` as well?

fletch3555 (dismissed):

resources/views/tools/bread/edit-add.blade.php
MrCrayon: I remember in one place I left it because I thought it made more sense to leave it untranslated but I'll have to check it tomorrow to see what was the reason and if it still make sense.

MrCrayon: I checked again and the only place where $display_name is actually used in that file is in the breadcrumb on top because the other two places are overridden by the i18n interaction.
Anyway I remembered why I left it out but the reasoning doesn't look that strong anymore, so I think I can change that too.
Thanks for the review.

emptynick:
Are you done here?
Can you fix the conflicts or do you want me to do it?

MrCrayon:
Done

resources/views/tools/bread/edit-add.blade.php
```diff
@@ -96,7 +100,7 @@
                                            name="display_name_singular"
                                            id="display_name_singular"
                                            placeholder="{{ __('voyager::bread.display_name_singular') }}"
-                                           value="{{ $dataType->display_name_singular ?? $display_name }}">
+                                           value="{{ $display_name }}">
```
emptynick, line 103: This var is defined in the controller and is the first value when adding a BREAD.
I think you should check if adding a BREAD and if not, use the translated attribute.
But I'd prefer to have this in the controller, the definitions are already there.

Edit: Sorry, just saw this is inside the edit-condition.
Nonetheless I think that the "logic" should be in the controller.

Edit 2: Nevermind, this would create a lot overhead on how the code currently works.

resources/views/tools/bread/edit-add.blade.php
```diff
@@ -37,7 +41,7 @@
     <li class="active">
         @if(isset($dataType->id))
         <a href="{{ route('voyager.bread.edit', $table) }}">
-            {{ $dataType->display_name_singular }}
+            {{ $display_name }}
```
emptynick, line 44: Same as above

resources/views/tools/bread/edit-add.blade.php
```diff
@@ -111,7 +115,7 @@
                                            name="display_name_plural"
                                            id="display_name_plural"
                                            placeholder="{{ __('voyager::bread.display_name_plural') }}"
-                                           value="{{ $dataType->display_name_plural ?? $display_name_plural }}">
+                                           value="{{ $display_name_plural }}">
```
emptynick, line 118: Same as above

emptynick (changes requested):

emptynick (approved):

PR #4059, opened 2019-04-04
French translations

MrCrayon:
Syncing with EN:

- Added some missing fields.
- Removed some obsolete.
- Translated some.


cotiga:
There is still a lot of new untranslated lines.
Thank you not to "commit" until everything is translated!

Il reste beaucoup de nouvelles lignes non traduites.
Merci de ne pas "commiter", tant que tout n'est pas traduit !

MrCrayon:
I don't speak french so I can't translate, you are welcome to submit translations.

I submitted the strings, even if they are not translated, to avoid the problem of showing:
voyager::generic.remember_me
when english is not in the list of languages chosen for the platform and the fallback language does not contain all strings.

emptynick:
I wrote a little [php script](https://cschweppe.de/translations.php?locale=fr) comparing all locales to english.
This will give you a good overview on whats missing/whats unnecessary.
Remove `?locale=fr` to show all locales or change it to whatever you like.
`Syntax error` means there's something strange. In frenchs case its this:
https://github.com/the-control-group/voyager/blob/5e175637fca405577f3b7a06ee029ec339686406/publishable/lang/fr/compass.php#L4-L8

MrCrayon:
@emptynick oh yes I saw that but I didn't notice the quotes so I thought the translator for some reason wanted to keep the original text and I didn't remove it.

emptynick:
Are you done here?

MrCrayon:
Yes, I removed that problem in compass and all fields should be synced with english, I can't really do anything else here.

emptynick (approved):

PR #4083, opened 2019-04-13
Clear menu-cache when deleting items

### LINKED ISSUE #4071: Failed to Remove Menu Item from Menu Builder
## Version information
 - Laravel: v5.8
 - Voyager: v1.2
 - PHP: v7.1.3
 - Database: MySQL v5.7.24

## Description
I Removed Menu Item from Menu builder. It is also deleted from Database but the View is Still Rendering that Menu Item and Menu Builder Shows Deleted Item Entry. When I tried to delete it again, it throws 404 error.

## Steps To Reproduce
1. Make Clean Install of Laravel and Install Voyeger.
2. Go to the Database and Create a Table.
3. Go to BREAD Builder and Generate BREAD for the Recently added table. It will automatically add menu item entry.
4. Go to Menu Builder and Delete that Generated Menu Item and Refresh the Page.
5. You will see, that menu item is still rendering.

## Expected behavior
As it's already deleted from menu builder, it should not be rendered in Side Bar menu.

## Additional context
I Tried clearing View Cache by running **php artisan view:clear** but the problem is still there.

emptynick:
When a menu-item is destroyed, the `save` method is not called, thus the menu-cache is not cleared.
This PR adds cache-clearing when an item/menu is deleted

Fixes #4071


AdrienPoupa:
Instead of overriding Eloquent's save/destroy function, you could use events: https://laravel.com/docs/5.8/eloquent#events

emptynick:
But this would introduce a whole new file to just call one method.
I prefer this way instead.

MrCrayon:
@emptynick please reconsider your approach, using events doesn't always imply creating an observer, you can just add it in the model boot method:
```php
    public static function boot()
    {
        parent::boot();

        static::deleted(function ($model) {
           $model->removeMenuFromCache();
        });

        static::saved(function ($model) {
            $model->removeMenuFromCache();
        });
    }
```

Also I think your destroy method will only remove Cache from the first Menu in the **really** remote case someone is going to delete more than one menu at the same time.

emptynick:
Those static-anonymous functions have a lot of disadvantages.
They are static, so calling `$model->removeMenuFromCache();` is not going to work if `removeMenuFromCache` isn't static as well.
If you make it static, you can't access properties without using hackish get/set method built-in to Eloquent and so on.

I made it observed by another class now.

MrCrayon:
> They are static, so calling `$model->removeMenuFromCache();` is not going to work if `removeMenuFromCache` isn't static as well.

$model is passed as parameter, it works for me and it's the way I always used it and saw in examples.

> I made it observed by another class now.

good :+1:

emptynick:
> $model is passed as parameter, it works for me and it's the way I always used it and saw in examples.

Yes, but for menu-items you need to get the parent menu by calling a not-static method `menu()` and so on.

MrCrayon:
```php
$model->menu->removeMenuFromCache();
```
also works from MenuItem

fletch3555:
My preference is leaning toward the events approach.

emptynick:
Done

fletch3555 (approved):

Issue #4086, opened 2019-04-16
[Feature] Different menu for roles.


wvffle:
## Description of problem

Custom admin menu for roles.

## Proposed solution
Connect roles to menu types in the builder.

## Additional context
I've tried to check for users role and display another menu in `sidebar.blade.php` like this:
```blade
        <div id="adminmenu">
          @if (app('VoyagerAuth')->user()->hasRole('reception'))
            <admin-menu :items="{{ menu('reception', '_json') }}"></admin-menu>
          @else
            <admin-menu :items="{{ menu('admin', '_json') }}"></admin-menu>
          @endif
        </div>
```

But I ran into a problem where the links were not working. I had to remove the check for `$menuName` in `src/Models/Menu.php` as follows and it works now but with any update I'd have to do it again :<

```php
        if ($type == '_json') {
            $items = static::processItems($items);
        }
```

fletch3555:
The full issue template is required, including version number.

This has been requested numerous times, and almost always ends in an answer to use permissions and a single menu since permissions are already tied to roles

ghost:
Please help me
I am having issue with roles how can I prevent admin from being edit self or to make any other user super admin. Admin should not even able to delete the super admin user.  I am using voyager 1.2 with laravel 5.8

emptynick:
This is not the place to ask for this.
See https://github.com/the-control-group/voyager/issues/3852

danielorkae:
i've added a pull request to solve this feat.
see: https://github.com/the-control-group/voyager/pull/4107

emptynick:
Closing.
Please see the discussion on why here: https://github.com/the-control-group/voyager/pull/4107

nuhu-ibrahim:
Follow the link [here](https://nuhuibrahim.com/blog/using-multiple-menu-builder-in-laravel-voyager) to see my solution that is permanent to this problem.

PR #4107, opened 2019-04-29
Custom menu by user role

danielorkae:
i added custom menu support based on the user role.
for that, i created a helper that checks if the menu exists, so i put the condition that if there is a menu with the authenticated user's role name, then use this menu.

helper name: `menu_exists`

### DIFF
--- a/resources/views/dashboard/sidebar.blade.php
+++ b/resources/views/dashboard/sidebar.blade.php
@@ -30,7 +30,7 @@
 
         </div>
         <div id="adminmenu">
-            <admin-menu :items="{{ menu('admin', '_json') }}"></admin-menu>
+            <admin-menu :items="{{ menu(menu_exists(Auth::user()->role) ? Auth::user()->role : 'admin', '_json') }}"></admin-menu>
         </div>
     </nav>
 </div>
--- a/src/Helpers/helpers.php
+++ b/src/Helpers/helpers.php
@@ -14,6 +14,14 @@ if (!function_exists('menu')) {
     }
 }
 
+if (!function_exists('menu_exists')) {
+    function menu_exists($menuName)
+    {
+        return TCG\Voyager\Facades\Voyager::model('Menu')
+            ->where('name', $menuName)->count() > 0;
+    }
+}
+
 if (!function_exists('voyager_asset')) {
     function voyager_asset($path, $secure = null)
     {


MrCrayon:
@danielorkae I'm not against it, for what it matter, but I think that whoever decides or not to merge this pull request will be able to take a better decision if you could provide a case scenario in which having a different menu per role could be a better solution than using permissions to have different menu items showing per role as it was suggested many times in the past.

Also in case it gets merged I think it should be made clear in documentation that menu is based on "main" role, since a user can have more than one role.

fletch3555:
I would also like to hear the answer to @MrCrayon's question.  What's the benefit to this over just assigning permissions to your roles?

danielorkae:
I made this PR based on this issue: https://github.com/the-control-group/voyager/issues/4086

I've also had cases where the permissions-only menu did not work for what I'd like to do.
for example, I've created an event registration management website. in it, the inscriptions were grouped in the read of the events resource, and it was necessary that the permission for navigation was activated. so the "inscriptions" item appeared in the menu, but it should not. my solution was to create a specific menu for the "organizer". implemented as in PR.

as this is a solution that depends on whether there is a menu with the same name as the user role, I believe that the two menu modes may coexist in the voyager.
is something I already implement as a rule in my systems, and would be happy to see in the official version.

resources/views/dashboard/sidebar.blade.php
```diff
@@ -30,7 +30,7 @@
 
         </div>
         <div id="adminmenu">
-            <admin-menu :items="{{ menu('admin', '_json') }}"></admin-menu>
+            <admin-menu :items="{{ menu(menu_exists(Auth::user()->role) ? Auth::user()->role : 'admin', '_json') }}"></admin-menu>
```
fara0: Auth::user()->role->name

vkiranmaniya:
Custom menu  According to the different role is required in some use cases like, Administrator has also the capability to add more Admin/Staff/Management Person with Specific Rules and Authorities. I'm developing this with Voyager and I would like to hear you about it.

fletch3555:
Not planning to add this right now.  You can override the sidebar view and call whatever menu you'd like if needed.

neapolis79:
> Not planning to add this right now. You can override the sidebar view and call whatever menu you'd like if needed.

Excuse me if I resume this closed thread, but personally I think that the ability to manage permissions for different menus for each role is useful and makes sense. Managing admin menu permissions makes sense only if you consider the sidebar of the admin area, but if for example for any site I create a navigation menu to use in the header (as shown in the menu tutorial video) and I have more than one role, I can't manage permissions for this new menu, for any role, not even for the admin one.
The way things are now, I can create more than one menu, but I can completely manage only one. IMHO

fletch3555:
Of course you can. You'll likely have a custom rendering for your menu, so just add the auth checks in there.  Voyager already lets you create/manage permissions, and assign them to roles... so the only part that's missing is the check for that permission.  Since you're talking about rendering these menus outside of Voyager, that specific bit of functionality is outside the scope of Voyager and well within your ability to implement.

neapolis79:
i appreciate it a lot> Of course you can. You'll likely have a custom rendering for your menu, so just add the auth checks in there. Voyager already lets you create/manage permissions, and assign them to roles... so the only part that's missing is the check for that permission. Since you're talking about rendering these menus outside of Voyager, that specific bit of functionality is outside the scope of Voyager and well within your ability to implement.

Yes i'll do for sure because i need in one of my projects, but only to understand and to fullify my curiosity, if rendering menu outside voyager is out of scope, and inside voyager you can only fully manage the admin menu...i think that i'm missing the use case for creating additional menu managing only order and not permission...

Anyway, thank you for your answer and your effort.

PR #4108, opened 2019-04-30
Added sorting to multiple images field

aldozorzi:
Hello there,
I added drag and drop sort functionality to multiple images field, I think it can be useful to other developers.
I used jquery nestable plugin (already used for menu builder) to avoid other inclusions, but I used new css classes to not interfere with other scripts.
I added a little route and VoyagerMediaController::sort method to save info in DB.
I had to edit a bit of CSS, so publishable new files were compiled, but all edits were done in sass sources.

I hope I will find my functionality on next releases ;)

Bye!


fletch3555 (changes requested):
Please don't commit app.css and app.js.  We have a script that will automatically compile them and recommit after a merge.  Once you've reverted that, we can review this

fletch3555:
You deleted app.css and app.js.  They need to be untouched (i.e. match the 1.2 branch).  Deletion is still a change to the file, and still causes merge conflicts

emptynick:
I think the media-picker supersedes image/multiple images and files in the future and already supports drag/drop.
I'm going to close this. Please use the media-picker to use drag-drop.

bonucci:
> I think the media-picker supersedes image/multiple images and files in the future and already supports drag/drop.
> I'm going to close this. Please use the media-picker to use drag-drop.

The problem is that shares same folder, for example if i have many posts, i have images shared between them that haves nothing to do with it. Of course that i could create a folder, but i shouldnt have to do this work, the ideal scenario would be the images not event being showed in the container even if them already been uploaded. This is whats missing in media-picker to make it complete in all kind of scenarios.

MrCrayon:
I agree, also, if I'm not mistaken, to use media picker you need browse_media permission that gives you access to all files.

PR #4113, opened 2019-05-01
Fixes #4095 - Manage deletion of uploaded files

### LINKED ISSUE #4095: [BUG] Can't delete uploaded files from BREAD form without browse_media permission
## Version information

 - Laravel: v5.8.3
 - Voyager: v1.2.3
 - PHP: 7.2
 - Database: MariaDB 10.3.12

## Description
Can't delete uploaded files from BREAD form without browse_media permission

## Steps To Reproduce
1. Install Voyager
2. Create a BREAD with one image or file field (Pages can be used if voyager has been installed with-dummy)
2. Remove browse_media permission from current role
3. Create an entry for BREAD with an image or file
4. Edit the entry and try to remove image or file

## Expected behavior
The image or file should be cleared from the entry and removed from file system.

## Screenshots
Schermata da 2019-04-22 10-53-01

## Additional context
This is because voyager uses VoyagerMediaController to remove attachments for all BREAD so right now the only way to let a user delete a previously uploaded image/files is grant that user browse_media permission that gives complete access to all media gallery.
I think this should be handled by VoyagerBaseController.

MrCrayon:
Transfer remove method from VoyagerMediaController to VoyagerBaseController
Changed permission check, now only edit permission for current BREAD is checked (before it checked browse_media and for some reason delete permission for current BREAD).
remove file and eventual thumbnails from file system

Fixes #4095


fletch3555 (approved):

emptynick:
I think `remove` is too generic and might cause confusion, because it removes media/files (and not the entry for example).
Maybe we should name it `remove_media` or `remove_file`.
Thoughts @fletch3555 ?

fletch3555:
There's nothing saying the route name needs to follow a certain pattern... What about `voyager.<datatype>.media.remove`?

resources/views/bread/edit-add.blade.php
```diff
@@ -183,7 +183,7 @@ function deleteHandler(tag, isMulti) {
             $('.form-group').on('click', '.remove-single-file', deleteHandler('a', false));
 
             $('#confirm_delete').on('click', function(){
-                $.post('{{ route('voyager.media.remove') }}', params, function (response) {
+                $.post('{{ route('voyager.'.$dataType->slug.'.remove') }}', params, function (response) {
```
fletch3555: Route name reads like you're removing the `$slug`, not media.  Please rename routes per my previous comment (`voyager.[$slug].media.remove`, or similar)

fletch3555 (dismissed):

src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -466,6 +467,138 @@ public function restore(Request $request, $id)
...
+            $model = app($dataType->model_name);
+            $data = $model::find([$id])->first();
+
+            // Check if field exists
+            if (!isset($data->{$field})) {
+                throw new Exception(__('voyager::generic.field_does_not_exist'), 400);
+            }
+
+            // Check permission
+            $this->authorize('edit', $data);
```
emptynick, line 529: Why is this `edit`?

MrCrayon: Because the only place where we can delete uploaded files is when we are changing BREAD data and the only permission the user needs is edit.

emptynick: But you can delete things also when adding right?

MrCrayon:
To be honest I'm not 100% sure but untill you are not done with the add part there shouldn't be anything to delete either in database or in filesystem.
I'll check tomorrow anyway I'm not in front of a PC right now.

src/Http/Controllers/VoyagerBaseController.php
MrCrayon: To be honest I'm not 100% sure but untill you are not done with the add part there shouldn't be anything to delete either in database or in filesystem.
I'll check tomorrow anyway I'm not in front of a PC right now.

MrCrayon: I can confirm the delete button will only appear on edit and there wouldn't be anything to delete on add.

emptynick (approved):

PR #4132, opened 2019-05-13
Update file.blade.php

tjebe:
When no file uploaded for file type, viewing or editing resulted in invalid download link.
json_decode of "[]" returns false, so file.blade.php does not properly interpret the fact that there is no file.


DaLooK (commented):
This issue is already fixed in the 1.2 branch but not released yet (at time of post)

emptynick:
True. Closing

Issue #4115, opened 2019-05-02
[BUG] glyphicons-halflings font not loadinf


crofton77:
## Version information

 - Laravel: v5.8.15
 - Voyager: v1.2.3

## Description
When I added a description to a field I get 
`GET http://voice.test/fonts/bootstrap/glyphicons-halflings-regular.woff2 net::ERR_ABORTED 404 (Not Found)`
`GET http://voice.test/fonts/bootstrap/glyphicons-halflings-regular.woff net::ERR_ABORTED 404 (Not Found)`

with the CSS loading from the 'http://voice.test/admin/assets?path=css%2Fapp.css' path
`
@font-face {
    font-family: Glyphicons Halflings;
    src: url(../fonts/bootstrap/glyphicons-halflings-regular.eot);
    src: url(../fonts/bootstrap/glyphicons-halflings-regular.eot?#iefix) format("embedded-opentype"),
        url(../fonts/bootstrap/glyphicons-halflings-regular.woff2) format("woff2"),
        url(../fonts/bootstrap/glyphicons-halflings-regular.woff) format("woff"),
        url(../fonts/bootstrap/glyphicons-halflings-regular.ttf) format("truetype"),
        url(../fonts/bootstrap/glyphicons-halflings-regular.svg#glyphicons_halflingsregular) format("svg")
}
`

## Steps To Reproduce
1. Add a description to a field
2. got to that section and click 'Add New'
4. See description icon not loading

## Expected behavior
The CSS path correctly linked to load the font files and description icon showing correctly

MrCrayon:
As a temporary solution you can copy fonts folder from /vendor/tcg/voyager/publishable/assets/.

For a definitive solution I think url should be changed in app.css with assets?path=fonts/bootstrap/etc..

winex01:
@MrCrayon thanks.

xXcoronaXx:
I have a similar issue after upgrading to Voyager 1.2 and Laravel 5.6.

<img width="467" alt="Captura de pantalla 2019-05-16 a las 11 08 59" src="https://user-images.githubusercontent.com/3109179/57842345-921e3e00-77cc-11e9-84d0-67122aa114fa.png">
<img width="357" alt="Captura de pantalla 2019-05-16 a las 11 08 51" src="https://user-images.githubusercontent.com/3109179/57842346-92b6d480-77cc-11e9-897e-eb42740a5705.png">

I dont't if could be the same or not.

fletch3555:
[...]

PR #4143, opened 2019-05-17
the-control-group/voyager#4115 Rewrite .css files on-the-go

DaLooK:
* There was an issue where the css file was trying to load a relative path. This replaces relative paths (only on one level) to use the "/assets?path=xxx" url scheme


MrCrayon:
I think it's better to solve the problem when css is built #4146

DaLooK:
Closing in favor of @MrCrayon #4146

PR #4151, opened 2019-05-19
Show checkboxes column if mass action is displayed

### LINKED ISSUE #4147: [BUG] checkbox column for mass action not shown without delete permission
## Version information

 - Laravel: v5.8.15
 - Voyager: v1.2.3
 - PHP: 7.3
 - Database: 10.3.12-MariaDB

## Description

Checkbox column to select rcords for mass action is not showing if user doesn't have delete permission.

## Steps To Reproduce
1. Install Voyager with dummy or create your own data
2. Create mass action for a Posts (or something else)
3. Remove delete permission for Posts
4. Browse Posts

## Expected behavior
Checkbox column should be shown otherwise there is no way to select records for mass action.

## Screenshots
Schermata da 2019-05-18 07-39-07

## Additional context
In bread/browse.blade.php only delete permission is checked to show checkbox column

MrCrayon:
This was not as easy to fix as I was expecting to be.

bread.browse didn't know if mass action was going to be displayed or not because check was done inside bread.partials.actions and I don't think there is an easy way to pass data back from an @include also policy for action (not mass action) might depend on record data so we can't check once for all.

What I did was to move shouldActionDisplayOnDataType check to the controller VoyagerBaseController that is now passing to view an array of actions objects, in bread.partials.actions stays the policy check that might depend on record data.

Fixes #4147


resources/views/bread/browse.blade.php
```diff
@@ -13,6 +13,7 @@
             </a>
         @endcan
         @can('delete', app($dataType->model_name))
+            @php $showCheckboxColumn = true @endphp
```
fletch3555, line 17: seems weird setting a value of `true` then using `@isset()` below.  Perhaps give it a default value of false then juse `@if()` it instead....? Doesn't really save any code, but I think it reads cleaner.  Thoughts?

MrCrayon: I agree it's going to be more readable and it only adds one line of code.
I'll submit the changes asap.

emptynick (approved):

PR #4154, opened 2019-05-21
Check if User table has role_id before add it !

hirwaf:


hirwaf:
Close due to CI errors

hirwaf:
Close because of CI error

Issue #4020, opened 2019-03-21
Voyager "File" types get deleted for "text" sql type 


ryanm352:
## Version information
 - Laravel: v5.7
 - Voyager: v1.2.2
 - PHP: 7.1.3
- MySQL: 5.6.41-84.1
 - Database: [type] [version] (e.g. MySQL 8.0)

## Description:
Creating BREAD and using SQL type "text" with a voyager "file" type causes an issue where the previously saved upload file will be erased if any edits on the data are saved. This causes issues with mysql < 5.7 as json types don't exist (gatorhost servers)

To re-create:
--------------
Define a database table, with a columns:
"title" as "varchar"
"brochure" as "text" (json type not avail in mysql < 5.7) (gatorhost)
Add BREAD and define the BREAD type "brochure" as "File". Save.

Add an item to your new bread including a file.
Save.
Re-edit bread item, edit "title" only. Save.
Previously saved file updates as "[]"

## Fix:

``
vendor/tcg/voyager/src/Http/Controllers/ContentTypes/File.php
``

line 16: **return null instead of json_encode([]);**
``
if (!$this->request->hasFile($this->row->field)) {
            return null;
}
``

cotiga:
Same probleme on Laravel 5.8 and Voyager 1.2.2
Thanks for your code, work fine ;)

fletch3555:
Fixes should be submitted as pull requests, or with an explicit request that someone else submit the PR if you don't feel experienced enough to do so yourself.  Otherwise we'll just wait for you to submit it and it'lll get lost/forgotten about.

cotiga:
PR #4156
thank you ryanm352 again

PR #4156, opened 2019-05-22
Update File.php

cotiga:
Fix issue #4020
Thanks to ryanm352


DaLooK:
Added extra checks to better handle empty file fields without setting them to null. To pass the unit tests.
PR #4158

cotiga:
I updated the file that posed the problem #4020
For DaLook's changes, I don't tested.
Thank you

fletch3555:
#4158 is more comprehensive.  Closing this in favor of that one.

PR #4161, opened 2019-05-24
Html tooltips for fields description

trippo:
Add possibility to have html tooltips for fields description on BREAD


emptynick (approved):
Can't hurt

fletch3555 (approved):

PR #4165, opened 2019-05-25
Fix the row deletion if the field json is null

SAMIBETTAYEB:
When the file field contains a malformed json, It will throw an exception, so we need to check if the json_decode return is not null before supplying a null as a foreach parameter.

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -486,7 +486,7 @@ class VoyagerBaseController extends Controller
 
         // Delete Files
         foreach ($dataType->deleteRows->where('type', 'file') as $row) {
-            if (isset($data->{$row->field})) {
+            if (isset($data->{$row->field}) AND !is_null(json_decode($data->{$row->field}))) {
                 foreach (json_decode($data->{$row->field}) as $file) {
                     $this->deleteFileIfExists($file->download_link);
                 }


fletch3555:
According to the [php docs for isset](https://www.php.net/manual/en/function.isset.php):

> isset() will return FALSE when checking a variable that has been assigned to NULL

Now, from the [php docs on json_decode](https://www.php.net/manual/en/function.json-decode.php):

> NULL is returned if the json cannot be decoded or if the encoded data is deeper than the recursion limit.

So, from what I've read, the ONLY case this change will catch, is that of invalid JSON data being passed to it, which shouldn't happen since the data is previously set by `json_encode()`

SAMIBETTAYEB:
Yes, that's the case (malformed json).

SAMIBETTAYEB:
Yes, that's the case, for malformed json.

fletch3555:
Okay, so why do you have malformed json...? Data being written to that field is run through json_encode() by Voyager, so it should never be malformed

SAMIBETTAYEB:
Maybe it will never be malformed by Voyager, but it can be by using a custom client space functionalities, or by using other programming languages to CRUD to the database, so we should not give a chance to these bugs to happen.

emptynick:
We can never guarantee that there won't be malformed JSON.
Otherwise we had those checks everywhere in our code.
Closing.

Issue #4169, opened 2019-05-30
[BUG] S3 Disk string length to be >= 1


pamekar:
## Version information

 - Laravel: v5.8.17
 - Voyager: v1.2.3
 - PHP: 7.2
 - Database: MySQL 8.0

## Description

When using S3 as the default filesystem driver, you'll get this error when accessing the dashboard.

`Found 1 error while validating the input provided for the GetObject operation: [Key] expected string length to be >= 1, but found string length of 0 (View: /public_html/main/vendor/tcg/voyager/resources/views/media/manager.blade.php)`

## Steps To Reproduce
1. Change file system driver to S3
2. Try to access admin dashboard
3. See error

## Expected behavior
I should be able to have access to the admin dashboard without seeing this error message.

pamekar:
Twas fixed in #4170

pamekar:
Twas fixed in #4170

PR #4170, opened 2019-05-30
Fix S3 disk found string length of 0

pamekar:
When using S3 as the default filesystem driver, you'll get this error when accessing the dashboard.

Found 1 error while validating the input provided for the GetObject operation: [Key] expected string length to be >= 1, but found string length of 0 (View: /public_html/main/vendor/tcg/voyager/resources/views/media/manager.blade.php)

The error can be found in `/voyager/resources/views/media/manager.blade.php` on line 9.
```
 <div class="img_icon" :style="imgIcon('{{ Storage::disk(config('voyager.storage.disk'))->url('') }}'+file)"></div>
```
Adding a "/" to the `url` method solves this error. I opened an issue #4169 on this.

### DIFF
--- a/resources/views/media/manager.blade.php
+++ b/resources/views/media/manager.blade.php
@@ -6,7 +6,7 @@
                 <div class="file_link selected" aria-hidden="true" data-toggle="tooltip" data-placement="auto" :title="file">
                     <div class="link_icon">
                         <template v-if="fileIs(file, 'image')">
-                            <div class="img_icon" :style="imgIcon('{{ Storage::disk(config('voyager.storage.disk'))->url('') }}'+file)"></div>
+                            <div class="img_icon" :style="imgIcon('{{ Storage::disk(config('voyager.storage.disk'))->url('/') }}'+file)"></div>
                         </template>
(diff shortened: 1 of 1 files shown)


emptynick (approved):

samuelwei:
Hi @emptynick this same issue is back with laravel 9, due to the filesystem changes, see: #5684 and https://github.com/laravel/framework/issues/41509 
Could you please fix this? My local workaround was to use "./" instead of "/", but not sure if this is a future proof and working with all filesystem drivers

didix16:
Hello, I'm having the same issue. I'm using Laravel 9 and Voyager v1.6. Any fix for this problem?
imagen

ash2osh:
{project folder}/vendor/tcg/voyager/resources/views/media/manager.blade.php

use ./ instead of /

in

               <div class="img_icon"
                                    :style="imgIcon('{{ Storage::disk(config('voyager.storage.disk'))->url('./') }}' + file)">

George-nacouzi:
changing the vendor files is not a good practice, any better way to solve this ?

PR #4185, opened 2019-06-06
Add files via upload

cedlinx:
I installed Voyager and got the following warning and suggested workaround. 
  'Unable to locate "app/User.php".  Did you move this file?'
  'You will need to update this manually.  Change "extends Authenticatable" to "extends \TCG\Voyager\Models\User" in your User model'
And I usually like to put my models in the Models folder. I have also seen many apps that do the same. So, I decided to add the capability to search for User.php in both App and App\Models.

The slight changes I made to InstallCommand.php did it for me. Maybe someone would know a neater way to do that. But for now, I think Voyager install should not ignore those of us who like to use the Models folder for our models. What do you think?

### DIFF
--- a/InstallCommand.php
+++ b/InstallCommand.php
@@ -1,0 +1,151 @@
+<?php
+
+namespace TCG\Voyager\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Filesystem\Filesystem;
+use Intervention\Image\ImageServiceProviderLaravel5;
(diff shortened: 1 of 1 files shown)


fletch3555:
No intention to do this.  The docs outline what to do if you use a "non-standard" location for your model files.

cedlinx:
Alright. Keep up the good work

PR #4189, opened 2019-06-07
Fix issue #3623

### LINKED ISSUE #3623: UnexpectedValueException Invalid route action: [AnyController].
## Version information

 - Laravel: v5.7.9
 - Voyager: v1.1.7
 - PHP: 7.1.7
 - Database:  Ver 14.14 Distrib 5.7.19, for Win64 (x86_64)

## Description

at the time of adding a new Bread menu item (after create Bread),
error occurs when have is a invoke controller (Single Action Controllers)
## Steps To Reproduce
1. Go to console
2. Create a invoke controller
example `php artisan make:controller IndexController --invokable`
3. Register you controller in router
example `Route::get('/', 'IndexController');`
4. Go to /admin/bread and create new bread for any table (maybe you need create a table),
fill data and press submit
5. Look error.

## Expected behavior
Standart behavior bread creating

## Screenshots
https://screenshots.firefox.com/M6IoFQwFGXF6wYMb/flame.raiders

## Additional context
While looking for what's the problem

mallardduck:
The line I'm removing is the cause of #3623. Once removed the BREAD creation seems to work as expected, however with it included it just causes exceptions to be thrown.

Fixes #3623


fletch3555:
There was a reason for that to be there, but at the moment I don't see it.  My guess is it was removed from an unrelated change and this line was left even though it was no longer necessary.  We'll review this.

emptynick (approved):

PR #4197, opened 2019-06-13
Fix: Custom guard usage inside postLogin & authorize

ArturKp:
`$this->guard()` still defaults to the project default guard through the AuthenticatesUsers trait but instead it should use the guard binded to `VoyagerAuth`.

### DIFF
--- a/src/Http/Controllers/VoyagerAuthController.php
+++ b/src/Http/Controllers/VoyagerAuthController.php
@@ -34,7 +34,7 @@ class VoyagerAuthController extends Controller
 
         $credentials = $this->credentials($request);
 
-        if ($this->guard()->attempt($credentials, $request->has('remember'))) {
+        if (app('VoyagerAuth')->attempt($credentials, $request->has('remember'))) {
             return $this->sendLoginResponse($request);
         }
 
--- a/src/Http/Controllers/VoyagerAuthController.php
+++ b/src/Http/Controllers/VoyagerAuthController.php
@@ -12,7 +12,7 @@ class VoyagerAuthController extends Controller
 
     public function login()
     {
-        if (app('VoyagerAuth')->user()) {
+        if ($this->guard()->user()) {
             return redirect()->route('voyager.dashboard');
         }
 
@@ -34,7 +34,7 @@ class VoyagerAuthController extends Controller
 
         $credentials = $this->credentials($request);
 
-        if (app('VoyagerAuth')->attempt($credentials, $request->has('remember'))) {
+        if ($this->guard()->attempt($credentials, $request->has('remember'))) {
             return $this->sendLoginResponse($request);
         }
 
@@ -52,5 +52,15 @@ class VoyagerAuthController extends Controller
     public function redirectTo()
     {
         return config('voyager.user.redirect', route('voyager.dashboard'));
+    }
+
+    /**
+     * Get the guard to be used during authentication.
+     *
+     * @return \Illuminate\Contracts\Auth\StatefulGuard
+     */
+    protected function guard()
+    {
+        return app('VoyagerAuth');
     }
 }
--- a/src/FormFields/MediaPickerHandler.php
+++ b/src/FormFields/MediaPickerHandler.php
@@ -26,7 +26,7 @@ class MediaPickerHandler extends AbstractHandler
         }
 
         if (isset($options->base_path)) {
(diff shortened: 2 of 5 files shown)


src/Http/Controllers/VoyagerAuthController.php
```diff
@@ -34,7 +34,7 @@ public function postLogin(Request $request)
 
         $credentials = $this->credentials($request);
 
-        if ($this->guard()->attempt($credentials, $request->has('remember'))) {
+        if (app('VoyagerAuth')->attempt($credentials, $request->has('remember'))) {
```
fletch3555: `VoyagerAuth` is just `auth()` ([by default](https://github.com/the-control-group/voyager/blob/1.2/src/VoyagerServiceProvider.php#L71)), so shouldn't it be `app('VoyagerAuth')->guard()`...?

If so, I'm not sure how this change fixes anything since it still uses the default guard.

ArturKp: In my case, my project default is set to `api` and VoyagerAuth is now configured to use `web` instead.

```
$this->app->singleton('VoyagerAuth', function () {
    return Auth::guard('your-custom-guard');
});
```


Guard was changed like it says in the docs: [customization/custom-guard](https://docs.laravelvoyager.com/customization/custom-guard). Also all other default guard calls were already changed to use VoyagerAuth in #3968.

ArturKp: On the other hand I'm also getting somewhat confused. As I submitted the PR I knew it wasn't a fix for all my problems, but it allowed me to at least log in and it still seems to make much sense regarding changes made in #3968. 

But #3968 was made so that we can use a custom guard like it's also described in the docs. However since 1.2.0 Voyager is using Laravel's own Policies and Gate's which use the default guard anyway (`AuthorizesRequests`). If these assumptions are correct then it might need a lot more changes to make `VoyagerAuth` work like it's supposed to or it might be wise to remove `VoyagerAuth` completely from this project. Does any of this make sense?

ArturKp:
Some more background info about this PR and issue it solves.

Our project's default Guard is set to use JWT tokens and api middleware. Voyager routes on the other hand use built-in sessions and web middleware.

So far previous developer had copied and modified some code to make everything work. Primarily seems to be:

```
VoyagerAuthController.php - To make login use the web guard instead of the default
routes.php                - To add web middleware to most of the routes
```

One day I stumbled upon Voyager's docs which stated that it's possible to just change the default auth guard ([customization/custom-guard](https://docs.laravelvoyager.com/customization/custom-guard)). This made it sound like our overrides were no longer needed. So I removed some of our overrides, changed the default guard for Voyager, but couldn't get it to work. After some digging found that:

1. `VoyagerAuthController@postLogin` still uses default Guard instead the one set for VoyagerAuth. 
2. After v1.2 all `->authorize()` calls also default to the project's default guard (AuthorizesRequests).

---

**Edit**: I'm fairly certain that following also need an update. If somebody can confirm I can add to this PR: 

3. `MediaPickerHandler@createContent:29` uses `\Auth::user()`
4. `VoyagerMediaController@upload:209` uses `\Auth::user()`

emptynick:
> **Edit**: I'm fairly certain that following also need an update. If somebody can confirm I can add to this PR:
> 
> 1. `MediaPickerHandler@createContent:29` uses `\Auth::user()`
> 2. `VoyagerMediaController@upload:209` uses `\Auth::user()`

Thats correct. Those files were created/changed after we implemented #3968.
Are you able to provide a phpunit test which tests login/logout and some other things with a custom guard?

ArturKp:
> Thats correct. Those files were created/changed after we implemented #3968.
> Are you able to provide a phpunit test which tests login/logout and some other things with a custom guard?

👍🏻 Yes. Should be able to do this next week.

superhaggis:
Just thought I'd add that an alternative approach for `src/Http/Controllers/VoyagerAuthController.php` would be to add the following override at the bottom of the class:

```php
    protected function guard()
    {
        return app('VoyagerAuth');
    }
}
```
😃

ArturKp:
@superhaggis Yep, that's better. Made the change.

emptynick:
Looks like you are done here?

ArturKp:
> Looks like you are done here?

Yes. Everything's done.

emptynick (approved):

PR #4200, opened 2019-06-15
Added edit bread button to bread's browse view

anoopnaik:
It's really helpful in case of many breads and you want to quickly change the bread view.


fletch3555:
I'm not entirely opposed to this, but at the bare minimum, it needs a permission check around it.

I'd also like to see some testing for it, and a fix for the tests that it's currently breaking

emptynick:
Also, I think this should only be visible when `show_dev_tips` config is set to true or the environment is develop.

PR #4209, opened 2019-06-20
Update bread.php

zatrano:


fletch3555:
Next time, please add some description or justification of why you're making the change.  In this particular care I can see it's very simple, but generally this would not be adequate.  Thank you

fletch3555 (approved):

PR #4221, opened 2019-06-21
Update generic.php

zatrano:
Add
'logout'                 => 'Çıkış',


fletch3555:
Next time, when making simple i18n changes like this and #4220, especially for the same language in the same file, just make a single pull request.

fletch3555 (approved):

Issue #4031, opened 2019-03-24
belongsTo Relationship bug Voyager 1.2


tunapiq:
## Version information

 - Laravel: v5.7.#
 - Voyager: v1.2.#
 - PHP: 7.#

## Description

BelongsTo relationship will not accept or store the relationship values.

Bug fix:
Path: resources/views/formfields/relationship.blade.php
Change
`                <select
                    class="form-control select2-ajax" name="{{ $options->column }}"
                    data-get-items-route="{{route('voyager.' . $dataType->slug.'.relation')}}"
                    data-get-items-field="{{$row->field}}"
                >`

To 

`                <select
                    class="form-control select2-ajax" name="{{ $relationshipField }}"
                    data-get-items-route="{{route('voyager.' . $dataType->slug.'.relation')}}"
                    data-get-items-field="{{$row->field}}"
                >`

Finally, the .select2-ajax doesn't add a "none" option when a selection is being made. This means that a user will not be able to deselect an option if they decide they want to.

suxur:
I'm also experiencing this issue.

xikaos:
@suxur @tunapiq You guys tried to enable all BREAD actions for the field that refers to the other table?
I was experiencing the same issue until saw [this](https://github.com/the-control-group/voyager/pull/2141#issuecomment-357517483) and was able to make it work.

tunapiq:
The relationship bug is a result of sending the wrong field name to the controller.

 name="{{ $options->column }}"
should be:
 name="{{ $relationshipField }}" 

the $relationshipField variable was created but not used.  That is the correct fix for the bug, toggling all options on the BREAD shouldn't be necessary.

fletch3555:
PR's welcome @tunapiq

xikaos:
@fletch3555 Just made the PR with the fix @tunapiq suggested.

tunapiq:
If anyone needs it, here is a fix for the missing "None" option for the select2-ajax (relationship select drop-downs), change 

$results = [];
to
```
$results = [
    [
        'id'   => '',
        'text' => "--None--",
    ]
];
```
in the method VoyagerBaseController@relation. Doing that will give a user the ability to deselect an option if they so decide it.

fletch3555:
@tunapiq, It's not as simple as that.  What if the field isn't nullable?  This would allow a user to select an empty value, and when saving, it would error.

tunapiq:
[...]

PR #4228, opened 2019-06-24
Fix belongsTo relationship field name.

xikaos:
This issue was confirmed by 3 users in [this issue](https://github.com/the-control-group/voyager/issues/4031).

Credits to @tunapiq for suggesting this fix and mine for making the PR.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -23,7 +23,7 @@
             @else
 
                 <select
-                    class="form-control select2-ajax" name="{{ $options->column }}"
+                    class="form-control select2-ajax" name="{{ $relationshipField }}"
                     data-get-items-route="{{route('voyager.' . $dataType->slug.'.relation')}}"
                     data-get-items-field="{{$row->field}}"
                     data-method="{{ isset($dataTypeContent) ? 'edit' : 'add' }}"


emptynick (approved):

MrCrayon:
@emptynick I think it should be better to revert this commit.
This PR doesn't solve the bug because bug had nothing to do with name value used.
The problem is that with belongsTo relationship there are two rows in data_rows that try to save the value and one will be NULL, so if the second one is not the chosen one for the name used in form value will be reset to NULL.

So it doesn't matter which one we use in name as far as people know that they have to order them in a certain way.

What this PR removes is validation for belongsTo that could have been set in the field but now can't because we can't set validation rules for relationships.

Other bug #4437 as reference for the problem still happening.

madman-81:
@MrCrayon Could this be the cause of "Not null violation" errors that I see after updating Voyager from 1.2.6 to 1.2.7? 

To reproduce: I have a table/bread with a belongsTo relationship to another table/bread. When I open an existing record for editing and try to save it (so without making any changes), I'll get a "not null violation" on the belongsTo field. Note that this field had a value when I saved it. 

When I downgrade to 1.2.6 and try the same steps with the same record, all works fine again.

MrCrayon:
@madman-81 quite likely that's the reason.
You can check in your BREAD configuration if the whatever_id field comes after the relationship field than that was overwriting with NULL the value.

You can update again your platform and change fields order, that should fix it.

For more info you can check my PR #4439 and related issues linked there.

madman-81:
@MrCrayon Yes, that has done the trick! Thanks for the quick reply. 

Does your new PR fix this problem? In my opinion the order of the columns should not matter for the behaviour. Ideally you should be able to change the whatever_id field to a relationship instead of having them both in the bread configuration.

MrCrayon:
@madman-81 yes, my PR makes the order irrilevant.
As for not having to have another field actually you can set the relationship in whatever_id details even if it's a while I don't use it that way.
Check https://voyager.readme.io/docs/relationships

PR #4230, opened 2019-06-25
TinyMCE Link add file upload

merterhk:
Edit for TinyMCE link plugin to upload other file formats.


fletch3555:
Please undo your change to app.js.  These compiled files shouldn't be committed since we have an automated process that does the compilation for us once merged.  Committing them leads to merge conflicts.

src/Http/Controllers/VoyagerController.php
```diff
@@ -63,6 +63,10 @@ public function upload(Request $request)
             } else {
                 $status = __('voyager::media.error_uploading');
             }
+        }else if(in_array($ext, ['pdf','zip', 'doc', 'docx'])){
+            $request->file('image')->storeAs(config('voyager.storage.subfolder') . $path, $filename. '.' . $file->getClientOriginalExtension());
```
emptynick, line 67: Where does `voyager.storage.subfolder` come from?

src/Http/Controllers/VoyagerController.php
```diff
@@ -63,6 +63,10 @@ public function upload(Request $request)
             } else {
                 $status = __('voyager::media.error_uploading');
             }
+        }else if(in_array($ext, ['pdf','zip', 'doc', 'docx'])){
```
emptynick, line 66: Where do these extensions come from?
We have a defined set of extensions for the media-manager. Maybe we can use those?

emptynick (changes requested):

MACscr:
Nothing ever got resolved here? I really need to be able to attach pdf, zip, doc, excel, etc, files.

Issue #1923, opened 2017-10-02
Mysql support for ENUM


FrodeMMeling:
- Laravel Version: 5.5.13
- Voyager Version: v1.0.4
- PHP Version: 7.1.8
- Database Driver & Version: Ver 14.14 Distrib 5.7.19, for osx10.12 (x86_64) using  EditLine wrapper

### Description:
ENUM is disabled and "not supported" for Mysql. Why is this?
Strange since Voyager itself is using it in for instance pages.

### Steps To Reproduce:
Create a table and try to add an ENUM.

DaKe-Zhang:
having the same issue.

calder12:
This also is an issue when trying to add columns to Pages or Posts, it throws an error that ENUMs are not supported and won't save the table changes.

FrodeMMeling:
I have a workaround for this, until its fixed/supported.
- In you new table, Create the enum field as an integer in database
- Go directly in database and change type to ENUM (look at pages structure for details)
- Set it as Dropdown in BREAD edit and add the possible fields in Option details (Look at pages BREAD for details).

This seems to work, while waiting for fix/support

SAM-AEL:
Have to Edit Database directly cause of this error.. enum works fine on database but voyager keeps saying "generic.exception : enum is not supported" on saving database.. weird cause it was working fine till today.. waiting for fix

fletch3555:
@marktopper

whyayala:
I believe this is a known issue with Doctrine, not necessarily Voyager 

https://github.com/laravel/framework/issues/1186

fletch3555:
@whyayala, it is, which is why we use @marktopper's [package](https://github.com/larapack/doctrine-support) which implements the workarounds suggested on that issue.  Seems like there's a new issue introduced somewhere (either Voyager or his package, or elsewhere that broke these workarounds... not sure)

nishant-git:
Quickfix : Change enum type to varchar in voyage.

jf-m:
Well.. According to this line 
https://github.com/the-control-group/voyager/blob/b8bfb3c71632f808a2f4902cb68d308e1dbc8f42/src/Database/Types/Mysql/EnumType.php#L15

As well as this one
https://github.com/the-control-group/voyager/blob/b8bfb3c71632f808a2f4902cb68d308e1dbc8f42/src/Database/Platforms/Mysql.php#L25

It pretty much seems like Voyager does not support mysql enum neither.

[...]

PR #4234, opened 2019-06-29
Ensure enum column types do not prevent editing tables in mysql datab…

damms005:
In my experience working with enum types in Mysql, this problem with Voyager vs Enum types usually arise when progressively porting legacy code bases to Voyager.

[While it is clear not to use enums](http://komlenic.com/244/8-reasons-why-mysqls-enum-data-type-is-evil/), porting progressively as described above becomes difficult since Voyager currently does not allow any tables with enum cols to be edited (which usually the purpose of editing the schema may not even be about the enum col).

This PR is my attempt to resuscitate [#1923](https://github.com/the-control-group/voyager/issues/1923). Since [doctrine are stateless](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.9/reference/types.html), I took a hint from [the recommended fix](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/cookbook/mysql-enums.html), which is easy because [Voyager already has such abstraction](https://github.com/the-control-group/voyager/blob/b00a09028a264c3ed61b39d0fa6db9928a506bb9/src/Database/Types/Mysql/EnumType.php).

I believe that if this PR is accepted as the right step in the right direction, it is only a matter of time before the UI and other components of Voyager will completely support enum types.

### DIFF
--- a/src/Database/Schema/Column.php
+++ b/src/Database/Schema/Column.php
@@ -8,11 +8,12 @@ use TCG\Voyager\Database\Types\Type;
 
 abstract class Column
 {
-    public static function make(array $column)
+    public static function make(array $column, string $tableName = null)
     {
         $name = Identifier::validate($column['name'], 'Column');
         $type = $column['type'];
         $type = ($type instanceof DoctrineType) ? $type : DoctrineType::getType(trim($type['name']));
+        $type->tableName = $tableName;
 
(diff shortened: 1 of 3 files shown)


cristianmarint:
@the-control-group pls merge! :)

codebygina:
It works!

emptynick (approved):

PR #4238, opened 2019-07-02
Update multilingual.js

MertOcak:


resources/assets/js/multilingual.js
```diff
@@ -25,54 +25,43 @@
  *
  */
 ;( function( $, window, document, undefined ) {
-
```
fletch3555, line 28: Please refrain from making these arbitrary whitespace changes.  I'm sure it was likely some auto-formatting from your editor, but you need to remain aware of what you're committing.  These changes provide no benefit at best, and actually break our code style at worst

fletch3555 (changes requested):
Please revert your whitespace-only changes.  It'll make this PR diff look MUCH cleaner/simpler

emptynick:
Unfortunately you deleted your fork, so we are not able to merge this. Closing.

PR #4242, opened 2019-07-05
fix set form input width

mrkindy:
fix set form input width previous behavior was getting value from child node "option" although in documentation  it should be a parent node like
`{
    "width" : 6
}`

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -59,6 +59,7 @@
                                 <!-- GET THE DISPLAY OPTIONS -->
                                 @php
                                     $display_options = $row->details->display ?? NULL;
+                                    $form_input_width = $row->details->width ?? 12;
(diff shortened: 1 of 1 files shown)


fletch3555:
I think I'd rather just update the docs to match. Any option @emptynick?

emptynick:
The docs already mention that you have to put `width` into `display`:
https://docs.laravelvoyager.com/bread/introduction#display-options
Or did I get something wrong?

mrkindy:
> The docs already mention that you have to put `width` into `display`:
> https://docs.laravelvoyager.com/bread/introduction#display-options
> Or did I get something wrong?

please check this link
https://docs.laravelvoyager.com/bread/introduction-1#width

emptynick:
I just moved all formfield specific things from the BREAD-introduction to the formfields-introduction (with the correct JSON).
Closing.

PR #4249, opened 2019-07-06
Allows NULL on relationship with AJAX request

### LINKED ISSUE #4031: belongsTo Relationship bug Voyager 1.2

hitech95:
When jSelect ask for the possible relationships the response ignore the required option of the field.

This will add back the functionality, it also prevents the output of the 'NONE' option when using the search query.

Fixes https://github.com/the-control-group/voyager/issues/4031

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -661,6 +661,14 @@ class VoyagerBaseController extends Controller
                 }
 
                 $results = [];
+
+                if (!$row->required && !$search) {
+                    $results[] = [
+                        'id'   => '',
+                        'text' => __('voyager::generic.none'),
+                    ];
+                }
+
                 foreach ($relationshipOptions as $relationshipOption) {
                     $results[] = [
                         'id'   => $relationshipOption->{$options->key},


emptynick:
As already mentioned in https://github.com/the-control-group/voyager/issues/4031 its not as simple as that.

hitech95:
I have the check to see if the field is required or not.
So unless your database schema is not consistent it should be working.

it is the same soution proposed (and merged) on #3170 but for the ajax requests.

gelinger777:
When can you merge this? This is very important patch.

hitech95:
> 
> 
> When can you merge this? This is very important patch.

I've created the overrides for all the controllers on my project so I have this patch applied even if this is not merged.

royercross (commented):
Would be better, if you add it on line 664+ just after "$results = [];" so it get added first than the results.

gelinger777:
is this released in composer package?

gelinger777:
I agree with @royercross It should come first not at the end

hitech95:
> 
> 
> Would be better, if you add it on line 664+ just after "$results = [];" so it get added first than the results.

Updated with the changes.



> 
> 
> is this released in composer package?

If you need this (before the merge) you should override the controllers and apply the patch by yourself.

emptynick (approved):

PR #4250, opened 2019-07-06
Skip fields if details columns is not set

hitech95:
Some fields could have the details set as NULL or the type is not set.

This patch allow to skip thoose fields withthout details or type.
As specified in the code comments the only exceptions are:
 - Checkboxes
 - BelongsToMany relationship

Example why this is needed:
I have some columns that are managed by another laravel package, I remove thoose request arguments and I write them with that library functions and then I'll pass the request to the Voyager BREAD controller.

Unfortunatly without this patch the Voyager will override my changes due to the fact that the field is absent from the request and the field is not skipped.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -58,7 +58,9 @@ abstract class Controller extends BaseController
             if (!$request->hasFile($row->field) && !$request->has($row->field) && $row->type !== 'checkbox') {
                 // if the field is a belongsToMany relationship, don't remove it
                 // if no content is provided, that means the relationships need to be removed
-                if (isset($row->details->type) && $row->details->type !== 'belongsToMany') {
+                if (isset($row->details->type) && $row->details->type === 'belongsToMany') {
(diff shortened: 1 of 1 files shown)


emptynick:
You should be able to just uncheck all checkboxes for that field and Voyager won't touch them.
This line of code is extremly delicate and touching it caused a lot of trouble in the past (see TravisCI).

hitech95:
> 
> 
> You should be able to just uncheck all checkboxes for that field and Voyager won't touch them.
> This line of code is extremly delicate and touching it caused a lot of trouble in the past (see TravisCI).

Ok, how should I proceed?
I have the data coming from the BREAD page but I HAVE to intercept that before the BREAD controller and manipulate it, and then update or create the model.

Since the field skip feature is not working (with default code) how can I proceed?

emptynick:
You can use Laravels Mutators to manipulate data stored to the DB and/or eloquent-events.
Closing this as it's not necessary.
If there are plausible reasons this is needed this can of course be reopened.

Issue #4137, opened 2019-05-14
[BUG] Tinymce font loading over HTTPS gives mixed content error.


sten:
## Version information

 - Laravel: v5.8.10
 - Voyager: v1.2.3
 - PHP: 7.2.1
 - Database: Mysql 5.7.22

## Description
When voyager is used over HTTPS and you load the create or edit form of a BREAD with a rich text editor, the font of Tinymce is not loaded. This results in a broken UI, i.e. the buttons do not have any icons.

## Steps To Reproduce
1. Host Voyager over HTTPS
2. Go to the create or edit of a BREAD with a rich text field
3. Check the rich text editor for missing icons
4. Check the browser console for errors, like: 
- Mixed Content: The page at 'https://www.mydomain.com/admin/pages/create' was loaded over HTTPS, but requested an insecure font 'http://www.mydomain.com/admin/assets?path=js/skins/voyager/fonts/tinymce.woff'. This request has been blocked; the content must be served over HTTPS.
- Mixed Content: The page at 'https://www.mydomain.com/admin/pages/create' was loaded over HTTPS, but requested an insecure font 'http://www.mydomain.com/admin/assets?path=js/skins/voyager/fonts/tinymce.ttf'. This request has been blocked; the content must be served over HTTPS.

## Expected behavior
The TinyMCE font should be properly loaded and the icons should be displayed in the editor.

## Additional context
I have been debugging this a bit. It appears the fonts are loaded in `skin.min.css` via the url() function of CSS. However `skin.min.css` is loaded from an ajax call from `app.js` via the assets REST service. Within app.js the assets REST service is called over HTTPS, which is correct. I assume the dynamically loaded skin.min.css does not have any HTTPS context.

estebangallego:
Same issue, please advice. 
![Screen Shot 2019-06-18 at 6 48 09 PM](https://user-images.githubusercontent.com/5769433/59731081-b0c38a80-91f9-11e9-8bc5-d31ebd5c86a8.png)

apiaget:
Hello,

I also encountered that issue.

I looked at my chrome inspector and tried to spot the differences between the fonts that load correctly and the tinymce one. I noticed that there is a slash (`/`) in the failed request between `voyager-assets` and `?path=...`.
A failed request : `https://karac.ch/admin/voyager-assets/?path=js/skins/voyager/fonts/tinymce.woff`
A successfull request : `https://karac.ch/admin/voyager-assets?path=fonts/voyager.woff`

[...]

PR #4251, opened 2019-07-07
Use a label tag element for the remember me text

apiaget:
Hello !

I suggest replacing the span by a label on the login screen of the admin panel.

With this change, clicking on the remember me label also clicks on the associated checkbox.

There is no visual change (tested on Firefox, Chrome and Edge).

Any feedback is welcome ;-)


apiaget:
Oups I wanted to create a second pull request for this modification but my commit came in this one.

So I also changed the url links to the tinymce fonts from `voyager-assets/?path=...` to `voyager-assets?path=...`.

This modification removes mixed content errors and therefore the broken UI as we could see in the screenshots of that issue : #4137 

I couldn't find any non minified source for the skin.min.css file so I assumed that the modification had to be done there.

Any feedback is welcome ;-)

fletch3555:
The slash before the querystring shouldn't cause any mixed-content errors.  Can you revert that change and open a separate issue describing what you're seeing and why you think this is the solution?

apiaget:
I reverted the commit about the TinyMCE problem.

I should add that there is already an issue describing that exact problem, I will add more screenshots and informations to it : #4137 

I may submit a second pull request about that TinyMCE after that.

emptynick (approved):
Can't hurt. Thanks

fletch3555 (approved):

PR #4252, opened 2019-07-08
Category model fix in "Add/Edit Post" template

superhaggis:
## Version information

 - Laravel: v5.8.27
 - Voyager: v1.2.6
 - PHP: 7.2
 - Database: MySQL 5.7

## Description

By default, the "Add/Edit Post" template uses a hard-coded model class facade to populate a dropdown box with Categories.  Using a custom Category model with a different namespace results in a 500 error.

## Steps To Reproduce
1. Sub-class the Voyager Admin model in your own application namespace (by adding Voyager::useModel(...) calls to AppServiceProvider#register)
2. Log into Voyager Admin backend
3. Click the "Posts" menu item
4. Click "Add New"

## Expected behavior
The "Add/Edit Post" screen should load without error.

### DIFF
--- a/resources/views/posts/edit-add.blade.php
+++ b/resources/views/posts/edit-add.blade.php
@@ -212,7 +212,7 @@
                             <div class="form-group">
                                 <label for="category_id">{{ __('voyager::post.category') }}</label>
                                 <select class="form-control" name="category_id">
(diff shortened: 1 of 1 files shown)


fletch3555 (approved):
Sorry about that! Thought we caught all of these....

PR #4253, opened 2019-07-08
BugFix: Setting key can't have more than 1 dot character

AndreiTelteu:
I view this as a bug. 
I can not see any reason why this would be considered a feature, right ?
Or maybe we can improve this further with the use of laravel's helpers `array_get` or `data_get`


emptynick:
It's a feature that you are able to get all settings in a group by calling `setting('group_name')`

emptynick:
Closing as this is a desired feature. Although we might overthink this way in a future version.

Issue #4263, opened 2019-07-13
[BUG] Select search returns no results on belongsTo relationship using additional_attribute


DaLooK:
## Version information

 - Laravel: v5.7.28
 - Voyager: v1.2.6
 - PHP: 7.1
 - Database: MySQL 5.7.26

## Description

When using a belongsTo relationship and selecting an **additional_attribute** as the display column, at the time of adding a new model and searching using the select2 ajax searchbar, an error is reported as the MySQL query tries to look for the additional_attribute as a column in the table.

## Steps To Reproduce
1. Add an [additional_attribute](https://docs.laravelvoyager.com/customization/custom-realtionship-attributes) to a model (_Model1_)
2. Add a **belongsTo** relationship from another model (_Model2_) to the model with the additional_attribute (_Model1_)
3. Select the additional_attribute as the display column for the **belongsTo** relationship
4. Try create a new _Model2_ and use the search bar in the relation field to _Model1_
5. No results should be returned
6. An error should be logged to the error log with something like **SQLSTATE[42S22]: Column not found: 1054 Unknown column**

## Expected behavior
When searching for _Model1_ names, they should be filtered and returned to the list

## Screenshots
![image](https://user-images.githubusercontent.com/7662026/61176573-a5f5de80-a588-11e9-8ff6-7adb6f0fbb9c.png)
![image](https://user-images.githubusercontent.com/7662026/61176576-c45bda00-a588-11e9-87f6-4a6f44b6b345.png)
![image](https://user-images.githubusercontent.com/7662026/61176579-d3db2300-a588-11e9-89bc-7751dc39954a.png)
![image](https://user-images.githubusercontent.com/7662026/61176584-db9ac780-a588-11e9-9bee-028206e2e33a.png)
![image](https://user-images.githubusercontent.com/7662026/61176590-f0775b00-a588-11e9-81fd-9a9a3deb2915.png)

MrCrayon:
Fixed in #4562

PR #4264, opened 2019-07-13
Fix issue with belongsTo relationships and additional_attributes

DaLooK:
* Now we check if the display column of the belongsTo relationship is an additional_attribute, if so, we filter using collections instead of eloquent queries

Fixes https://github.com/the-control-group/voyager/issues/4263


emptynick:
I think we need some testing for this.

MrCrayon:
I don't know how I completely missed this one :thinking: 
Superseded by #4562

PR #4283, opened 2019-07-23
Disable translations

howdu:
If multilanguage is disabled in config don't translate menu

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -108,7 +108,7 @@ class Menu extends Model
     {
         $items = $items->transform(function ($item) {
             // Translate title
-            $item->title = $item->getTranslatedAttribute('title');
+            $item->title = config('voyager.multilingual.enabled') ? $item->getTranslatedAttribute('title') : $item->title;
             // Resolve URL/Route
             $item->href = $item->link(true);
 


fletch3555 (approved):

MrCrayon:
I know this has already been merged but I just saw it.

There is no explanation why this pull request is needed but I was thinking that if leaving getTranslatedAttribute can cause any problem it should be addressed in getTranslatedAttribute Method of Translatable Trait, otherwise we should check everywhere if multilingual is enabled.

Am I wrong?

fletch3555:
That's a very good point @MrCrayon.  Would you like to open a PR with that change?  If you're busy, I can take care of it.

MrCrayon:
@fletch3555 I can do it but I don't have clear what's the problem this pull request was trying to resolve. Is there a related issue that I can check?

howdu:
If multi language is disabled from config don't run translation database queries. 

The call to getTranslatedAttribute() from Models/Menu.php needs to work similar to is_bread_translatable() checking if voyager.multilingual.enabled.

MrCrayon:
So if I understand it correctly there is no issue solved but it's a performance improvement.

PR #4290, opened 2019-07-26
Avoid urlencoded title of Breadcrumb

vkiranmaniya:
If the route contains space, breadcrumb shows the URL encoded text as breadcrumb item. Like, if Route is 'Consumer Report', Breadcrumb will show Consumer%20report, it should be 'Consumer Report'.  I changed the Breadcrumb title URL decoded.


resources/views/dashboard/navbar.blade.php
```diff
@@ -21,10 +21,10 @@
                         $url .= '/'.$segment;
                         @endphp
                         @if ($loop->last)
-                            <li>{{ ucfirst($segment) }}</li>
+                            <li>{{ ucfirst($segment($segment)) }}</li>
```
fletch3555: What's the purpose of this change..? Doesn't seem like it would even work.

vkiranmaniya: Breadcrumbs show 'Cunsumer%20Report' When I overrode the default route file of Voyager.
Step to reproduce
1. Go to tcg\voyager\route\voyager.php and define route as Route::post('Consumer Report' , 'someController@report');
2. Create a view that extends Voyager::master view.
3. Return that view from the controller method `report`
4. Now navigate to that route and check Breadcrumb items

It's not showing breadcrumb title properly when the route has space '%20' encoded inside it.
However, I was mistaken on line 24, there should be {{ucfirst(urldecode($segment))}}.
i fixed it next commit.

vkiranmaniya: > What's the purpose of this change..? Doesn't seem like it would even work.

I updated the code and committed the changes to fork. please review the pull request.

emptynick (approved):

PR #4291, opened 2019-07-26
Add platform mssql

mirkin-pixel:
Please add platform mssql to voyager, some of the options are not working right now.


fletch3555:
Please search before opening issues. This has been requested/mentioned multiple times.

Having said that, this is a Pull Request, not an issue, so it's completely wrong.  If you're new to github or otherwise feel like you have no idea what you're doing, please join our slack and post there instead.

PR #4292, opened 2019-07-27
Hide delete button when has no delete permission

nysso:
Hide delete button in Settings page when the user has no **delete settings** permission.
So instead of the forbidden action exception (code 403),
the user won't be able to see/click the delete button


fletch3555 (approved):

fletch3555:
Good find! Not sure how that one slipped past us.

PR #4295, opened 2019-07-30
Skip translation if multilingual is not enabled

MrCrayon:
Following #4283 this PR skips all checks and db queries if multilingual is not enabled.

As coding style I would prefer to have:
```php
    public function getTranslatedAttribute($attribute, $language = null, $fallback = true)
    {
        // If multilingual is not enabled don't check for translations
        if (!config('voyager.multilingual.enabled')) {
            $value = $this->getAttributeValue($attribute);
        } else {
            list($value) = $this->getTranslatedAttributeMeta($attribute, $language, $fallback);
        }

        return $value;
     }
```

Please let me know which one you prefer.

### DIFF
--- a/src/Traits/Translatable.php
+++ b/src/Traits/Translatable.php
@@ -131,6 +131,11 @@ trait Translatable
      */
     public function getTranslatedAttribute($attribute, $language = null, $fallback = true)
     {
+        // If multilingual is not enabled don't check for translations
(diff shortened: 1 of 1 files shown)


emptynick:
Should we revert the changes from #4283 then?

MrCrayon:
Yes, that one is not needed anymore.

emptynick (approved):

Issue #4315, opened 2019-08-03
[BUG] Multiple Images Field, Cannot delete image


ostrichegret:
## Version information

 - Laravel: v5.8.30
 - Voyager: v1.2.6
 - PHP: 7.2.3
 - Database: MySQL 8 8.0.15

## Description

Cannot delete uploaded images on Multiple Images Field

## Steps To Reproduce
1. Install Laravel and Voyager
2. Create table 'test'
3. column test_id = int, test_name = varchar 190, test_gallery = text
4. Create BREAD on test table: test_id = text_field, test_name = text_field, test_gallery = Multiple Images
5. Input record on test
6. Add test name and put 2 images on test_gallery, submit
7. Edit the field and try to remove the images on test_gallery = Fail, add image still works. 

## Expected behavior
It should be able to delete the uploaded images on multiple images field

## Screenshots
Header:
Provisional headers are shown
Accept: */*
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
DNT: 1
Origin: http://127.0.0.1:8000
Referer: http://127.0.0.1:8000/admin/test/1/edit
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36
X-CSRF-TOKEN: qkwcuvofk0gUul7oSq644wL3b87XluIg1VqG9sCP
X-Requested-With: XMLHttpRequest
slug: test
filename: test\August2019\oHM45IsWkl2bOllHEAtP.jpg
id: 
field: test_gallery
multi: true
_token: qkwcuvofk0gUul7oSq644wL3b87XluIg1VqG9sCP

Response:
{"data":{"status":400,"message":"Field does not exist"}}

emptynick:
Please search before opening new issues.
Duplicate of https://github.com/the-control-group/voyager/issues/4095

ostrichegret:
@emptynick the browser media permission is set.

Header:
Provisional headers are shown
Accept: /
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
DNT: 1
Origin: http://127.0.0.1:8000
Referer: http://127.0.0.1:8000/admin/test/1/edit
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36
[...]

PR #4316, opened 2019-08-03
change hardcoded dataTypeContent->id to dataTypeContent->getKey()

ostrichegret:
Fix this error:
https://github.com/the-control-group/voyager/issues/4315

and possible bugs with other fields


ostrichegret:
too aggresive search and replace to wrong code

PR #4325, opened 2019-08-08
Support temporary url

G3z:
Laravel supports temporary url for storages like S3.
Now it's supported in Voyager :)


emptynick (changes requested):
The Travis-CI tests have to pass, what they currently don't.

publishable/config/voyager.php
```diff
@@ -61,6 +61,8 @@
 
     'storage' => [
         'disk' => env('FILESYSTEM_DRIVER', 'public'),
+        'use_temporary_url' => true,
```
MrCrayon: You are setting this as true in a driver that doesn't support temporaryUrl

publishable/config/voyager_dummy.php
```diff
@@ -57,6 +57,8 @@
 
     'storage' => [
         'disk' => env('FILESYSTEM_DRIVER', 'public'),
+        'use_temporary_url' => true,
```
MrCrayon: Same as above

MrCrayon (commented):
temporayUrl only works on S3 driver

MrCrayon:
You are duplicating same checks done in FilesystemAdapter, not sure if it breaks any standard but could it be just be like this?
```php
    public function storage_url($file)
    {
        $disk = Storage::disk(config('voyager.storage.disk'));

        if (config('voyager.storage.use_temporary_url', false)) {
            try {
                $cache_time = now()->addMinutes(config('voyager.storage.temporary_url_cache', 5));

                return $disk->temporaryUrl($file, $cache_time);
            } catch (RuntimeException $e) {
            }
        }

        return $disk->url($file);
    }
```

G3z:
that was my first idea but I didn't like catching an exception as generic as RuntimeException.
if you prefer that, I can change it
I returned `return $disk->url($file);` in the catch block, I think it should be ok (maybe should i log something ?)

MrCrayon:
RuntimeException is the one dispatched:
https://github.com/laravel/framework/blob/v6.8.0/src/Illuminate/Filesystem/FilesystemAdapter.php#L523
so I think it's the perfect one to catch.

Honestly let's see what @fletch3555 says, he is the strict one :grin: and I can't approve PR anyway, I'm just advising.

Edit: wrong url

G3z:
I noticed the exception. I feared that being too generic it could be thrown by something else (like the adapters) for different reasons than the one we are catching.
If it's fine for you it will be for me :)

PR #4331, opened 2019-08-12
[1.x] Add config `voyager.clean_data_rows`

ajcastro:
This lets the developer decide whether to delete fields which don't have equivalent table columns


emptynick:
This needs to go to 1.x because the config-file needs an upgrade.

emptynick:
While I like the idea, I dislike configuration-options.
Instead I think we should have a checkbox for each row where you can check to keep the row (pre-checked based on the existing fields in the table).
That way users can decide which rows to keep and which to delete (after deleting fields for example)

PR #4332, opened 2019-08-12
Update MultipleImage.php: keep original files names

ADavid25:
Keep original name files with the bread field name option "preserveFileUploadName":true


Makashov:
But isn't it better to keep file names random?

Issue #4334, opened 2019-08-12
Allow to have non-default admin guard


niels-numbers:
## Version information
 - Laravel: v5.6.39
 - Voyager: v1.2.6
 - PHP: 7.2.14my
 - Database: MySQL Server version 5.7.25 

## Description of problem
Creating a separate admin table has been requested multiple times (#2572,#2534,#2568, #3398, #2002). 
One of many concerns was the following:

> If Voyager admin's can not be separated from default user table , then one has to use something like \Auth()->guard('normal') instead of \Auth() in the whole application because the default guard is already used by Voyager.

(from #2572) or in other words

> I have 3 user types in my project; customers, business admins and the app admins. I using a multiple authentication system for the project and I would like to leave laravel's default user authentication and User model for customer authentication.

(from #2002)

In #3968 emptynick created the possibility to use a custom guard. This is a great improvement. However, even if one creates a custom admin guard, one is forced to use this as the default guard for the whole application (otherwise you  get issues like #4264 or #4333). But this forces every developer who uses `voyager` to rewrite in his application `\Auth()->guard('normal')->user()...`. 

In short, we can change the Laravel admin guard, but then we are forced to set this guard as the default one.

## Proposed solution
Lets make it work that people can set any voyager admin guard - without forcing user to use it as default guard.  This avoids conflicts and is a full fix to #2572,#2534,#2568, #3398, #2002, #4264, #4333

I would propose the following changes:

1. The guard provided in `VoyagerAuth` has to be set (if provided) in VoyagerAuthController so that voyager admin can login even with non-default voyager guard
2. The method `authorize` in  `Voyger/Http/Controllers/Controller` has to be overwritten (by 1 line) in order to make
```
// via controller
$canViewPost = $this->authorize('read', $post);
``` 
work with non-default voyager guard.

Changes 1. & 2. are very easy and I don't see any conflicts with it.

The major change would be, that `@can` could not be used in the view files anymore, because it relates to the default user guard.

So one had to change the docs https://voyager-docs.devdojo.com/core-concepts/roles-and-permissions 

from 

```
// via user object
$canViewPost = $user->can('read', $post);
$canViewPost = Auth::user()->can('read', $post);
```

to 

```
// via user object
$canViewPost = $user->can('read', $post);
$canViewPost = app('VoyagerAuth')->user()->can('read', $post);
```

And in all blades one would need to replace 

`@can(..)`

by 

`@if (app('VoyagerAuth')->user()->can(..)` 

[...]

Issue #3398, opened 2018-07-12
Refreshing login page when default guard is set as `api` 


gileneusz:
## Version information
 - Laravel: v5.6
 - Voyager: v1.1
 - PHP: 7.1.3
 - Database: MySQL 8.0

## Description

I cannot log into fresh Voyager instalation with default guard set as `api` inside config auth.php
Voyager is refreshing login page without any error. This is not occuring when default guard is set as `web` and there is no easy solution how to fix, or even set 'web' guards inside web routes in laravel that serve mainly as an api endpoint. 

## Steps To Reproduce
1. Make fresh instalation of laravel and Voyager
2. Set inside config.auth.php
```
    'defaults' => [
        //'guard' => 'web',
        'guard' => 'api',
        'passwords' => 'users',
    ],
```
3. Try to login as admin
4. See error - refresh admin page

## Expected behavior
Possible login as an admin.

brianchristopherpeck:
Check this out for an explanation of how `'guard' => 'api'` works... https://mattstauffer.com/blog/multiple-authentication-guard-drivers-including-api-in-laravel-5-2/
It sounds like there's something wrong with the way the below steps are implemented:
1.) Add an api_token column to your `users` table. 60-character string, unique.
2.) Instead of using the auth middleware in your route definition, use the auth:api middleware.
3.) In your API routes, use Auth::guard('api')->user() to get your user instead of Auth::user().
I'm pretty sure this isn't a voyager issue

fletch3555:
The issue is that Voyager makes use of the default guard, and does not (currently) provide a way to override it.  When you set the default to "api", a number of middleware are no longer included.

The fix here should be to add a config to Voyager to enable setting what guard to use.  There is another (fairly recent, I believe) issue with a similar guard-related problem, where I suggest the same solution.  Not sure offhand whether there's a PR open yet to add it.  If not, and you'd like to do it, please feel free.

urukalo:
temp solution what i use:
set def guard to `web`
in base api controller(make one if you dont have) constructor override config ( with `Config::set('auth.defaults.guard', 'api');` or similar)

this is ugly solution, it is bad to change config data dynamically (i didnt check impact on caching and other performance related functionalities)  ... but it work for now 😈

emptynick:
Implemented in https://github.com/the-control-group/voyager/pull/3968

dlws21:
[...]

Issue #2572, opened 2018-01-31
Feature Request: Seperating Users from Admins easily.


niels-numbers:
I think it would be useful, if the Voyager Admin would ship with the option to separate users and admins in different tables.

At the moment, separating users and admins is quite some struggle (https://github.com/the-control-group/voyager/issues/2568).

Here are  some reasons why I think it makes sense to allow easily a separation:

1. A user for the application may have a lot of required fields in the DB without default value. In this case, creating a simple admin user would require to make up some random values for the required fields. Further, the command `php artisan voyager:admin your@email.com --create` will not work.

2. Sometimes its nice to have a "Login as User X" feature in the admin dashboard. If admins and users use the same guard, then you constantly need to relogin as an admin, whenever you wanna login as a user.

3. If Voyager does not offer a solution, then one has to use something like `\Auth()->guard('normal')` instead of `\Auth()` in the whole application because the default guard is already used by Voyager.

I could try to create that feature and make a PR, but before I work on that, I would like to know if you guys would be interested in such a feature or if there is a reason why its not there yet.

fletch3555:
There are at least 2 other open issues, a number of closed ones, and an open PR for this.

niels-numbers:
Well thats awesome! Good news! Sorry that I didn't see the PR, its  only 8 days old https://github.com/the-control-group/voyager/pull/2534

fletch3555:
No worries! Appreciate the suggestion.  Someone else just beat you to it ;)

Issue #2568, opened 2018-01-30
How to create seperate user_admin table?


niels-numbers:
- Laravel Version: v5.5.27
- Voyager Version: v1.0.11
- PHP Version: 7.1.7-1+ubuntu16.04
- Database Driver & Version: MySQL 5.7.18-0ubuntu0.16.04.1 - (Ubuntu) 

I want to have users and admin_users in two separate tables.

I found that this has been asked before at https://github.com/the-control-group/voyager/issues/2354 and at https://github.com/the-control-group/voyager/issues/934 but both threads did not work out for me.

This is what I have done:

1. Created a VoyagerAdmin Model

`php artisan make:mode VoyagerAdmin -m`

2. Customized VoyagerAdmin Table

```
public function up()
{
    Schema::create('voyager_users', function (Blueprint $table) {
        $table->increments('id');
        $table->string('name');
        $table->string('email')->unique();
        $table->string('password')->nullable();
        $table->timestamps();
    });
}
```

3. Customized Voyager Migration

Renamed 'users' to 'voyager_users' in `/database/migrations/****_**_**_*****_add_voyager_user_fields.php`

4. Bind at AppServiceProvider

```
use TCG\Voyager\Facades\Voyager;

/**
 * Bootstrap any application services.
 *
 * @return void
 */
public function boot()
{
    // Workaround for Mysql Versions < 7.
    Schema::defaultStringLength(191);

    Voyager::useModel('User', \App\VoyagerUser::class);
}
```
5. Updated config file:

```
    'user' => [
        'add_default_role_on_register' => true,
        'default_role'                 => 'user',
        'namespace'                    => App\VoyagerUser::class,
        'default_avatar'               => 'users/default.png',
    ],

```
6. Updated Database

```
php artisan migrate
php artisan voyager:install
```

I can create a new user with

`php artisan voyager:admin your@email.com --create`

![image](https://user-images.githubusercontent.com/1765602/35562686-68112ac2-05b4-11e8-8c41-9edad30e59aa.png)

However, I cannot login with that. I get the error message that these login credentials could not be found in the database.

Have I done anything wrong?

fletch3555:
Did you check the database? Was your user created in the correct table?

[...]

PR #4336, opened 2019-08-13
[1.x] Allow to have non-default admin guard

### LINKED ISSUE #4262: [BUG]
## Version information
 - Laravel: v5.7
 - Voyager: v1.2.6
 - PHP: 7.2
 - Database: MySql 5.7.24

## Description
403 This action is unauthorized.
I can enter the dasboard but I can not enter any item of the menu, since I get the error.
I think it's because of the authorize function, since I'm using a custom guard as specified in the document.

## Steps To Reproduce
1. Go to Dashboard
2. Click on Users
3. See error

niels-numbers:
Fixes #4262 and #4334.

This is a follow up to #3968, and adds the support for non-default admin guards as requested a couple of times (#2572,#2534,#2568, #3398, #2002).

Changing the default guard driver in `VoyagerAdminMiddleware` makes sure that  

```
// via controller
$canViewPost = $this->authorize('read', $post);
```

and 

    @can

work when the guard is changed to a non-default admin guard. This fixes a bug that when the  guard was changed (possible since #3968) to a non-default guard, it wasn't possible to open any protected page `403 This action is unauthorized.` (#4262) and action buttons did not show up (#4334).

I also created a short description in the docs, how to separate admin and users , since it has also been asked a couple of times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ Made with ❤️ by [The Control Group](https://www.thecontrolgroup.com)
 
 ![Voyager Screenshot](https://s3.amazonaws.com/thecontrolgroup/voyager-screenshot.png)
 
-Website & Documentation: https://laravelvoyager.com
+Website & Documentation: https://voyager.devdojo.com/
 
-Video Tutorial Here: https://laravelvoyager.com/academy/
+Video Tutorial Here: https://voyager.devdojo.com/academy/
 
 Join our Slack chat: https://voyager-slack-invitation.herokuapp.com/
 
--- a/docs/customization/separate_admin_table.md
+++ b/docs/customization/separate_admin_table.md
@@ -1,0 +1,67 @@
+# Separate admin table
+
+Since Voyager 1.2 you can keep your application users and voyager admins in a seperate table.
+
+First, create your customized model, for example `VoyagerUser`, and make sure it extends `\TCG\Voyager\Models\User`:
+
+```php
+class VoyagerUser extends \TCG\Voyager\Models\User
+{
+
+}
+```
+
+Make sure that the columns `name`, `email` and `password` are inside your migration file:
+
+```php
+public function up()
+{
+    Schema::create('voyager_users', function (Blueprint $table) {
+        $table->increments('id');
+        $table->string('name');
+        $table->string('email')->unique();
+        $table->string('password')->nullable();
+        $table->timestamps();
+    });
+}
+```
+
+Register your `VoyagerUser` model in the `boot` method of your `AppServiceProvider`:
+
+    Voyager::useModel('User', \App\VoyagerUser::class);
+
+Create a new voyager [guard](https://laravel.com/docs/5.8/authentication#adding-custom-guards) in `config/auth.php`, for example a guard called `admin`:
+
+```php
+  'guards' => [
+        'web' => [
+            'driver' => 'session',
+            'provider' => 'users',
+        ],
+
+        'admin' => [
+          'driver' => 'session',
+          'provider' => 'voyager',
+        ]
+    ],
+
+  'providers' => [
+        'users' => [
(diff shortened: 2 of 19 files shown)


niels-numbers:
Actually, I just realized that if one changes the default guard in the middleware, then there is no need to use `app('VoyagerAuth')->user()` instead of `Auth::user()` anywhere.

Even though this PR will work fine, its maybe cleaner to revert #3968, and instead of specifying a custom guard in the `AuthServiceProvider` one may rather specify the `guard` in `config/voyager.php` which is then loaded in the middleware. It would also avoid the issue to extract the guard name with regex from the unique identifier for the auth session value.

If you want to avoid to change the default guard in the middleware, we could also just overwrite `authorize` in `Controller.php` and force  to use `@if (app('VoyagerAuth')->user()->can(..)` instead of  `@can`  in all blade files.

What do you think?

emptynick:
We try to avoid config values as much as possible, because they are hard to maintain and upgrade.
Instead, why don't we return a String in the `VoyagerAuth` singleton?

Also, you need to add a link to your docs-file in the [summary.md](https://github.com/the-control-group/voyager/blob/1.2/docs/summary.md)
And don't worry about styling. It will fix itself once its merged.

niels-numbers:
@emptynick sure returning a String in `VoyagerAuth` is just as good. Does it mean you agree with reverting #3968 and changing the guard in the middleware? If so, I would update this PR.

emptynick:
Yes, but there are some more usages of `app('VoyagerAuth')` so you might want to search for it through the whole code.

I think in that case you can bring your documentation to `docs/customization/custom-guard.md `

As this requires users using the old way to change code, we need to push this to 1.x and add it to the upgrade guide.

niels-numbers:
@emptynick in that case, would it not be better to merge the current PR in branch 1.2? Because I believe most people who change the voyager guard have the `web` guard as default.

I could create a new PR for branch 1.x where I implement the discussed changes.

emptynick:
But people who changed the default guard use something like:
```
$this->app->singleton('VoyagerAuth', function () {
    return Auth::guard('your-custom-guard');
});
```
And need to change it to
```
$this->app->singleton('VoyagerAuth', function () {
    return 'your-custom-guard';
});
```
which we can't do in a patch version.

niels-numbers:
@emptynick the current PR assumes that
```
$this->app->singleton('VoyagerAuth', function () {
    return Auth::guard('your-custom-guard');
});
```
is used. The current PR makes it possible to run Voyager when a custom guard was set as above and if it is not set as default (this is currently not possible in 1.2 without this PR). Thus I would recommend to apply this PR in 1.2.

In addition for a PR in 1.2  I could remove the `config` parts and revert #3968 and do some other minor changes. This would make the above work only if its a session guard.

---

We also discussed, that it would be nice to change `VoyagerAuth` to 
```
$this->app->singleton('VoyagerAuth', function () {
    return 'your-custom-guard';
});
```
We cannot due this in 1.2, therefore I would create a seperate PR for 1.x that does this change. This would also work for token guards.

niels-numbers:
I included almost all discussed changes and its fully compatible with Voyager 1.2. I also removed any references to `config`. I did not modify the singleton `VoyagerAuth`!

I could create another commit for 1.x where one may change `VoyagerAuth` to 
```php
$this->app->singleton('VoyagerAuth', function () {
    return 'your-custom-guard';
});
```
If we change that, one would have to update
1. VoyagerAuthController
2. VoyagerAdminMiddleware
3. Update Manual

emptynick:
Take a look here https://github.com/the-control-group/voyager/commit/be3495350d44b2d9125f9ed056eccea676fb2832

niels-numbers:
Since the provided changes here were merged in commit #4384 I close this

PR #4342, opened 2019-08-16
validateBread takes array as an attribute

Makashov:
validateBread must take array as an attribute in  /src/Http/Controllers/Controller.php but not Request object.

Cause it returns  `Validator::make($request, $rules, $messages, $customAttributes);`
And as we know `Validator::make` works with arrays.


src/Http/Controllers/Controller.php
```diff
@@ -157,7 +157,7 @@ public function insertUpdateData($request, $slug, $rows, $data)
     /**
      * Validates bread POST request.
      *
-     * @param \Illuminate\Http\Request $request The Request
+     * @param array                    $request The Request
```
fletch3555: Please rename `$request` to `$data` to avoid confusion.  Otherwise I'm good with this

fletch3555 (dismissed):

emptynick (approved):

PR #4381, opened 2019-09-05
Replaced `title_case` with `Str::title`

AminulBD:
Because the `title_case` is removed from the laravel.


fletch3555:
Thanks, but Voyager doesn't currently support laravel 6, and we're handling those changes in a separate PR

PR #4384, opened 2019-09-06
Revamp custom guards

### LINKED ISSUE #4262: [BUG]

emptynick:
This PR changes the custom guard functionality.
Before, the `VoyagerAuth` singleton returned an instance of a guard.
This singleton was renamed to `VoyagerGuard` and now returns a string (e.g. `web`).
This allows us to simply call `auth()->setDefaultDriver('guard-name');`

I removed the configuration option `voyager.user.namespace` because this was only used to get the user-model which can now be determined on the guard used.

Also
- Added everything related to the upgrade-guide
- Added an example on how to use a different table/model for voyager-admins

Fixes #4262 and closes #4336


fletch3555 (approved):

mohamedadel2018:
i'm don't understand how you solve the problem , i have this problem with my default  admin and can't browse my Beard

emptynick:
This is a PR and isn't about a problem.

Issue #4039, opened 2019-03-26
[BUG] belongsToMany "Allow Tagging" can't create new tags


wishbone-productions:
## Version information
 - Laravel: 5.8.*
 - Voyager: 1.2.*
 - PHP: 7.1.7
 - Database: Maria DB 10.1.25

## Description
When creating a new belongsToMany relationship with "allow taggable" option. Its not possible to create new tags and use them.  The tagging itself works if the tags are already created.

## Steps To Reproduce
1) Create belongsToMany relationships
2) Edit entity, go to tag field
3) Try create new tag.
4) The POST request is sent and the tag itself is created. 
5) The POST returns a list view but not the JSON data needed to create the tag
6) The ajax function is not properly designed to receive JSON data

## Expected behavior
Data is not returning json and is not handled properly

## Hacks to make it work:
Could use something more neat but i`m fairly new to laravel / voyager

JS vendor\tcg\voyager\resources\assets\js\app.js line 102
(insert json parse function)
```
$.post(route, {
	[label]: e.params.args.data.text,
}).done(function(data) {
	data = JSON.parse(data);
	var newOption = new Option(e.params.args.data.text, data.id, false, true);
	$el.append(newOption).trigger('change');
}).fail(function(error) {
	toastr.error(errorMessage);
});
```
PHP code in the Model of the tag
Let the model return the json data on creation
```
public static function boot()
{
	parent::boot();

	self::created(function($model){
		echo json_encode($model);
		exit;
	});
}
```

G00sfraba:
I can still reproduce this in 1.2.6.
Also when you start typing in "taggable" field it looses focus after typing 2 characters. No errors or 404 thrown in console.

If you still manage to click on the new tag it throws this error in the console:
![изображение](https://user-images.githubusercontent.com/1669800/59713678-a369c880-9218-11e9-8dd1-a87f42a933a0.png)

gavinj1999:
Any progress on this please?

GnilitskyWeb:
> Also when you start typing in "taggable" field it looses focus after typing 2 characters. No errors or 404 thrown in console.

This is a bug in select2 on 4.0.3 version
https://github.com/select2/select2/issues/4525#issuecomment-317741989

shafiqkr18:
I have added js code in line 102 & also added Model function, but it is still not working. Also i think it is not good idea to edit app.json. Does anyone fixed this issue?

winex01:
had the same issue

[...]

PR #4397, opened 2019-09-08
Fix relationship-tagging not working

emptynick:
Also fixed that when selecting the "None" option it would just be added to the list instead of removing all selected options.
Works also for the non-taggable-version.

Fixes https://github.com/the-control-group/voyager/issues/4039


fletch3555 (approved):
:+1:

iamriajul:
When we will get the release with this fix?

PR #4403, opened 2019-09-11
add .active even when link has no children

Aldo-f:
When using  {{ menu('main', 'bootstrap') }} the rendered menu doesn't add the .active class.
Now it does. 

image


fletch3555:
Please avoid making whitespace/formatting changes.  It makes it exceptionally difficult to tell what the relevant changes are.

PR #4413, opened 2019-09-15
fix Carbon parse null as current date for date/timestamp fields

### LINKED ISSUE #4412: Voyager incorrectly displays value as current date for date columns
## Version information
 - Laravel: v5.8
 - Voyager: v1.2
 - PHP: 7.2
 - Database: MySQL 5.7

## Description
Voyager incorrectly displays value as current date for `date` columns (due to the behavior of `Carbon::parse()`) when the following conditions are met:

- bread entry has a `format` option
- the value for the column is null 

## Steps To Reproduce
1. Edit the bread for any table that has `date` column (default value for this date column should be NULL)
2. On the bread edit page, set an option to format the date for any of the date fields (e.g. `    "format": "%Y-%m-%d"`) 
3. Go to the Add page for the bread edited above and add a new record. Leave the date-type column formatted in bread above blank
4. Browse the page for the concerned table - the blank date field displays current date.

## Expected behavior
When a database value is null, Voyager should consistently display it as such -- blank.

damms005:
Fixes #4412 

This PR fixes cases whereby Voyager incorrectly displays current date as value for date columns, due to the behavior of Carbon::parse


damms005:
@fletch3555 will this be better if the test is rather for truthfulness (truthy) instead of the current nullity test?

PR #4428, opened 2019-09-21
Media Picker Does Not Support Cloud Storage

OmarRobinson:
Put and get images based on Laravel / Voyager configurations


MrCrayon:
@syclone while not using cloud upload fails because Storage should be used without $realpath, also it would be good if you could apply same fix to the crop method.
Another nitpick you can use `$image->encoded` to get image content as string.

You can check 
https://github.com/MrCrayon/voyager/commit/1a55e5653bfb3d09ce99ef812070bb35fe616d11

MrCrayon:
Closing in favor of #4565 that is based on this one but has been updated.

PR #4429, opened 2019-09-23
Add custom ordering label

arxeiss:
You can set custom ordering label in Order page by setting Model accessor. 

```php
public function getOrderLabelAttribute()
{
    return $this->first_name." ".$this->last_name;
}
```

### DIFF
--- a/resources/views/bread/order.blade.php
+++ b/resources/views/bread/order.blade.php
@@ -29,6 +29,8 @@
                                         <span>
                                             <img src="@if( !filter_var($result->{$display_column}, FILTER_VALIDATE_URL)){{ Voyager::image( $result->{$display_column} ) }}@else{{ $result->{$display_column} }}@endif" style="height:100px">
                                         </span>
+                                    @elseif($result instanceof \Illuminate\Database\Eloquent\Model && $result->hasGetMutator('order_label'))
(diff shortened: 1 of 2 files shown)


resources/views/bread/order.blade.php
```diff
@@ -29,6 +29,8 @@
                                         <span>
                                             <img src="@if( !filter_var($result->{$display_column}, FILTER_VALIDATE_URL)){{ Voyager::image( $result->{$display_column} ) }}@else{{ $result->{$display_column} }}@endif" style="height:100px">
                                         </span>
+                                    @elseif($result instanceof \Illuminate\Database\Eloquent\Model && $result->hasGetMutator('order_label'))
```
MrCrayon, line 32: We shouldn't check a specific accessor name.

MrCrayon (changes requested):
We should add the accessor as option in BREAD configuration
Schermata da 2019-11-20 11-12-31
`$additional_attributes` is what is already used as label option in relationships, explained here [custom-realtionship-attributes.md](https://github.com/the-control-group/voyager/blob/c6dad15f4eff4bad97a85488dcc93cbf795eb2f6/docs/customization/custom-realtionship-attributes.md)

Issue #4398, opened 2019-09-09
ErrorException (E_ERROR) get_class() expects parameter 1 to be object, string given


omarkadd:
## Version information

 - Laravel: v5.8.35
 - Voyager: v1.2.7
 - PHP: 7.3.4
 - Database:MYsql 5.7

## Description

ErrorException (E_ERROR)
get_class() expects parameter 1 to be object, string given (View: /home/u535562063/domains/najdeh-system.com/vendor/tcg/voyager/resources/views/bread/partials/actions.blade.php) (View: /home/u535562063/domains/najdeh-system.com/vendor/tcg/voyager/resources/views/bread/partials/actions.blade.php)
Previous exceptions
get_class() expects parameter 1 to be object, string given (View: /home/u535562063/domains/najdeh-system.com/vendor/tcg/voyager/resources/views/bread/partials/actions.blade.php) (0)
get_class() expects parameter 1 to be object, string given (0)

## Additional context
right after upgrading to 1.2.7

emptynick:
We need a little more information.
This doesn't happen on a default installation. So do you have custom Actions? If yes, how do they look?
Can you post a screenshot of the stacktrace?

omarkadd:
this error happens where i use a custom view nothing fancy i just added a new table in the browse the action.blade is reading from this table .. in the past it worked great after update this happened as for the table it has the same action buttons as the default table

omarkadd:
i downgraded voyager to 1.2.6 for now  till i find a solution to the problem thanks

takamunetanii:
> i downgraded voyager to 1.2.6 for now till i find a solution to the problem thanks

I solved it in the same way.thank you.
In my case, it happens only user table.

impsart:
Also getting this error. I'm using Laravel Wave that relies on voyager. Only package I can think of that might be having conflicts with at the moment.

mattrabe:
For anyone suddenly suffering from this...

This is absolutely related to upgrading from 1.2.6 to 1.2.7. But, **it is more specifically related to using a 1.2.6 bread/browse.blade.php view on top of a 1.2.7 voyager core.** This describes my situation, and it describes the OP's situation.

It also seems to interplay with cache (maybe view cache), so you **may not see this right after an upgrade**. It may pop up some time later, seemingly out of the blue when your cache happens to refresh...

[...]

PR #4436, opened 2019-09-25
Update actions.blade.php to fix get_class fail on string

gruz:
In older Voyager version this code worked for me:

```php
{
...
   public function boot()
    {
... 
            Voyager::replaceAction(DeleteAction::class, TenantDeleteAction::class);
...
     }
...
}
```

But after Voyager update it stopped working as "get_class" fails on string.

### DIFF
--- a/resources/views/bread/partials/actions.blade.php
+++ b/resources/views/bread/partials/actions.blade.php
@@ -1,7 +1,8 @@
 @if($data)
     @php
         // need to recreate object because policy might depend on record data
-        $class = get_class($action);
+        $class = is_object($action) ? get_class($action) : $action;
+        
         $action = new $class($dataType, $data);
     @endphp
     @can ($action->getPolicy(), $data)


MrCrayon:
I think the problem should be solved in replaceAction (and addAction) method making sure a class is added/replaced and not a string.

Checks can be taken from useModel method.

MrCrayon:
I can't reproduce the problem, are you sure you are not using a custom browse view?

gruz:
> I can't reproduce the problem, are you sure you are not using a custom browse view?

Well, I use a custom one. I didn't think about it.

MrCrayon:
In #4151 where I changed actions.blade.php I also made some changes on browse.blade.php, so you should update your custom view to make it work.

mattrabe:
I agree that editing the actions.blade.php file is not the appropriate solution for the get_class error. It does work, but it appears to be a band-aid...

I recently ran into the get_class error, found the following ticket, and just commented my solved experience to: #4398

emptynick:
I think we can close this as it doesn't need to be fixed.
If you encounter this issue please read @mattrabe comment here https://github.com/the-control-group/voyager/issues/4398#issuecomment-537762684

PR #4446, opened 2019-10-03
Link column in browse to edit/show

mattrabe:


MrCrayon:
> Link the first column in browse

I don't see any check that link will be applied only on the first column.

IMHO I don't see this as a useful addition, there are already buttons for view/edit.
It might also cause unwanted false positive for filters if you are using client side pagination.

fletch3555:
I'm inclined to agree with @MrCrayon , though I see the first() restriction.  I would be more inclined to link the entire row, but that has its own issues.  What if the first column is a relationship (which should theoretically link to the related item)?

MrCrayon:
My bad I didn't see the first().
If the first column is a relationship is not going to be linked, only text type is linked.
I still would prefer this to not be implemented, when datatable filters are used they are going to check the whole content so getting also link etc.
Personally I use a forked version with modified browse template so it's not a problem but it might be a problem for others.

mattrabe:
I figured this might not get a warm reception. This comes from a designer that I work with regularly asking for the column containing the Title or Name to be linked to the Edit or Read screen, which I think makes a lot of sense from a UI perspective and is rather common in the wild (e.g. Wordpress. I know, I know - I have nothing nice to say about WP code - but their admin UI is well-trusted). So, I've been manifesting it in various ways via a customized browse views over and over, and figured I would PR it back and see if anyone likes the idea. Maintaining divergent BREAD views [can also problematic over time](https://github.com/the-control-group/voyager/issues/4398).

I won't be offended if this is not merged. However, it should be noted that it seems to work fine with the client-side datatables. I can't explain why - maybe datatables is now smart enough to strip the tags (how hard is that really!?). Below is a video showing it working. Note the URLs of each of the links on mouseover. The IDs of the first three rows are (in order) 2, 22, 14, while Home further down the list is 1. You will see that even though each of these is wrapped in a link with `href="/admin/pages/{id}/edit"` that they are NOT reordered to based on those links when datatables is reordered: [Screen-Recording-2019-10-03-at-6.00.05-AM.mov](http://flowerpress.xyz/tmp/Screen-Recording-2019-10-03-at-6.00.05-AM.mov)

So, not sure how that works, but it evidently does... Would be good to know why

MrCrayon:
I think the problem with datatable is with the filter,  when you try to search something, but I can't check right now.

I was thinking if we use a parameter in BREAD details to set when a field should become a link people would be able to choose if using this feature or not, also we could choose to link a field different from the first one. 
Something like: 
```json
{
    "link": true
}
```

mattrabe:
Searching seems to work fine as well: [Screen-Recording-2019-10-03-at-6.44.13-AM.mov](http://flowerpress.xyz/tmp/Screen-Recording-2019-10-03-at-6.44.13-AM.mov)

I agree with @MrCrayon regarding this feature being explicitly specifiable. That is what I originally thought might kill this PR... BREAD details param would probably be a good place to do it. We may want to have `editLink` and `readLink` options to allow specifying what it links to

MrCrayon:
Yes, searching is working I think maybe I had problems when I developed custom filters, I don't remember.

As for the parameter maybe it could be link with value edit or show and after checking for permisions we can just use:
```php
route('voyager.' .$dataType->slug. '.'.$row->details->link)
```

mattrabe:
@MrCrayon I really like that approach. Unfortunately, the value for the Read screen permission and the Read screen view are not the same. Permission is `read`, while view is `show`. So I think to do this right, `link` should be an object that maps permissions to views:

```
{
  "link": {
    "edit": "edit",
    "read": "show",
    "another_permission": "some-view",
  },
}
```

mattrabe:
`details` parameter added as above. Much better.

resources/views/bread/browse.blade.php
```diff
@@ -178,7 +178,23 @@
...
+                                                        $view = null;
+                                                        if (property_exists($row->details, 'link')) {
+                                                            $view = collect($row->details->link)->first(function($view, $permission) use ($dataType) {
+                                                                return Auth::user()->can($permission, app($dataType->model_name));
+                                                            });
+                                                        }
+                                                    @endphp
+
+                                                    @if ($view)
+                                                        <a href="{{ route('voyager.' .$dataType->slug. '.' .$view, [ 'id' => $data->id ]) }}">{{ $content }}</a>
```
MrCrayon: To be ready for Laravel 6 route should be:
```php
route('voyager.'.$dataType->slug.'.'.$view, $data->id)
```
Edit: removed also spaces

MrCrayon (changes requested):
I would use $link instead of view, but matter of preference.
Except route to correct everything else looks good but I didn't test it yet.

resources/views/bread/browse.blade.php
mattrabe: done

MrCrayon: I think documentation needs to be updated to explain this new parameter.
Probably add a voyager/docs/bread/formfields/text.md like number.md

Another thing that needs to be changed that I didn't notice before is the permission check, you should use $data instead of $dataType because, depending on policy, permissions can change from one element to another.

As example in a standard new install if you give a user only browse permission for users BREAD, that user will be able to browse users and edit and view itself but not other users.
Using $data instead of $dataType gets the same behaviour.
Schermata da 2019-10-05 10-18-29

P.S.
I would like to clarify that I'm a contributor just like you, so whatever I say is just my opinion don't feel compelled to implement it, you might want to wait @fletch3555 
 or @emptynick feedback before working on changes.

MrCrayon:
```php
$link = collect($row->details->link)->first(function($view, $permission) use ($data) {
    return Auth::user()->can($permission, $data);
});
```

mattrabe: @MrCrayon hmmmm, `$data` is undefined in the blade and the Controller. Did you mean a different variable?

Doc added for text formfield

MrCrayon:
@mattrabe yup it's $data

mattrabe:
@MrCrayon duh, `use ($data)`... 🤓 Thanks

MrCrayon:
Can you rebase and change the target to 1.4?

mattrabe:
@MrCrayon done. I merged in the latest from 1.4 as well

PR #4453, opened 2019-10-07
Redirect after saving based on permissions

### LINKED ISSUE #2625: Permissions error when user can add/edit but not browse

MrCrayon:
If no browser permission saving will redirect to the same page.

One common use case scenario is while a user without browse_users permission saves his profile.

Fixes #2625


emptynick:
Would love to see a test for this with someone who hasn't the permission to browse.

MrCrayon:
I'll try to add it in a few days.

MrCrayon:
@emptynick
I added the test inside UserProfileTest since there is already a test for saving with browse permission there and is also the place where the "no browse permission" problem is more likely to happen.

emptynick (approved):

PR #4455, opened 2019-10-07
Cache check permission 

MrCrayon:
Cache check permission in basePolicy for CACHE_TIME seconds.

This cache last only for few seconds so there is no need to destroy cache after changing permissions or roles, it's mostly useful to avoid repeating the same query to check permission in browse view.

In my testing in voyager with dummy, browse Post, that has 3 actions this PR makes page generation 16% faster.

Of course with server side pagination the difference in time is negligible but noticeable with client side  pagination and many records.

### DIFF
--- a/src/Policies/BasePolicy.php
+++ b/src/Policies/BasePolicy.php
@@ -3,7 +3,6 @@
 namespace TCG\Voyager\Policies;
 
 use Illuminate\Auth\Access\HandlesAuthorization;
-use Illuminate\Support\Facades\Cache;
 use TCG\Voyager\Contracts\User;
 use TCG\Voyager\Facades\Voyager;
 
@@ -13,7 +12,7 @@ class BasePolicy
 
     protected static $datatypes = [];
 
-    const CACHE_TIME = 5;
+    protected static $cache = [];
 
     /**
      * Handle all requested permission checks.
@@ -78,15 +77,18 @@ class BasePolicy
      */
     protected function checkPermission(User $user, $model, $action)
     {
-        return Cache::remember("{$user->getTable()}-{$user->id}-can-{$action}-{$model->getTable()}", self::CACHE_TIME, function () use ($user, $model, $action) {
+        $key = "{$user->getTable()}-{$user->id}-can-{$action}-{$model->getTable()}";
+
+        if (!isset(self::$cache[$key])) {
             if (!isset(self::$datatypes[get_class($model)])) {
                 $dataType = Voyager::model('DataType');
                 self::$datatypes[get_class($model)] = $dataType->where('model_name', get_class($model))->first();
             }
 
             $dataType = self::$datatypes[get_class($model)];
+            self::$cache[$key] = $user->hasPermission($action.'_'.$dataType->name);
+        }
 
-            return $user->hasPermission($action.'_'.$dataType->name);
-        });
+        return self::$cache[$key];
     }
 }
--- a/src/Policies/BasePolicy.php
+++ b/src/Policies/BasePolicy.php
@@ -3,6 +3,7 @@
 namespace TCG\Voyager\Policies;
 
 use Illuminate\Auth\Access\HandlesAuthorization;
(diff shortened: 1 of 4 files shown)


emptynick:
How did you measure page generation?
I'm getting longer load-times in both debugbar and Chrome inspection tools by a few mills.

MrCrayon:
1. Installed voyager with dummies
2. Removed unique key from posts.slug (just to make next step easier)
3. Duplicated post until reaching 1000 posts
4. Added to AppServiceProvider boot method
```php
config(['timer' => microtime()]);
```
5. Extendend partials/app-footer.blade.php
```blade
@php
    function pageTracker()
    {
        list($usec, $sec) = explode(" ", config('timer'));
        $time_start = (float)$usec + (float)$sec;
        
        list($usec, $sec) = explode(" ", microtime());
        $time_end = (float)$usec + (float)$sec;

        return 'Generated in ' . number_format($time_end - $time_start, 3).'s';
    }
@endphp
<footer class="app-footer">
    <div class="site-footer-right">
        @if (rand(1,100) == 100)
            <i class="voyager-rum-1"></i> {{ __('voyager::theme.footer_copyright2') }}
        @else
            {!! __('voyager::theme.footer_copyright') !!} <a href="http://thecontrolgroup.com" target="_blank">The Control Group</a>
        @endif
        @php $version = Voyager::getVersion(); @endphp
        @if (!empty($version))
            - {{ $version }}
        @endif
        @if(env('APP_DEBUG'))
         - {!! '<b>'.pageTracker().'</b>' !!}
        @endif
    </div>
</footer>
```
(I have that function in helpers.php but for testing works there too)
6. Access posts index page multiple times with and without cache enabled and log it.

I just tried, there is a bit of variance because I didn't close other programs but it's consistently faster with cache enabled.

Schermata da 2019-10-10 10-31-24

This with the three standard permission checks.

MrCrayon:
I had an idea... since we are not caching it for other requests we could just use an array, same as is used for $dataTypes.

```php
    protected static $cache = []; 
 
    protected function checkPermission(User $user, $model, $action)
    {
        $key = "{$user->getTable()}-{$user->id}-can-{$action}-{$model->getTable()}";

        if (!isset(self::$cache[$key])) {
            if (!isset(self::$datatypes[get_class($model)])) {
                $dataType = Voyager::model('DataType');
                self::$datatypes[get_class($model)] = $dataType->where('model_name', get_class($model))->first();
            }

            $dataType = self::$datatypes[get_class($model)];
            self::$cache[$key] = $user->hasPermission($action.'_'.$dataType->name);
        }

        return self::$cache[$key];
    }
```

This way is even quicker
Schermata da 2019-10-10 10-55-23

emptynick:
That looks even better (and a bit simpler)

MrCrayon:
@emptynick there is a problem with tests, with this PR permissions are kept in memory so when we test:
```php
    /**
     * UserDimmer widget isn't displayed without the right permissions.
     */
    public function testUserDimmerWidgetIsNotShownWithoutTheRightPermissions()
    {
        // We must first login and visit the dashboard page.
        $user = \Auth::loginUsingId(1);

        // Remove `browse_users` permission
        $user->role->permissions()->detach(
            $user->role->permissions()->where('key', 'browse_users')->first()
        );
```
Permission is not really removed because it was cached from the previous test, so even if in normal use it's working the test fails.

I think we need a method in BasePolicy to purge cache to be called before any test that involves permissions, something like:
```php
public static function purgeCache()
{
    self::$cache = [];
}
```

Thoughts?

fletch3555:
Is the policy the best place to store this? I feel like we should have a "PermissionCache" service class (or whatever you want to call it), which handles fetching/caching/purging permissions, thus cleaning up the policy significantly.

MrCrayon:
Yes, totally.
I went for the minimum change but I can refactor.

MrCrayon:
I'm not working on this anymore since I completely refactored caching in https://github.com/MrCrayon/voyager/tree/refactor-caching

PR #4460, opened 2019-10-08
Re #4459: Datatable ordering broken on browse with no checkboxColumn

### LINKED ISSUE #4459: Datatable ordering broken on browse with no checkboxColumn
## Version information

 - Laravel: v5.8.35
 - Voyager: v1.2.7
 - PHP: 7.2.8
 - Database: MySQL 5.7.8

## Description

Pull request #4151 introduced the setting of a new variable named `$showCheckboxColumn`, which is set within bread.browse view. If this var is `false`, then the first column of checkboxes is not displayed. This works well and is nice. However, before this variable is set in the view, the Controller assumes that the checkbox column is present, by adding `+ 1` to the index of the `orderColumn` at VoyagerBaseController.php::64. This means that when the Delete & Mass Delete actions are not present for a user, their client-side sorting will always be incorrect, since datatables will be trying to sort on the wrong column.

## Steps To Reproduce
1. Set the `order_column` for a DataType. Make sure that it also has the `delete` permission turned **ON** for your user role.
1. View the BREAD browse in a browser. Observe that ordering is correct, and the `order_column` you specified is "selected" by default.
1. Now turn **OFF** the `delete` permission for your user role on that DataType.
1. Look at the BREAD browse in a browser again.
1. *Observe that ordering is incorrectly happening on the column AFTER the one you specified as `order_column`.*

## Expected behavior

Ordering should work on `order_column` independent of `delete` permission

## Screenshots

Screen Shot 2019-10-07 at 6 02 55 PM

mattrabe:
Fix by moving $showCheckboxColumn def into Controller, reorder a couple things, and only add `+ 1` to $index for $orderColumn if $showCheckboxColumn

Fixes #4459


emptynick:
Tagging @MrCrayon as he wrote #4151

MrCrayon:
I see checkbox column was never taken in consideration, so it's quite an old bug.

@emptynick 
PR looks ok to me, not sure why it's failing:
TCG\Voyager\Tests\CustomGuardTest::testVoyagerControllersUseCustomGuard

Another solution would be to add something like this in the template:
```blade
        @if(!$showCheckboxColumn)
            @php $orderColumn[0][0]--; @endphp
        @endif
```
Smaller change but a bit hacky, @mattrabe solution looks cleaner overall.

@mattrabe
I think these two lines shouldn't be moved:
```php
        $usesSoftDeletes = false;
        $showSoftDeleted = false;
```

mattrabe:
I see why this is failing StyleCI currently, and will fix. I also do not understand why it's failing TravisCI. The PR I made the other day also failed TravisCI at first but then passed it after a subsequent commit (which didn't contain any change to intentionally pass Travis, so not sure why...)

@MrCrayon Yea, I moved those two vars at first because they were actually created by me some time ago, but I felt I put them in the "wrong" place. However, after the reordering I did that was required because of the various blocks depending on eachother, they really didn't have to be moved after all.

Will post another commit in a little bit

emptynick:
I just changed your branch to 1.x which _should_ fix Travis.
I think in 1.2 it interferes with custom guards, while 1.x uses "classic" `Auth::user()`
Edit: Yes it does fix Travis.

emptynick (approved):

PR #4465, opened 2019-10-09
Laravel 6.1/6.2 support

emptynick:
This lets you install Voyager on top of Laravel 6.1 and 6.2.
No other steps are necessary.


composer.json
```diff
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "illuminate/support": "~5.5.0|~5.6.0|~5.7.0|~5.8.0|~6.0.0",
+        "illuminate/support": "~5.5.0|~5.6.0|~5.7.0|~5.8.0|~6.0.0|~6.1.0|~6.2.0",
```
fletch3555: Can we just switch this to `"~5.5|~6.0"`?  Leaving off the 3rd number should allow it to match "5.*" and "6.*"

.travis.yml
```diff
@@ -14,6 +14,7 @@ env:
     - LARAVEL_VERSION=5.7.*
     - LARAVEL_VERSION=5.8.*
     - LARAVEL_VERSION=6.0.*
+    - LARAVEL_VERSION=6.1.*
```
fletch3555: Do we care about testing all the minor versions of 6? Now that they're following semver, we should be fine with just `6.*`

emptynick: True, forgot that.

emptynick:
All done

fletch3555 (approved):

PR #4488, opened 2019-10-18
Add optional helperText for formfields

mattrabe:
Also fix edit-add label `for` attribute


MrCrayon:
@mattrabe doesn't really fix the _for_ attribute because _for_ should contain _id_ element and we don't have id currently.
Adding id would fix it but I would be careful to not interfere with ids already used either in voyager (not sure if any) or in custom views already developed.

Adding a prefix would be a good measure I think, like vfe short for voyager form element.
```blade
<label class="control-label" for="vfe_{{ $row->field }}">
```
```blade
<input id="vfe_{{ $row->field }}">
```

mattrabe:
this branch got botched - will fix and re-open...

PR #4492, opened 2019-10-19
Add optional helperText for formfields

mattrabe:
- Use Bootstrap form-text class (and text-muted for styling) …
- Add aria-described by to all applicable inputs


mattrabe:
@MrCrayon This is the redo of pull request #4488. Not touching the `for` attribute on this one, and added bootstrap classes & aria

src/Helpers/helperHelperText.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+if (!function_exists('outputAriaForHelpterText')) {
```
fletch3555: "Helpter" ...? I'm assuming that's a typo (you have it everywhere it's called too)

fletch3555 (changes requested):

src/Helpers/helperHelperText.php
mattrabe: yup, fixed

mattrabe:
@emptynick What do you guys think of merging this one? The requested change was fixed some time ago. I use this in production. It's a feature addition that doesn't change any existing functionality.

emptynick:
Hey, sorry for the delay.
Can't we simply include that text directly under the formfield in the edit/add view so it doesn't need to be repeated for every formfield?

mattrabe:
@emptynick Yes, and this PR does exactly that. But the ARIA tag for accessibility is an attribute on the input/select/etc itself, so that has to be per-formfield. The aria attribute I’m using is `aria-describedby` and it only references the name of the single helper text `p` in add-edit (it does not repeat the actual helper text).

PR #4503, opened 2019-10-24
No sense to restrict menu Json output

gbesoain:
There is no sense to restrict json input to menu name "admin" only

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -73,7 +73,7 @@ class Menu extends Model
 
         $items = $menu->parent_items->sortBy('order');
 
-        if ($menuName == 'admin' && $type == '_json') {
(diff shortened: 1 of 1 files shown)


fletch3555:
While I don't necessarily disagree, could you please describe the reasoning behind this change?

gbesoain:
I needed to create new admin menus (with different names off course) and wanted them as JSON output, my local solution was to override Voyager Menu Model.

But the question should be, what is the reason behind the restriction of JSON output only to menu named "admin"?

Regards

MrCrayon:
@gbesoain I think to not break compatibility and because type is used for too many things already we could use $options.
`$options->processItems` could be used to check if items should be processed, and it can be set as true as default if type is admin.

PR #4510, opened 2019-10-28
Update the way of generating breadcrumbs segments.

hassanjuniedi:
Using Request::path() instead of Request::url() will give us only the used portion of URI for generating segments.
Also using route('voyager.dashboard',[], false) will return relative route.
The whole idea is to avoid conflicts in URL Schema (https, http) and that when using nginx proxy_pass from https to http. which in the end will break the breadcrumbs generating.


alexisrazok:
The bug fixes also using url()->current() instead of Request::url().

alexisrazok (approved):
The error is fixed with this solution, also it is possible to fix it using url()->current()

PR #4538, opened 2019-11-05
Support for table prefix

### LINKED ISSUE #1014: Installation with existing database?
How does one go about using this package with an existing app/database. Seems to be only good for a brand new database.

?

MrCrayon:
Use --with-prefix=prefix_ to install all voyager tables with a prefix. (Need to add documentation)

Running the whole test suite with prefix gives no error but I can't guaranty that everything works so I would tag the functionality as experimental.

- Change all voyager tables with the use of helper get_prefix_table
- Left unchanged slugs because that will affect also templates
- Prefixable Trait applies prefix to table name
- Added a test to verify install with prefix works

Closes #1014


migrations/2016_01_01_000000_add_voyager_user_fields.php
```diff
@@ -9,8 +9,8 @@ class AddVoyagerUserFields extends Migration
      */
     public function up()
     {
-        Schema::table('users', function ($table) {
-            if (!Schema::hasColumn('users', 'avatar')) {
+        Schema::table(get_prefixed_table('users'), function ($table) {
```
fletch3555, line 12: We **_really_** shouldn't be changing old migrations... Works well for new users, but could very easily burn existing users trying to upgrade.

MrCrayon: I understand your concern but I don't think we would be able to add prefix support without changing old migrations.

If that can help whenever you are updating or rolling back `get_prefix_table` helper will output the same table name passed if voyager.database.table_prefix is not set (as it should for previous installations) or empty (default value for new installations).

So, unless I'm missing something, it shouldn't change anything for old users unless they willingly add a table_prefix.

sergiogregorutti:
This is a nice feature. What's the status of this? Can we use it on new projects?

emptynick:

minutmorph:
Hi @emptynick , is there an expected release date for V2? thanks

cyberfab007:
Why would such a feature sit on a shelf for almost 2 years ?

PR #4545, opened 2019-11-07
Fix escaping in templates

MrCrayon:
Remove unnecessary ```{{ }}``` with string inside

Power of regex \{\{\s\'([^\}]+)?\'\s\}\}

No idea why there are so many double if to output the same thing, but I left them there for now.
```blade
<input type="radio" id="option-{{ $index }}" name="{{ $setting->key }}" value="{{ $index }}"
@if($default == $index && $selected_value === NULL) checked @endif 
@if($selected_value == $index) checked @endif>
```

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -49,14 +49,14 @@
                                     <div class="col-2">
                                         <select id="search_key" name="key">
                                             @foreach($searchNames as $key => $name)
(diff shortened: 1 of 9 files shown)


fletch3555:
That double-if looks like a find/replace error.... feel free to consolidate them(though probably in a separate PR)

MrCrayon:
Yes, I didn't want to make that kind of change in this one.
I think I'll wait for this to be merged just to avoid the conflicts.

emptynick (approved):

Issue #3185, opened 2018-05-16
widget not formatted on 4th widget


amc123-glitch:
"php": "^7.1.3",
"fideloper/proxy": "^4.0",
"laravel/framework": "5.6.*",
"laravel/tinker": "^1.0",
"tcg/voyager": "^1.1"

### Description:
I want to add 4 widgets to the dashboard. I can successfully get the third widget to show in format, but when I add another widget after that it is placed far right, under the third widget instead of being pushed to the left. In addition, If I add more widgets or even just copy and paste the same two new widgets, they show up formatted from left to right as expected.

The first 2 widgets are the ones generated by the --with-dummy data tag. In addition the widgets I made are stored in my App directory in a folder labeled widgets and then referenced as App\\Widgets\\somefilename.

### Steps To Reproduce:
- create a folder labeled widgets
- create php file for widget
- copy and paste widget code from the dummy generated in the vendor/tcg/src/widgets section and modify for whatever the widget is showing
- php artisan config:cache and refresh

hemantachhami19:
I am also having the same problem!!
New Widgets doent get formed ,nor removed while modifying voyager.php

amc123-glitch:
@hemantachhami19 I opted to build my own backend admin panels, partially for this reason and since the team decided to let this issue just sit answered

Spraynard:
I have had the same problem with my layout, and I know the reason for this behavior.

The blade template `dimmers.blade.php` is used to output all of the `Dimmers` i.e. `Widgets`, in voyager context, to HTML. If you look at the template, you can see that it is quite specifically made to only support up to _three_ different widgets (if the count sees two widgets, we will output different classnames to our widgets than if there are three)

This is done by calling the `Voyager::dimmers()` function, shown below

```
    /**
     * Get a collection of the dashboard widgets.
     *
     * @return \Arrilot\Widgets\WidgetGroup
     */
    public function dimmers()
    {
        $widgetClasses = config('voyager.dashboard.widgets');
        $dimmers = Widget::group('voyager::dimmers');

        foreach ($widgetClasses as $widgetClass) {
            $widget = app($widgetClass);

            if ($widget->shouldBeDisplayed()) {
                $dimmers->addWidget($widgetClass);
            }
        }

        return $dimmers;
    }
```

[...]

PR #4559, opened 2019-11-14
Enhance Widget Layout

Spraynard:
This update contains two changes to two different files.

I had an issue similar to Issue #3185 and was able to implement a solution that works along with the current voyager configuration but allow for more than three widgets to be on the dashboard window.

This is what the dashboard window looks like with 1 - 3 Widgets

voyager_widgets_1

This is what the dashboard window looks like with 5 widgets

voyager_widgets_2

Please let me know if there is any more information needed


tests/Unit/VoyagerTest.php
```diff
@@ -42,6 +42,27 @@ public function testDimmersReturnsCollectionOfConfiguredWidgetsWhichShouldBeDisp
...
+        $this->assertEquals(1, $dimmers[0]->count());
+    }
+
+    /**
+     * Dimmers returns an array filled with widget collections.
+     *
+     * This test ensures that each widget collection has a maximum of three widgets/dimmers
+     * creating as many groups as are needed in order to encompass the full range of dimmers.
+     */
+    public function testEachDimmerGroupHasAMaxAmountOfThreeDimmers()
```
MrCrayon: If you want to test that Group has max amount of three Dimmers you should set 3 dimmers in config and assert equals 1, then set 4 dimmers and assert equals 2.

Right now you could have 4 dimmers in the first group without error.

MrCrayon (changes requested):
Except Test everything else looks good. :+1:

tests/Unit/VoyagerTest.php
Spraynard: Thank you for the guidance. I'll commit an updated VoyagerTest.php

tests/Unit/VoyagerTest.php
```diff
@@ -64,5 +83,7 @@ public function testEachDimmerGroupHasAMaxAmountOfThreeDimmers()
         $dimmers = Voyager::dimmers();
 
         $this->assertEquals(2, count($dimmers));
+        $this->assertEquals(3, $dimmers[0]->count());
```
MrCrayon, line 67: This assertion proves what you were trying to prove in the previous test, so I think the other one can be removed.

MrCrayon (changes requested):
I think you can remove the first test.

tests/Unit/VoyagerTest.php
```diff
@@ -42,6 +42,48 @@ public function testDimmersReturnsCollectionOfConfiguredWidgetsWhichShouldBeDisp
...
-        $this->assertEquals(1, $dimmers->count());
+        $this->assertEquals(1, $dimmers[0]->count());
+    }
+
+    /**
+     * Dimmers returns an array filled with widget collections.
+     *
+     * This test ensures that each widget collection can hold up to three widgets
+     */
+    public function testDimmerCollectionContainsThreeDimmersMax()
```
MrCrayon: How can you test that collection contains three dimmers max if you are only passing three? :thinking:

MrCrayon (changes requested):
This test can be removed.

MrCrayon (approved):
Looks good, thanks for this PR

Spraynard:
Thank you very much for the help as well. It's much appreciated.

PR #4563, opened 2019-11-15
Fix soft delete

### LINKED ISSUE #4540: SoftDeletes Restore button conditional on browse view no longer per-row
## Version information

 - Laravel: v5.8.35
 - Voyager: v1.3.0
 - PHP: 7.2.8
 - Database: MySQL

## Description

Restore button does not show correctly on Browse view for SoftDeletes models. Conditional for whether Restore button shows is based only the the first row, and is applied to all rows (instead of checking per-row). Easily explained by a code change (detailed below)

## Steps To Reproduce
1. Create a SoftDeletes model, add at least 3 rows to the model
1. Go to Browse view for that model
1. Delete the SECOND row
1. Click "Show Deleted" toggle so that you are viewing soft-deleted records
1. **Observe that Restore button does NOT show for any row**
1. Now delete the FIRST row.
1. Click "Show Deleted" toggle so that you are viewing soft-deleted records
1. **Observe that Restore button shows to ALL rows, even though only the first 2 were deleted**

## Expected behavior

Restore button should only show for rows with a non-null `deleted_at`, Delete button should show for the inverse.

## Additional context

This appears to have been caused by moving `shouldActionDisplayOnDataType` in 3b461e9f2d7f4821fcbb020aaaa72a5ccbd6d858. When that function was running in the view partial, it was called for each row, but now in the Controller it is called only once for the whole model, and `$dataTypeContent->first()` is explicitly passed. RestoreAction depends on being able to check for `$this->data->deleted_at` per row: https://github.com/the-control-group/voyager/blob/1.3/src/Actions/RestoreAction.php

If we want to run this in the Controller (which seems sensible for more than one reason), we may need to pass the $actions result as a "row" on $dataTypeContent.

I'd be glad to make a patch for this since it will very soon affect a client project-in-progress, but would love to hear from @MrCrayon since he created pull request #4151 which includes the above commit, as well as @emptynick @fletch3555 . Any initial thoughts on my suggested approach or approach in general?

### LINKED ISSUE #4560: SoftDeletes cannot be recognized if inherited
## Version information

 - Laravel: v6.5.1
 - Voyager: v1.3.0
 - PHP: 7.2.12
 - Database: MariaDB 10.1.37

## Description

If SoftDeletes is defined in a parent model, voyager didn't recognize it using a child model

## Steps To Reproduce
1. Create a Model that use SoftDeletes
2. Create a Model that extends the previous model
3. Create a BREAD view for a table using the child Model
4. Browse the table, Show/Hide Delete button will not compare

## Expected behavior
If i define use SoftDeletes on a parent model i'll aspect that it will recognized also in the child model.

## Additional context
I think that the problem is in the TCG\Voyager\Http\Controllers\VoyagerBaseController.php, because in all the check for SoftDeletes is used the method class_uses that search only in the specific model, instead of class_uses_recursive that search also in parent model

MrCrayon:
Use class_uses_recursive to recognize if parent Class uses SoftDeletes.
Changed BasePolicy so that it can take care of checking if record can be deleted or restored based also on use of SoftDeletes.

More details on the second change in the issue fixed.

Fixes #4540
Fixes #4560


mattrabe (dismissed):
I have tested this fix and it works on my end. Nice fix

emptynick (approved):

humayunjavaid:
> I have tested this fix and it works on my end. Nice fix

How can i get this fix on my local machine?

mattrabe:
@8humayun8 https://stackoverflow.com/questions/33525885/composer-require-branch-name/52366093

MrCrayon:
What @mattrabe said is right, but I wouldn't advice to do it right now because assets are not compiled. 
I think this will be in next version hopefully to be released this week :crossed_fingers: 

@fletch3555  can you merge this one?

humayunjavaid:
> @8humayun8 https://stackoverflow.com/questions/33525885/composer-require-branch-name/52366093

Sorry i am newbie in git. i am unable to get branch name . can you please tell be the exact command. Thanks

fletch3555 (approved):

mrdeeds88:
I got this error after success restored a post. Restore function works nicely on other models.
image

How could I fix this?

PR #4569, opened 2019-11-18
Fix loading assets

MrCrayon:
We need to use __DIR__ to know where we are for tests to succeed.

### DIFF
--- a/src/Http/Controllers/VoyagerController.php
+++ b/src/Http/Controllers/VoyagerController.php
@@ -75,8 +75,13 @@ class VoyagerController extends Controller
 
     public function assets(Request $request)
     {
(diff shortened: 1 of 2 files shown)


emptynick (approved):

MrCrayon:
@lanhhuyet510 if you have any suggestion you can find me in Voyager Slack channel.
I know a whitelist would be better but it's not feasibile right now.

Thanks

lanhhuyet510:
> @lanhhuyet510 if you have any suggestion you can find me in Voyager Slack channel.
> I know a whitelist would be better but it's not feasibile right now.
> 
> Thanks

Noted that and confirmed fixed loading assets in v1.3.1
Thanks.

PR #4570, opened 2019-11-18
Adding button at media picker to add all files selected to input

gustavodaltoe:
Adds new button at media picker header to include all selected files at the input.
Screenshot (15)

### DIFF
--- a/resources/views/media/manager.blade.php
+++ b/resources/views/media/manager.blade.php
@@ -55,6 +55,11 @@
             <i class="voyager-refresh"></i>
         </button>
         <div class="btn-group offset-right">
+            <button type="button" class="btn btn-default" id="addAll" v-if="allowUpload" v-on:click="addAllFilesToInput();" 
+            :disabled="selected_files.length < 1 || !fileIs(selected_file, 'image')">
+                <i class="voyager-check"></i>
+                Add Selected
+            </button>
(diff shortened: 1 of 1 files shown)


resources/views/media/manager.blade.php
```diff
@@ -55,6 +55,11 @@
             <i class="voyager-refresh"></i>
         </button>
         <div class="btn-group offset-right">
+            <button type="button" class="btn btn-default" id="addAll" v-if="allowUpload" v-on:click="addAllFilesToInput();"
```
emptynick, line 58: Thanks for your PR.
We need to add a check if the media-manager is used as a media-picker.
Otherwise it would show the button when you normally browse media and might fail if you click it.
You can simply add `v-if="hidden_element"`

resources/views/media/manager.blade.php
```diff
@@ -55,6 +55,11 @@
             <i class="voyager-refresh"></i>
         </button>
         <div class="btn-group offset-right">
+            <button type="button" class="btn btn-default" id="addAll" v-if="allowUpload" v-on:click="addAllFilesToInput();" 
+            :disabled="selected_files.length < 1 || !fileIs(selected_file, 'image')">
+                <i class="voyager-check"></i>
+                Add Selected
```
emptynick, line 61: This needs to be translated.

emptynick (changes requested):

emptynick:
Superseded by https://github.com/the-control-group/voyager/pull/4821

PR #2604, opened 2018-02-03
WIP: Add migration feature

KABBOUCHI:
Create migration when create/update/delete table

TODO:
- Create migration for BREAD actions
- ~~Insert current migration history~~
- Enable database tool only for local env (optional)
- Tests
- Refactor and extract common code

 Any thought before I continue?

PR #4573, opened 2019-11-21
[doc] Add a note about migration is not managed

Cyrille37:
Please, review my so poor english.
Cheers.

Reference:
- https://github.com/the-control-group/voyager/pull/2604


Engr-Rowshan (commented):
This note should be mentioned into the documentation that Voyager not supporting automatic migration generation till now. It will help other to find out why there development version not working in production even after running php artisan migrate.

PR #4577, opened 2019-11-24
Prevent "SQLite doesn't support dropping foreign keys" error.

mchekin:
Since Laravel 5.7 using Blueprint::dropForeign() with SQLite throws an exception.

From https://laravel.com/docs/5.7/upgrade:
> SQLite does not support dropping foreign keys. For that reason, using the dropForeign method on a table now throws an exception. Generally, this should be considered a bug fix; however, it is listed as a breaking change out of caution.

> If you run your migrations on multiple types of databases, consider using  DB::getDriverName() in your migrations to skip unsupported foreign key methods for SQLite.

This causes failure when using Voyager with SQLite in migrations/2017_11_26_013050_add_user_role_relationship.php

**Currently you cannot use Voyager with SQLite** (Laravel > 5.7)

### DIFF
--- a/migrations/2017_11_26_013050_add_user_role_relationship.php
+++ b/migrations/2017_11_26_013050_add_user_role_relationship.php
@@ -2,6 +2,7 @@
 
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
 
 class AddUserRoleRelationship extends Migration
@@ -27,7 +28,9 @@ class AddUserRoleRelationship extends Migration
     public function down()
     {
         Schema::table('users', function (Blueprint $table) {
-            $table->dropForeign(['role_id']);
+            if (DB::getDriverName() !== 'sqlite') {
+                $table->dropForeign(['role_id']);
+            }
         });
 
         Schema::table('users', function (Blueprint $table) {


MrCrayon:
I wonder why there is no problem in test enviroment that uses sqlite.
https://github.com/the-control-group/voyager/blob/0306d7348f3dd05a4ea19c66f18bc9986bbe0902/tests/TestCase.php#L63

Does it still cause Exception if you set `DB_FOREIGN_KEYS=true`?

mchekin:
@MrCrayon 

Using 

> DB_FOREIGN_KEYS=true 

won't help since it's not being used when throwing the exception:
https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Schema/Blueprint.php#L149

I suspect it does not fail during the test since the migrate:reset command in commented out during the test tearDown():

https://github.com/the-control-group/voyager/blob/0306d7348f3dd05a4ea19c66f18bc9986bbe0902/tests/TestCase.php#L48

MrCrayon:
Your suggestion, based on what documentation says, makes sense and I think we should implement tearDown but I tried to install Voyager and I don't get any error:
 - Laravel: v5.7.15
 - Voyager: v1.3.0
 - PHP: 7.3.11
 - Database: SQLite 3.26.0 ~(not sure how to check version)~

mchekin:
@MrCrayon 

Did you try running **php artisan migrate:reset** so that `down()` method is called on the Voyager migrations?
Specifically on the problematic one:
https://github.com/the-control-group/voyager/blob/0306d7348f3dd05a4ea19c66f18bc9986bbe0902/migrations/2017_11_26_013050_add_user_role_relationship.php#L30

I do not think that SQLite version matters, since based on the Laravel code it throws an Exception on calling `dropForeign` for `SQLiteConnection`, regardless of the underlying version:

https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Schema/Blueprint.php#L149

MrCrayon:
To avoid any problem with leftovers I created a new fresh project with versions stated above, I didn't use tests.

I saw the check in Blueprint but it's just not happening :thinking: 
```
Migrating: 2017_11_26_013050_add_user_role_relationship
Migrated:  2017_11_26_013050_add_user_role_relationship
```
Did you a try a fresh install? Can you reproduce the problem?

Overall I think the Exception has a reason to stay, we should document it and find a solution for running tests.
~But all this after we understand why it's not failing right now.~

My bad of course the problem is happening while running `php artisan migrate:reset ` not on install, mistery solved :grin: 

I still think it should stay

mchekin:
You have migrated successfully.

But did you try to rollback/reset the migration?

MrCrayon:
> You have migrated successfully. But did you try to rollback/reset the migration?

Yeah my bad I noticed after that we are talking about the rollback.

I'm not familiar using SQLite but it doesn't sound safe to me to simply skip the Exception, we should let users know that something is not working as expected.

mchekin:
It's not about skipping the exception.
It's about not making a call that illegal for SQLite vendor in Laravel.

This check is what actually proposed in the Laravel update guide from
version 5.6 to 5.7.

And it says that it is no longer valid to call `dropForeign()` on SQLite
connection in Laravel.

https://laravel.com/docs/5.7/upgrade

The problem is that, at the moment, the aformentioned migration makse the
call to `dropForeign()` without checking for the vendor.

As a consequence you cannot use Voyager with SQLite in your project during
development where you need to rebuild you database. The migration rollback
fails.

This is not the Voyager user's fault since he cannot control the faulty
migration inside the Voyager library.

MrCrayon:
I understand your point and as I said I'm not familiar with SQLite use of foreign keys.

My question is:
If we skip `dropForeign()` for SQLite is still everything going to work as expected for everybody?

When I was quickly checking earlier on Laravel issues I was under the impression that as alternative for SQLite to drop a foreign key you need to create another table, so **if I understood it correctly** we either implement that in some way or we should notify the user that an action is needed.

https://stackoverflow.com/questions/1884787/how-do-i-drop-a-constraint-from-a-sqlite-3-6-21-table

In the other hand if you are using SQLite but are not using foreign keys you can copy migrations from voyager in your migrations folder and edit it.
Maybe a better way would be to make them available in `artisan vendor:publish` but I don't know why this is not the case already.

mchekin:
@MrCrayon 
I understand your concerns.

You are right that there is a way for dropping foreign keys in SQLite by creating new table:
https://www.techonthenet.com/sqlite/foreign_keys/drop.php

However Laravel's `Blueprint::dropForeign()` (in Laravel >5.7) does not support it, it just throws an Exception.
Quote from https://laravel.com/docs/5.7/upgrade :
> SQLite Foreign Keys
Likelihood Of Impact: Medium

> SQLite does not support dropping foreign keys. For that reason, using the dropForeign method on a table now throws an exception. Generally, this should be considered a bug fix; however, it is listed as a breaking change out of caution.

> If you run your migrations on multiple types of databases, consider using  DB::getDriverName() in your migrations to skip unsupported foreign key methods for SQLite.

Moreover, previous versions of Laravel (< 5.6) was not supporting `Blueprint::dropForeign()` with SQLite. They just did nothing.

I made a small check by running: `php artisan migrate:reset --pretend` and commenting out the other code in `AddUserRoleRelationship::down()` method (with Laravel 5.6):
```
public function down()
{
    });

    //Schema::table('users', function (Blueprint $table) {
    //    $table->bigInteger('role_id')->change();
    //});
}
```
https://github.com/the-control-group/voyager/blob/1.3/migrations/2017_11_26_013050_add_user_role_relationship.php#L30

This is the result:
```
...
Rolling back: 2017_11_26_013050_add_user_role_relationship
Rolling back: 2017_08_05_000000_add_group_to_settings_table
...
```

This is the result without commenting out second command:
```
Rolling back: 2017_11_26_013050_add_user_role_relationship
  AddUserRoleRelationship: DROP INDEX users_email_unique
  AddUserRoleRelationship: CREATE TEMPORARY TABLE __temp__users AS SELECT id, name, email, password, remember_token, created_at, updated_at, avatar, role_id, settings FROM users
  AddUserRoleRelationship: DROP TABLE users
  AddUserRoleRelationship: CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, name VARCHAR(255) NOT NULL COLLATE BINARY, email VARCHAR(255) NOT NULL COLLATE BINARY, password VARCHAR(255) NOT NULL COLLATE BINARY,
remember_token VARCHAR(255) DEFAULT NULL COLLATE BINARY, created_at DATETIME DEFAULT NULL, updated_at DATETIME DEFAULT NULL, avatar VARCHAR(255) DEFAULT 'users/default.png' COLLATE BINARY, settings CLOB DEFAULT NULL COLLATE BINARY,
role_id BIGINT DEFAULT NULL)
  AddUserRoleRelationship: INSERT INTO users (id, name, email, password, remember_token, created_at, updated_at, avatar, role_id, settings) SELECT id, name, email, password, remember_token, created_at, updated_at, avatar, role_id,
settings FROM __temp__users
  AddUserRoleRelationship: DROP TABLE __temp__users
  AddUserRoleRelationship: CREATE UNIQUE INDEX users_email_unique ON users (email)
```

As you see the command  `Blueprint::dropForeign()` does nothing with SQLite and goes straight to the next migration.

**Therefore disabling it does no harm, but fixes the issue as recommended by the framework.**

mchekin:
@MrCrayon 
Of course you are right that `Blueprint::dropForeign()` should be implemented in the right way.
It should be properly implemented in Laravel, instead of throwing the exception.

However leaving it as is in Voyager affects it's users.
And not just me.

https://laracasts.com/discuss/channels/testing/all-tests-started-failing
Quote:
>Thanks. I realised its due to me adding Voyager. The voyager migrations then break the tests. Not sure what to do with the migrations to stop the error so I decided to remove voyager and will reinstall it at the end of development I think.

mchekin:
@MrCrayon 
There is also more information in the Laravel pull request for this change, if you are interested:

https://github.com/laravel/framework/issues/24041
https://github.com/laravel/framework/pull/24052
https://github.com/laravel/framework/issues/25475

MrCrayon:
> Therefore disabling it does no harm, but fixes the issue as recommended by the framework.

We have a different interpretation of what is said in Laravel documentation, it's suggested to:
> **consider** using DB::getDriverName() in your migrations to skip unsupported foreign key methods for SQLite.

Because depending on your use case it might cause no harm, but that's a decision the user must take.

mchekin:
@MrCrayon 
I understand you are a busy person, but it seems you did not really read my previous 3 comments.
I will reiterate on the main point:

1. **In Laravel < 5.7: `Blueprint::dropForeign()` call does nothing with SQLite.**
2. **In Laravel >= 5.7: `Blueprint::dropForeign()` call throws an exception with SQLite.**

> Because depending on your use case it might cause no harm, but that's a decision the user must take.

That is the decision the **Laravel** user must make, but **Voyager** is forces this decision on the user by making this decision for him.

**Again, this command does not work for SQLite in ANY version of Laravel**
**It breaks your code if try to use it in newer versions of Laravel**

MrCrayon:
I read all your comments.
I think you are missing the point: there is a reason why Laravel Framework decided to stop ignoring that dropForeign method is not supported and started throwing an Exception.

mchekin:
> I think you are missing the point: there is a reason why Laravel Framework decided to stop ignoring that dropForeign method is not supported and started throwing an Exception.

Exactly, so that you won't use `Blueprint::dropForeign()` with SQLite.
I am not trying to fix **Laravel**, I am trying to fix **Voyager** :)

**Voyager is broken, because it's migrations do not support SQLite.**

**Please tell me. How is someone supposed to use Voyager with SQLite?**

mchekin:
@MrCrayon 
Should Voyager work out of the box with SQLite?
Or should the user of Voyager hack around and circumvent it's migrations because they are broken for some vendors?

mchekin:
@MrCrayon 
If I use Voyager, and I run it's migrations, which are inside the library, is it my fault that they are broken?
Should Voyager library be broken by design?

MrCrayon:
@mchekin
It's better to have an Exception that can be easily fixed copying and changing the migration than having silently something go wrong.
I'm not against the change if there is no problem but I'm not aware of all implications since I normally don't use sqlite.

I didn't see you saying that you tested with `DB_FOREIGN_KEYS=true`, can you try?

mchekin:
@MrCrayon 

You have already proposed it:
https://github.com/the-control-group/voyager/pull/4577#issuecomment-557875144

And I already answered it:
https://github.com/the-control-group/voyager/pull/4577#issuecomment-557934193

Moreover I already told you that the command does nothing `Blueprint::dropForeign()` on SQLite.

mchekin:
@MrCrayon 
> It's better to have an Exception that can be easily fixed copying and changing the migration than having silently something go wrong.

You cannot catch this Exception, since the migration is in Voyager.

**Does it make sense to you that anyone who wants to use Voyager with SQLite, has to disable Voyager migrations and copy all the migrations 20 migrations from Voyager into his own application, to fix this single line in one migration that is broken?** 

Is this good user experience from your standpoint?

Moreover, user should not be compelled to fix something that is broken inside the library.

And this line doesn't even do anything in SQLite anyway 🤦‍♂ 

I really don't see the problem.

> I'm not against the change if there is no problem but I'm not aware of all implications since I normally don't use sqlite.

I already explained to you all the implications and shown that this a recommended way by Laravel to deal with this issue.

Breaking user application is not giving user a choice, it is forcing the user to use hacky ways to fix something that should have worked in the library.

**A library like Voyager should be durable in order to be usable, but it cannot be considered durable if it breaks for some of the popular vendors.**

MrCrayon:
I suggest you to take a step back and read carefully my comment without prejudice.

1. I asked you to test `DB_FOREIGN_KEYS=true` because if you don't use that you are not enforcing foreign keys so your are not testing the use case where skipping `dropForeign` _might_ be a problem.
   Can you test with your fix applied and `DB_FOREIGN_KEYS=true`?

2. You don't need to disable Voyager migrations and don't need to copy all them you just need to copy one and change it, try.
   Publishing migrations and changing them is the way all package I know of manage conflicts.

3. I'm not native english speaker but _consider_ is not the same as _recommend_.

Hope next comment will be you saying that you tested what I asked and there is no problem so we can move on.

mchekin:
@MrCrayon 

1. Already told you this won't work (you just need to look at the source code) 
And **yes I tested it** with `DB_FOREIGN_KEYS=true`.

You have already proposed it (3 times):
And I already answered it (this is the 3rd time)

This time I will not only provide you with the link to the source code:
https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Schema/Blueprint.php#L149

But I will also embed it for clarity:
```
protected function ensureCommandsAreValid(Connection $connection)
    {
        if ($connection instanceof SQLiteConnection) {
            if ($this->commandsNamed(['dropColumn', 'renameColumn'])->count() > 1) {
                throw new BadMethodCallException(
                    "SQLite doesn't support multiple calls to dropColumn / renameColumn in a single modification."
                );
            }
            if ($this->commandsNamed(['dropForeign'])->count() > 0) {
                throw new BadMethodCallException(
                    "SQLite doesn't support dropping foreign keys (you would need to re-create the table)."
                );
            }
        }
    }
```

I hope you can read this code and understand that if you use `dropForeign` with SQLite Laravel throws an exception (regardless of anything else you use).

2. I do not know any way to run just some Voyager migration and not others.
I only know about `'autoload_migrations' => false,` option in `\config\voyager.php`.
If you know how to do it please tell me overwrite a specific Voyager migration please tell me.

3. This is the difference:
```
consider: 
1. think carefully about (something), typically before making a decision.
2. look attentively at.

recommend:
1. put forward (someone or something) with approval as being suitable for a particular purpose or role.
2. commend or entrust someone or something to (someone).
```

Of course, the language of the Laravel documentation is **consider** because no one can force you to support SQLite in your library.

Here I provide you with all the information you need.
You just need to read it.

MrCrayon:
1. I already checked Blueprint code as I told you 12 days ago, the problem is that you are still not understanding what I asked you. I asked you to test with **your fix applied and `DB_FOREIGN_KEYS=true`**, if you apply your fix that Blueprint method is not even called so :thinking: 
It's still unclear if you tried and what was the result.
2. I already told you in my previous comment, copy that single migration in your migration folder, edit it and next time you run migrations that one will be used instead of the one in vendor.
3. You just proved my point and your explanation makes no sense if they used **suggest** that didn't mean they were forcing you either but still they used **consider** for reasons that you are still missing to understand.

If you are actively using `DB_FOREIGN_KEYS=true` you would want to know that migration is not doing what it says, that's why there is the Exception because only the user can know if it's ok to skip it or not.

I'm not saying that in this particular migration we cannot safely skip it but it needs to be tested.

mattrabe:
Anyone experiencing this: The solution is:

> copy that single migration in your migration folder, edit it and next time you run migrations that one will be used instead of the one in vendor.

Put this conditional around the `dropForeign()` call: `if (DB::getDriverName() !== 'sqlite') {}`

PR #4580, opened 2019-11-26
Form fieldtag

riyastir:
Added Form Field Tag in Bread

Tag


emptynick:
This is much better implemented as a hook.

On a side note, we are using NPM which makes it much easier to update versions, fix vulnerabilities and so on instead of copying source directly to files.

PR #4590, opened 2019-11-28
fix belongsTo relationship dropdown

### LINKED ISSUE #4589: Only one option available for a belongsTo relationship
- Laravel: v5.8.35
 - Voyager: v1.3.0
 - PHP: 7.2
 - MySQL 5.7

## Description

When editing a `Model-A` that has a belongsTo relationship with `Model-B`, only one item from `Model-B` is available for selection.

## Steps To Reproduce
1. Add bread to `Model-A` and `Model-B`
2. Edit `Model-A` bread to have a belongsTo relationship with `Model-B`
3. Edit an existing record in `Model-A`
4. Only a record in `Model-B` is available to select from dropdown

## Expected behavior
When editing the belongsTo field in `Model-A`, it should bring option for all available records in `Model-B`

damms005:
This PR Fixes #4589 

Also, my check shows that this bug seems to have been there since v1.2. ([1.1](https://github.com/the-control-group/voyager/blob/1.1/resources/views/formfields/relationship.blade.php) => [1.2](https://github.com/the-control-group/voyager/blob/1.2/resources/views/formfields/relationship.blade.php))

If this assumption is true, then the fact that the issue is unnoticed for this long may suggest that most users override the `views/formfields/relationship.blade.php` file; which then again suggests that some improvements need be done in the file (...like adding support for links to related models rather than current labels to related models)

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -30,7 +30,7 @@
                 >
                     @php
                         $model = app($options->model);
-                        $query = $model::where($options->key, $dataTypeContent->{$options->column})->get();
+                        $query = $model::all();
                     @endphp
 
                     @if(!$row->required)


MrCrayon:
@damms005 I think we agree this one was not needed.

damms005:
Sorry @MrCrayon I forgot to close the PR after we had discussed the issue. My apologies

MrCrayon:
np :wink:

PR #4587, opened 2019-11-28
Fix delete bread action

MrCrayon:
In Laravel  ~6 associative array parameters passed are attached to query string.~ 6.6 passing null is not working anymore.
(Change here https://github.com/laravel/framework/pull/30659/files)

The += was also always adding ids on top of the previous one

Fixes #4586

PR #4593, opened 2019-11-29
In larave 6.6  null parameter is not accepted.

developerdesigner18:
In laravel 6.6 null parameter cleare issue and throw error in /admin/database "Missing required parameters for [Route: voyager.bread.delete] [URI: admin/bread/{id}]" So i change null to '' blank and it works fine now.


MrCrayon:
As far as I understand it if {id} is required building route it should not be empty.
That's exactly the bug they tried to fix in 6.6 with https://github.com/laravel/framework/pull/30659
So if this is going to be fixed, as I think, empty string is not going to work.

And there is already another PR open #4587

developerdesigner18:
Yes but voyager pass null which is not correct too. Passing '' blank works instead on null.

image

MrCrayon:
I'm aware, that's why:
> And there is already another PR open #4587

PR #4604, opened 2019-12-03
Force create of admin user.

tjventurini:
There is a conflict with eg. Spark. Users can't set the password during create. I changed to `forceCreate` and added a check if the user exists already. I hope that fixes the issue for the future :smiley:


src/Commands/AdminCommand.php
```diff
@@ -123,6 +123,13 @@ protected function getUser($create = false)
 
         // If we need to create a new user go ahead and create it
         if ($create) {
+            // check if user with given email exists
+            $User = $model::where('email', $email)->first();
```
MrCrayon: You should move this check later on because at this point you might still not have the email.

MrCrayon (dismissed):
Never used Spark so not sure what's the problem but you should move your check.

src/Commands/AdminCommand.php
tjventurini: I moved it further down, after the check for the email.

src/Commands/AdminCommand.php
```diff
@@ -2,12 +2,13 @@
 
 namespace TCG\Voyager\Commands;
 
+use App\User;
+use Illuminate\Support\Str;
```
fletch3555, line 5: Please avoid committing changes like these (rearranging "use" statements).  They provide no benefit, and simply make StyleCI unhappy.  I realize it likely wasn't intentional, so don't worry about it.  StyleCI will auto-fix it after merge anyway.

src/Commands/AdminCommand.php
```diff
@@ -2,12 +2,13 @@
 
 namespace TCG\Voyager\Commands;
 
+use App\User;
```
fletch3555: No need to include this line.  You're not using it anywhere, and it's not guaranteed to exist.

src/Commands/AdminCommand.php
```diff
@@ -131,6 +132,13 @@ protected function getUser($create = false)
             if (!$email) {
                 $email = $this->ask('Enter the admin email');
             }
+            
+            // check if user with given email exists
+            $User = $model::where('email', $email)->first();
```
fletch3555: Please use lowercase (camel-case, though it doesn't matter here) for variable names.

Also, can't you use `->exists()` on the model instead since you're not using the object anywhere.  This will allow you to combine this line with the conditional below it

src/Commands/AdminCommand.php
```diff
@@ -131,6 +132,13 @@ protected function getUser($create = false)
...
+            if ($User) {
```
fletch3555: Same here

fletch3555 (dismissed):

src/Commands/AdminCommand.php
tjventurini: True. Updated it.

emptynick (dismissed):

emptynick (approved):

tjventurini:
Baam ... half a year later it is merged ... Not sure if I should party or complain 😅

emptynick:
Sorry for the long delay, but we are quite busy with v2 at the moment.

PR #4605, opened 2019-12-03
Fix resource routes with UTF8 characters

### LINKED ISSUE #4598: BRED actions doesn't work for freshly created BREAD item
Laravel: 6.6.0
Voyager: 1.3.0
PHP: 7.3.6

Server startet by `artisan serve`
image

Made a clear install of Laravel and then Voyager;
Created a table by migration;
image
Created BREAD for the table;
Added items to newly created BREAD object;
Other actions on that object delivers 404
image

MrCrayon:
By default Laravel uses resource name as placeholder in resource routes, this causes problems with Symfony route recognition and also can easily surpass the 32 characters limit for placeholder.
We don't really need to use a custom placeholer, {id} is fine enough.
More info in Issue fixed.

Fixes #4598

### DIFF
--- a/resources/views/menus/browse.blade.php
+++ b/resources/views/menus/browse.blade.php
@@ -107,7 +107,7 @@
         });
 
         $('td').on('click', '.delete', function (e) {
-            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['menu' => '__menu']) }}'.replace('__menu', $(this).data('id'));
+            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['id' => '__menu']) }}'.replace('__menu', $(this).data('id'));
 
             $('#delete_modal').modal('show');
         });
--- a/routes/voyager.php
+++ b/routes/voyager.php
@@ -47,7 +47,7 @@ Route::group(['as' => 'voyager.'], function () {
                 Route::get($dataType->slug.'/{id}/restore', $breadController.'@restore')->name($dataType->slug.'.restore');
                 Route::get($dataType->slug.'/relation', $breadController.'@relation')->name($dataType->slug.'.relation');
                 Route::post($dataType->slug.'/remove', $breadController.'@remove_media')->name($dataType->slug.'.media.remove');
-                Route::resource($dataType->slug, $breadController);
+                Route::resource($dataType->slug, $breadController, ['parameters' => [$dataType->slug => 'id']]);
             }
         } catch (\InvalidArgumentException $e) {
             throw new \InvalidArgumentException("Custom routes hasn't been configured because: ".$e->getMessage(), 1);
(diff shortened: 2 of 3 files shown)


fletch3555:
Doesn't that parameter have to match the name of the route parameter?  If so, we use `Route::resource()` under the hood, which generates routes like `/admin/<resourceName>/{<resourceName>}` (e.g. `/admin/posts/{post}`)

MrCrayon:
The resource name `posts` and placeholder `{post}` don't have to match, by default  `Route:resource()` will use the same name unless you override it with `parameters`.

Placeholder needs to match what you use in url generator, if your route is defined as:
`/admin/posts/{post}`
you need to use:
`route('voyager.posts.edit', ['post' => $post->id])`
if route is defined as:
`/admin/posts/{id}`
you need to use:
`route('voyager.posts.edit', ['id' => $post->id])`

And in both cases this will work:
`route('voyager.posts.edit', [$post->id])`
`route('voyager.posts.edit', $post->id)`

I only changed three instances where in Voyager was explicitly used resource name as key generating route, in all other places we are not using a key.

resources/views/menus/browse.blade.php
```diff
@@ -107,7 +107,7 @@
         });
 
         $('td').on('click', '.delete', function (e) {
-            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['menu' => '__menu']) }}'.replace('__menu', $(this).data('id'));
+            $('#delete_form')[0].action = '{{ route('voyager.'.$dataType->slug.'.destroy', ['id' => '__menu']) }}'.replace('__menu', $(this).data('id'));
```
fletch3555, line 110: @MrCrayon, in response to your last comment, this is the line I was referring to.  You're using the `['x' => 'y']` syntax, which you agreed needed to be `widget` if the resource route was defined as `widgets`.  In this case, `id` shouldn't work, and instead should be the singular version of `$dataType->slug`, right? Or am I missing something?

Alternatively, we could remove the key altogether and just leave it as `route('....', ['__menu'])`, right?

routes/voyager.php
```diff
@@ -47,7 +47,7 @@
                 Route::get($dataType->slug.'/{id}/restore', $breadController.'@restore')->name($dataType->slug.'.restore');
                 Route::get($dataType->slug.'/relation', $breadController.'@relation')->name($dataType->slug.'.relation');
                 Route::post($dataType->slug.'/remove', $breadController.'@remove_media')->name($dataType->slug.'.media.remove');
-                Route::resource($dataType->slug, $breadController);
+                Route::resource($dataType->slug, $breadController, ['parameters' => [$dataType->slug => 'id']]);
```
fletch3555, line 50: Ah, nevermind... I see.  You're forcing **all** resource route parameters to be `id`

MrCrayon: Yes, exactly. 
We only have one parameter so not really any reason to not use just {id} if it solves the utf8 character problem.

fletch3555 (approved):

PR #4620, opened 2019-12-11
Ignore non-editable fields when updating a Translatable model

terax6669:
Right now when you have a BREAD field that's not editable but has translations turned on you can't update the model because of the exception.

This fixes it, so you can save your changes without error.


MrCrayon:
Can you explain which are the steps required to reproduce the problem you are trying to fix? 

If a field has edit uncheck it shouldn't interfere in any way with saving.

terax6669:
This loop only works when all translatable fields are also editable. I have a model with 3 translatable fields, but only one of them is editable in BREAD configuration. I can't update because I'm POSTing just 1 out of 3 fields.

MrCrayon:
I see, I was able to reproduce the problem, but I don't think skipping a valid check is the right way to fix this.
I think we should pass editRows / addRows and skip if $field is not in the list.

terax6669:
We are foreaching by translatable fields, not by input from user so we know those fields are ok to be translated.
So why force all fields to be present?

MrCrayon:
> So why force all fields to be present?

That's what I'm saying, now foreach is done from transaltable field list but they might not be all there.
That's why the solution is to foreach only translatable fields in editRows / addRows

terax6669:
Is this what you had in mind?

MrCrayon:
> Is this what you had in mind?

#4709 is what I had in mind.

This one looks a bit overcomplicated ~and are you sure it's working?~
~`$request` is not passed by reference anymore so any change should be lost.~

terax6669:
Request is an object, so it's always the same instance

MrCrayon:
Of course you are right, I was fooled by the use of it.

PR #4643, opened 2019-12-19
Fix bread ordering

terax6669:
Disable sorting on columns that are not sortable anyway.

Voyager does not support server-side sorting by relationship.


emptynick (dismissed):

emptynick (approved):

mko-tapptic:
I'm currently using this fix in a project using 1.4.2 so it should be compatible with 1.4 as well

terax6669:
I'm currently using this fix in a project using 1.4.2 so it should be compatible with 1.4 as well

emptynick:
Yeah there haven't been any changes related to this in 1.4

PR #4644, opened 2019-12-19
Fix server-side default sort direction

terax6669:
Right now when BREAD is server-side it's always sorted in descending order, this PR uses the BREAD preference instead.


MrCrayon:
Thanks for PR but I don't think that works because:
```php
if (!$sortOrder && isset($dataType->order_direction)) {
```

Anyway duplicate of #4557

MrCrayon:
Closing as duplicate of #4557

PR #4645, opened 2019-12-19
Change duplicate ids to a class and fix related style

terax6669:
In HTML the `id` attribute should be unique.


MrCrayon (approved):
That annoyed me too :grin: 

Thanks for fixing it.

@fletch3555 if you set up the action we can use this PR to compile assets.

PR #4647, opened 2019-12-19
Fix displaying translation 3 times when rich text box field is also translatable

terax6669:
Data from TinyMCE already includes the `<p>` tag.

Paragraphs can't be nested: https://www.w3.org/TR/html401/struct/text.html#h-9.3.1

Browsers automatically open/close the `<p>` tags resulting in 3 tags instead of 1. When using translations with Rich Text Box type it was filling all three `<p>` tags with the translation.

image


MrCrayon (approved):
Bug confirmed and PR is working, good job :+1:

fletch3555 (approved):

PR #4659, opened 2019-12-28
Configurable user class

commnerd:


publishable/config/voyager.php
```diff
@@ -11,6 +11,7 @@
     */
 
     'user' => [
+        'model'                        => \App\User::class,
```
fletch3555, line 14: Need to also update the "dummy" config: https://github.com/the-control-group/voyager/blob/1.3/publishable/config/voyager_dummy.php

fletch3555 (changes requested):
Can't say I'm really a fan of this.  We used to have a config just like this, but removed it in favor of just pulling it off the model class defined in BREAD.  Now obviously we can't do that in these migrations since BREAD hasn't even been seeded/created yet, which is why these remained hardcoded.  It's easy enough to override the migrations if you need to.

PR #4664, opened 2019-12-31
Persist DataTable selections - Filter, Sort, and Page

mattrabe:
Currently, if a user changes the Search/Filter, Sorting order, or Page within a client-side list, then navigates to another page (such as Edit or View a single record), and then clicks Back, they are brought to the original list but WITHOUT any of the Search/Sort/Page selections they had made before. Put another way, when a user changes those selections for the DataTable their change is not tracked anywhere in history and cannot be returned to.

This PR uses `history.pushState` to modify the browser's URL when a user changes a DataTable selection. On the receiving end, when a browse page is loaded (or is popped from history by clicking Back/Forward browser buttons) this calls functions on the DataTable and updates the list so that it matches the URL.

The end result: search/filter, sort, and pagination selections are retained in browser history, and clicking Back/Forward browser buttons or reloading the page work as expected and the list view does not "reset" to the first page/no search filter/default ordering.


emptynick (changes requested):
I'd love to see some tests for this.
E.g. opening a URL with parameters and see that the inputs contain what we desire.

Not sure if this is possible with our current testbench though as we don't use Dusk.
If it doesn't work you could just hit Search and check it from the controller.

mattrabe:
@emptynick Hm yea, Dusk would be a great way to test this. It's 100% javascript, so I'm not sure what there would be to test in the Controller - maybe I didn't understand that part (?). If you want, I could add dusk as a dev dependency and write up a few tests.

PR #4669, opened 2019-12-31
Remove duplicate route "voyager.roles.update"

d3turnes:
Override controller BreadBaseController to VoyagerRoleController for roles


MrCrayon (changes requested):
Thanks for finding the bug and submitting a PR for it.

To be more on line with Voyager functionalities long term solution would be to remove:
https://github.com/the-control-group/voyager/blob/05f5a4f744ba640bcbd91bcc66c914a89d710bfe/routes/voyager.php#L59

And set custom controller in BREAD settings.

Since that would be a breaking change we can't do it right now but can you change your PR?
 - Undo your changes
- Change `/routes/voyager.php#L59`
```diff
-        Route::resource('roles', $namespacePrefix.'VoyagerRoleController');
+        Route::resource('roles', $namespacePrefix.'VoyagerRoleController', ['parameters' => ['roles' => 'id']]);
```
- Change `/publishable/database/seeds/DataTypesTableSeeder.php#L50`
```diff
-                'controller'            => '',
+                'controller'            => 'TCG\\Voyager\\Http\\Controllers\\VoyagerRoleController',
```

Thanks!

MrCrayon (approved):
:+1:

PR #4673, opened 2020-01-04
Tests for changing role permissions

MrCrayon:
Tests if permissions can be changed in role.

This test should fail in 1.3.1 but for the way routes are defined in test enviroment DataType routes are set at last so changing permissions still works.
Introduced index for permissions in edit-add so it can be referenced in tests.


resources/views/roles/edit-add.blade.php
```diff
@@ -67,7 +67,7 @@
                                         <ul>
                                             @foreach($permission as $perm)
                                                 <li>
-                                                    <input type="checkbox" id="permission-{{$perm->id}}" name="permissions[]" class="the-permission" value="{{$perm->id}}" @if(in_array($perm->key, $role_permissions)) checked @endif>
+                                                    <input type="checkbox" id="permission-{{$perm->id}}" name="permissions[{{$perm->id}}]" class="the-permission" value="{{$perm->id}}" @if(in_array($perm->key, $role_permissions)) checked @endif>
```
fletch3555, line 70: Is there a particular reason this was added? Form arrays in Laravel are identified by `<fieldname>[]`.  No ID value necessary.

MrCrayon: It doesn't change anything saving but I didn't find any simple way to check or uncheck it in tests without a unique name, referencing id doesn't work.

emptynick (dismissed):

emptynick (approved):

PR #4675, opened 2020-01-05
WIP - nestable BREAD ordering

### LINKED ISSUE #4641: BREAD Reorder
## Version information

 - Laravel: v6.3
 - Voyager: v1.3
 - PHP: 7.2
 - Database: MySQL

## Description of problem

I need reorder any BREAD with childs. Now we have button #order, but i need reorder elements with childs! In Backpack we can do it. We just need to add parent_id, lft, rgt and depth columns in our table and we got REORDER with childs.
I have big problems now, coz my lovely Voyager do not have REORDER with CHILDS REORDER.
I have 400 page in my website. I just want to have 400 rows in my table and have parent_it, lft, rgt, depth fields. And simple functionality as Backpack with Reorder any level!

rs

This is NOT menu reorder with childs we need. Resource (database table with rows) reorder with childs need.

I dont' want any feature, but i need this! Is most grate feature what we need! We need to get tree of all resources in our BREAD - just 4 fields (parent_id, lft, rgt, depth).

Please guys make me happy! Please! I love Voyager. But I have biggest minus with it.

MrCrayon:
Make ordering nestable if Trait is used by Model, Model will need to set parent id field like:
```php
use TCG\Voyager\Traits\Nestable;

class Category extends Model
{
    use Nestable;

    protected $parentField = 'parent_id';
```

Fixes order not mantained while using descending order direction.

Closes #4641

Co-Authored-By: dzianisyaukhuta <dzianisyaukhuta@users.noreply.github.com>

### DIFF
--- a/resources/views/bread/order.blade.php
+++ b/resources/views/bread/order.blade.php
@@ -19,23 +19,7 @@
 
                 <div class="panel-body" style="padding:30px;">
                     <div class="dd">
-                        <ol class="dd-list">
-                            @foreach ($results as $result)
-                            <li class="dd-item" data-id="{{ $result->getKey() }}">
-                                <div class="dd-handle" style="height:inherit">
-                                    @if (isset($dataRow->details->view))
-                                        @include($dataRow->details->view, ['row' => $dataRow, 'dataType' => $dataType, 'dataTypeContent' => $result, 'content' => $result->{$display_column}, 'action' => 'order'])
-                                    @elseif($dataRow->type == 'image')
-                                        <span>
-                                            <img src="@if( !filter_var($result->{$display_column}, FILTER_VALIDATE_URL)){{ Voyager::image( $result->{$display_column} ) }}@else{{ $result->{$display_column} }}@endif" style="height:100px">
-                                        </span>
-                                    @else
-                                        <span>{{ $result->{$display_column} }}</span>
-                                    @endif
-                                </div>
-                            </li>
-                            @endforeach
-                        </ol>
+                        @include('voyager::bread.partials.order', ['items' => $results, 'nestable' => $nestable, 'parentId' => ''])
                     </div>
                 </div>
             </div>
@@ -50,7 +34,7 @@
 <script>
 $(document).ready(function () {
     $('.dd').nestable({
-        maxDepth: 1
+        @if(empty($nestable)) maxDepth: 1 @endif
     });
 
     /**
--- a/resources/views/bread/partials/order.blade.php
+++ b/resources/views/bread/partials/order.blade.php
@@ -1,6 +1,6 @@
 <ol class="dd-list">
     @foreach ($items as $result)
-        @if($nestable && $result->parent_id != $parentId)
+        @if(!empty($nestable) && $result->{$result->getParentField()} != $parentId)
             @continue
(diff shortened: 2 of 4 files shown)


resources/views/bread/order.blade.php
```diff
@@ -50,7 +34,7 @@
 <script>
 $(document).ready(function () {
     $('.dd').nestable({
-        maxDepth: 1
+        @if(empty($nestable)) maxDepth: 1 @endif
```
fletch3555, line 37: Why check `empty()` here when you return a boolean from the controller?

resources/views/bread/partials/order.blade.php
```diff
@@ -0,0 +1,23 @@
...
+                @include($dataRow->details->view, ['row' => $dataRow, 'dataType' => $dataType, 'dataTypeContent' => $result, 'content' => $result->{$display_column}, 'action' => 'order'])
+            @elseif($dataRow->type == 'image')
+                <span>
+                    <img src="@if( !filter_var($result->{$display_column}, FILTER_VALIDATE_URL)){{ Voyager::image( $result->{$display_column} ) }}@else{{ $result->{$display_column} }}@endif" style="height:100px">
+                </span>
+            @else
+                <span>{{ $result->{$display_column} }}</span>
+            @endif
+        </div>
+        @if(isset($result->nestableChildren) && !$result->nestableChildren->isEmpty())
```
fletch3555: So.. nestableChildren holds the child entries... do you need $testable at all? You could just check if this list is empty, right?

resources/views/bread/order.blade.php
MrCrayon: I'm trying to avoid problems in case someone already overrode order method and updates.
Following that logic I forgot to do the same in the partial view. 
Let me know if you prefer instead that we force the user to keep controller updated.

resources/views/bread/partials/order.blade.php
MrCrayon: I can only check if element is under the right parent if it's nestable, otherwise there is no parent id. 

I need to change parent_id that's the hard coded value I was using at start.

dzianisyaukhuta:
Thank you so much @MrCrayon !!!  I hope we still get a sort of BREAD in a tree view.

MrCrayon:
@dzianisyaukhuta the result is the same as your PR but with much less code :grin: 
Schermata da 2020-01-10 10-58-26

dzianisyaukhuta:
> @dzianisyaukhuta the result is the same as your PR but with much less code 
> Schermata da 2020-01-10 10-58-26

Bravo!

dzianisyaukhuta:
Can anybody approve it? @fletch3555

dzianisyaukhuta:
Why?

fletch3555:
Why what? Why is it not approved/merged? That answer is simple. It's still labeled `wip`

dzianisyaukhuta:
@MrCrayon big respect to you. You are awesome.

cotiga:
I added this to a fork under 1.4.2, works great.
Please add this kind of functionality.
Voyager seems to stagnate lately, please don't give up on us

emptynick:

PR #4686, opened 2020-01-10
Changed example

MrCrayon:
Because current example doesn't work and it doesn't really make any sense.

It doesn't work because _sometimes_ and _required_ are going to be both present.
It doesn't make much sense because we are repeating the one rule that doesn't change.

P.S.
There is currently a bug for which if the general rule is not set the other ones are ignored too.
https://github.com/the-control-group/voyager/blob/d2aa738bb9777ed8794d962c999df4623f44c3eb/src/Http/Controllers/Controller.php#L294


fletch3555 (approved):
👍

PR #4690, opened 2020-01-11
Show 'none' only on first page

howdu:


fletch3555:
Why would there be multiple pages if you're showing none....?

fletch3555:
The newest change seems unrelated to the first one. Is that correct?

fletch3555:
Again latest change is unrelated.  I'm closing this for now.  We prefer each PR only have one set of related changes.  Each of these need to be in their own branch

howdu:
Sorry I don't know how to do that. I'll create them as separate issues.

MrCrayon:
@howdu is best practice to create a new branch for every feature or bug fix, you can do:
`git checkout -b fix-ordering-bread 1.3`

PR #4692, opened 2020-01-12
Remove Route in InstallCommand

MrCrayon:
Routes needed to be defined on install for old Voyager versions because MenuItemsTableSeeder used to seed the full url into database.
That's not the case anymore so this can be removed.

This part of code was also causing routes on tests to be  defined in a different order than  normal use because on install Route was called before seeding DataTypes.


fletch3555 (approved):
👍

PR #4703, opened 2020-01-16
Fix displaying translatable rich text box to consistent with non-translatable

terax6669:
Updated bread's browse and read views to render in the same way.

Translated rich text on browse now strips most tags on translatable fields and truncates, same as non-translated.

Translated  rich test on read now renders with tags, same as non-translated.

Breaking change (possibly, but unlikely): changed `get_field_translations` parameters.


ahmedbally:
please follow up with that i miss this

PR #4716, opened 2020-01-24
Fix save bread when ajax calls are not done

### LINKED ISSUE #3960: [BUG] Undefined index: relationship_key
## Version information

 - Laravel: v5.7.24
 - Voyager: v1.1
 - PHP: 7.2.4
 - Database: MySQL 5.7.22

## Description

When submitting an edit of a BREAD with a relationship I (sometimes) get the following error:
`Undefined index: relationship_key_topic_belongstomany_topic_relationship `

In `DataType.php` this key value will be set according to the relationship. Which in my case is `topic_belongstomany_topic_relationship`.

When inspecting the requestData. The `realtionship_key_xxx` index is sometimes set and sometimes it's not. (hence the error)

## Steps to reproduce
Not really helpfull, but this is what I'm doing:

1. Select edit BREAD
2. Submit form
3. See whether the update works or not

## Expected behavior
The expected behavior is that this index (`relationship_key_xxx`) is set in the `$requestData`. Sometimes this is the case and sometimes it's not.

## Screenshots
screenshot 2019-02-19 at 16 56 09

## Additional context
This relationship is a relationship on the table itself. (A topic will show multiple topics and a topic can appear on multiple other topics as well).

I have made 2 models which reference Topic so I can get the related Topics to a certain Topic.

1 model under `App\Models\Voyager\Topic` (created by Voyager itself)

```
<?php

namespace App\Models\Voyager;

use TCG\Voyager\Traits\Translatable;

class Topic extends Model
{   
    use Translatable;

    protected $table = "topics";

    protected $translatable = [
        "title",
        "intro",
        "content_title",
        "content_sub_title",
        "content",

    ];

    protected $with = ['translations'];

    public function topics()
    {
        return $this->belongsToMany("App\Models\Topic", 'topics_related_topics', 'topic_id', 'related_topic_id');
    }
}

```

1 model under `App\Models\Topic` (self created to create a belongsToMany relationship)

```
<?php

namespace App\Models;

class Topic extends Model
{
    protected $table = "topics";

    public function topics()
    {
        return $this->belongsToMany("App\Models\Voyager\Topic", 'topics_related_topics', 'related_topic_id', 'topic_id');
    }
}

```

MrCrayon:
Every relationship set in BREAD made three ajax calls to collect table fields.
In a BREAD with many relationships and/or slow server this could result in some relationship not correctly populated when hitting Submit button.

This PR prepopulates selects with already saved values and leaves ajax calls only when table is changed or modal is opened.

Fixes #3960

TODO
- [x] Fix select width in modal

Everything is working so far but in modal select2 doesn't keep the correct width, I think it can be fixed changing the parent element to:
```
display: flex
```
and select2 span to:
```
flex: 1
```
Then we can't use jQuery show() because it will set again to block.


MrCrayon:
This should also work as simple temporary solution for html/css problem but I'd prefer something better.

```css
#new_relationship_modal .well .select2-container {
    width: auto !important;
}
```

damms005:
This may be complete with a GUI feedback of a "loading" indicator in cases when there is need to do an ajax request. Hence the user will know to wait before hitting on the Submit button.

MrCrayon:
Sure could be a final touch but since this PR eliminates all ajax calls at load the only case which could cause this bug again is if you create a new relationship and hit Save really fast, I don't see that happening since you'll still need to type Model name.

MrCrayon:
@emptynick looks good to me, thanks.
Why the QA label, did you need any clarification about what this PR is solving or how it's doing it?

emptynick:
Just added it so I don't blindly approve it without you approving it first :+1:

emptynick (approved):

PR #4718, opened 2020-01-25
Intervention from Laravel5 to Laravel Recent

mikesaintsg:
Capture

Intervention has updated the ImageServiceProvider from Laravel 5 to LaravelRecent for compatibility with Laravel6


fletch3555 (approved):

fletch3555:
I thought we fixed this like 2 months ago.... whoops.  Thanks!

MrCrayon:
The error on install was fixed in #4638 but I missed the mention in documentation.
We don't publish it anymore on install since it's not really required but I guess it doesn't hurt leaving it in the docs.

PR #4719, opened 2020-01-26
Eagerload data rows translations

MrCrayon:
Eager load data rows and dataTypeContent translations.
Currently only dataTypeContent translations are eagerloaded and only in browse view.


fletch3555 (approved):

terax6669:
#4767 

please fix this.

PR #4728, opened 2020-02-01
WIP - Enable creation of BREAD for views

### LINKED ISSUE #1926: multiple bread for Single table / Condition in Browse data
- Laravel Version: 5.5
- Voyager Version: 1.0.4
- PHP Version: 7.1
- Database Driver & Version: PGSQL

### Description:
Can we add multiple breads for SIngle DB table & Apply diff conditions for browse.  Sometime we need to show* some diff data from single table with some diff. conditions.

Anshu.

MrCrayon:
Creation of BREAD works just fine for views but they were not listed.
Now listTableNames also includes views.
Before merging would be needed to add translation keys for all languages and maybe adjust the info label in listings.

TODO
- [ ] Add field for all translations
- [ ] Add documentation

Closes #1926

### DIFF
--- a/publishable/lang/en/database.php
+++ b/publishable/lang/en/database.php
@@ -45,6 +45,7 @@ return [
     'table_actions'             => 'Table Actions',
     'table_columns'             => 'Table Columns',
     'table_has_index'           => 'The table already has a primary index.',
+    'table_is_view'             => 'This is a view',
     'table_name'                => 'Table Name',
     'table_no_columns'          => 'The table has no columns...',
     'type'                      => 'Type',
--- a/resources/views/tools/bread/index.blade.php
+++ b/resources/views/tools/bread/index.blade.php
@@ -27,6 +27,9 @@
                     @continue(in_array($table->name, config('voyager.database.tables.hidden', [])))
                     <tr>
                         <td>
+                            @if(in_array($table->name, $views))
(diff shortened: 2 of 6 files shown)


fletch3555:
Historically, I haven't wanted to add views to the UI because novice users will find a way to break it (which is actually quite easy) and then complain that Voyager is broken.

This issue with views is that they're like a virtual table backed by a query against actual tables.  This is fine as long as it's a simply subset of a single table(i.e. all rows meeting some condition).  But as soon as you do anything more creative, such as joins, grouping, etc, you suddenly make that view read-only, which naturally breaks 99% of Voyager's functionality.

MrCrayon:
I understand your concern, there is going to be an icon warning, the sentence needs to crafted.
We can also add a warning while creating BREAD and write in documentation that using views has limitations that needs to be understood.

So I guess the decision will be between enabling Voyager full potential for expert users or play it safe for non expert users.

I would go with the first one, let me know when you decide so I know if I should work on what's left or let it be.

Schermata da 2019-12-16 14-11-35

P.S.
Another thing that will be possible with views is server side search on relationship fields.
A view can have both `user_id` and `user_name` (the latter taken from `users` table), `user_name` can be disabled on add/edit and will be available for search on browse.

livrosroberto:
imho that would be a great feature as long as warnings and good documentation is provided, as @MrCrayon said.

After all,  a user that has some "advanced" views is more likely to know that that view is read-only.

just my 2 cents

emptynick:

PR #4741, opened 2020-02-08
Enable overriding some partial views

MrCrayon:
This PR makes possibile to easily hide or override:
 - dashboard.navbar
 - dashboard.sidebar
 - partials.app-footer

This can be useful as example if we want to show a page in a Modal/Popup or Iframe
Schermata da 2020-02-08 09-28-45


fletch3555:
The views are already able to be overridden.  All this does is allow it to be done per-page

MrCrayon:
@fletch3555 yes that's the main point, but also right now you can't extend any of those partials you'll need to copy all content and change it.

MrCrayon:
@fletch3555 would you prefer to have
```blade
@section('dashboard_navbar')
@endsection
```
inside `dashboard.navbar` template instead of `master`, like it's done for `media.manager`?

PR #4747, opened 2020-02-09
Add detailed authority management to bread,database,media

tu6ge:
I plan to set up a demo app on my server and provide a public demo account, so I need more detailed permission control


tu6ge:
The merge failed because of the permission ？

PR #4760, opened 2020-02-12
Extent VoyagerBaseController for filtering relations on mustiselect with rawWhere option

tabaktoni:
Add to Relationship Details option for raw where filtering 
{
    "whereRaw": "role_id=2 or role_id=3"
}

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -864,7 +864,11 @@ class VoyagerBaseController extends Controller
                     }
                 } else {
                     $total_count = $model->count();
-                    $relationshipOptions = $model->take($on_page)->skip($skip)->get();
+                    if (isset($options->whereRaw)) {
+                        $relationshipOptions = $model->whereRaw($options->whereRaw)->take($on_page)->skip($skip)->get();
+                    } else {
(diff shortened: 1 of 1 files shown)


marktopper (dismissed):

src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -864,7 +864,11 @@ public function relation(Request $request)
                     }
                 } else {
                     $total_count = $model->count();
-                    $relationshipOptions = $model->take($on_page)->skip($skip)->get();
+                    if (isset($options->whereRaw)) {
+                        $relationshipOptions = $model->whereRaw($options->whereRaw)->take($on_page)->skip($skip)->get();
+                    } else {
+                        $relationshipOptions = $model->take($on_page)->skip($skip)->get();
+                    }
```
emptynick, line 871: I think this can be simplified like this. We also need documentation for this.

```suggestion
                    if (isset($options->whereRaw)) {
                        $model = $model->whereRaw($options->whereRaw)->take($on_page)->skip($skip)->get();
                    }
                    $relationshipOptions = $model->take($on_page)->skip($skip)->get();
```

emptynick (changes requested):

emptynick:
Superseded by https://github.com/the-control-group/voyager/pull/4851

PR #4780, opened 2020-02-25
check delete permission against model not DataType

JackShev:
For deleting models authorisation should get model as param, because in policies we use specific model attributes.
Example: 
$model->deleted_at in BasePolicy
or $model->author_id in in PostPolicy


fletch3555:
It is getting an instance of the model, but not the specific instance the row is rendering.  This was done to reduce database queries, since we never used to cache permissions.  I'd like to see evidence that this doesn't introduce the n+1 problem.

MrCrayon:
@fletch3555 this change is in destroy method not index. 
I didn't test it yet but it makes absolutely sense to check authorization for each instance otherwise it could be possible to delete even records that we are not allowed to delete based on policy.

fletch3555:
You're correct.  I hadn't noticed the method name.

I have no issues with this.

JackShev:
I have changed it because in example posts module delete button is displayed when I'm an author of the post. But when try to delete there is error 403. It happens because in view policy is chacked against specific model. So I think it is logic to check it the same way in controller.

fletch3555 (approved):

PR #4782, opened 2020-02-26
Fix of BUG

georak:
Call to a member function getTranslatedAttribute() on null {"userId":4,"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function getTranslatedAttribute() on null at /app/vendor/tcg/voyager/src/Http/Controllers/VoyagerBaseController.php:55


JackShev (dismissed):

src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -52,7 +52,8 @@ public function index(Request $request)
             $searchable = SchemaManager::describeTable(app($dataType->model_name)->getTable())->pluck('name')->toArray();
             $dataRow = Voyager::model('DataRow')->whereDataTypeId($dataType->id)->get();
             foreach ($searchable as $key => $value) {
-                $displayName = $dataRow->where('field', $value)->first()->getTranslatedAttribute('display_name');
+                $field = $dataRow->where('field', $value)->first();
+                $displayName = $field !== null ? $field->getTranslatedAttribute('display_name') : null;
                 $searchNames[$value] = $displayName ?: ucwords(str_replace('_', ' ', $value));
```
emptynick: I'm not really a big fan of those one-line conditions so I suggest doing something like:

```suggestion
                $displayName = ucwords(str_replace('_', ' ', $value));
                if ($field !== null) {
                    $displayName = $field->getTranslatedAttribute('display_name');
                }
                $searchNames[$value] = $displayName;
```

emptynick (changes requested):

emptynick (approved):

Issue #2648, opened 2018-02-12
HasOne and HasMany Relationships can only link to the Primary key of the child record.


IvandaNothabeer:
- Laravel Version: 5.5.32
- Voyager Version: 1.0.15
- PHP Version: 7.0.23
- Database Driver & Version: MySQL 5.7.19

I have the following Parent - Child table structure ....

Parent
 - id, primary key, Auto-Increment
 - name, string
 - child_id, integer

Child
- id, primary key, Auto-Increment
- name, string
- parent_id, integer

The relationship is Parent HasOne/HasMany Child linked on Parent.child_id-->Child.parent_id

BREAD Relationships assume that the relationship is Parent.child_id -> Child.id

See relationship.blade.php, Line 48 ....

```
	@php 

		$relationshipData = (isset($data)) ? $data : $dataTypeContent;

		$model = app($options->model);

        	$query = $model::where($options->column, '=', $relationshipData->id)->first();

	@endphp
```

But this is not always true.

Need some method to select an alternative child id

xembill:
Same issue for me. 
ı change manually
``
$query = $model::where($options->column, '=', $relationshipData->id)->first();
``
To
``
$query = $model::where($options->column, '=', $relationshipData->{$options->key})->first();
``

But this is not best solution.

IvandaNothabeer:
I tested the same change and it does work :) , but really, this is just as inflexible as the current functionality.

IMO The Relationship configuration needs an additional parameter for the child table field_id.

fletch3555:
This will likely not be fixed for 1.0.*

We're working on rebuilding the entire bread editor from scratch, so it'll function completely differently.  I haven't decided whether that'll find it's way into 1.1, 1.2, or 2.0 yet

IvandaNothabeer:
For us, this is a critical function. We would like to add Voyager to several existing third party apps, but there's no possibility that the database structure can be changed.

Issue #4795, opened 2020-03-01
Relationship with primary key other than 'id' 


andre4nap:
## Version information

 - Laravel: v6.2
 - Voyager: v1.3
 - PHP: 7.3.4
 - Database: [type] [version] (e.g. MySQL 8.0)

## Description

In a model with primary key like as 'slug', when we make relationship on voyager admin, the related data is empty.
The bug is in `/vendor/tcg/voyager/resources/views/formfields/relationship.blade.php` where the relationship `hasOne` and `hasMany` are loaded with `$relationshipData->id` instead of `$relationshipData->getKey()`

MrCrayon:
Related to #2648

If you have time please review #4702 that should solve the problem.

andre4nap:
use model's function `getKey()` instead of `{$options->key}` so i think the issue is solved.

PR #4796, opened 2020-03-01
Relationship with primary key other than id

andre4nap:
Make relationship with primary key other than field 'id'. Now it's possible to use also a 'slug' as primary string without problems.


andre4nap:
#4795

andre4nap:
This patch should fix the issue #2648

JackShev (approved):
great!

RazorSiM (approved):
Tested and working as for the latest Laravel version

emptynick:
Superseded by https://github.com/the-control-group/voyager/pull/4702

PR #4802, opened 2020-03-03
Add in support for Laravel 7

midnite81:
This adds in support for Laravel 7


emptynick:
Hey, thanks!
I just added testing against Laravel 7 so we can make sure everything works with it.

midnite81:
Looks as though I missed the orchestra/tb dependency -  good spot :)

emptynick:
Yeah it's always a bit painful to find the corresponding versions for all those packages.
Pushing some more things.

midnite81:
This is why I love tests - makes finding bugs way more easy :)

midnite81:
It looks like larapack/hooks only require composer/composer ~1.4 but the symphony/console isn't supported until ~1.10 (but it's in RC for now)

emptynick:
Good catch.
`composer/composer` 1.9.3 (latest stable) supports `symfony/console` > 2.7 so I will updat the dependency in `larapack/hooks`.

emptynick:
Turns out we have to wait for `composer/composer` to release something with support for `symfony/console` 5.0 for now.
Nothing we can do at the moment.

midnite81:
Doh! That's annoying isn't it.

emptynick:
Lucky us: https://github.com/composer/composer/releases/tag/1.10.0

midnite81:
That's pretty good timing :)

emptynick:
Now it looks like we have to drop support for PHP 7.2.

Edit: We have to have `laravel/ui` which requires at least Laravel 5.8 but we still support 5.5 (as it is LTS)

fletch3555 (approved):

emptynick (approved):

midnite81:
Thanks @emptynick & @fletch3555 :)

PR #4803, opened 2020-03-04
Enable case-insensitive 'LIKE' queries for non-MySQL

kingIZZZY:
Some queries use the `LIKE` operator in a case-insensitive manner, thus assuming MySQL server.
Enable also for non-MySQL databases (PostgreSQL, and possibly others?)


fletch3555:
I wasn't aware that any RDBMS system had case-sensitive keywords... can you link to this in the docs somewhere?

MrCrayon:
PostgreSQL has ILIKE for case insensitive LIKE, I don't know about other DB.

kingIZZZY:
https://www.postgresql.org/docs/10/functions-matching.html#FUNCTIONS-LIKE

> The key word ILIKE can be used instead of LIKE to make the match case-insensitive according to the active locale. This is not in the SQL standard but is a PostgreSQL extension.

Issue #4275, opened 2019-07-20
[BUG] all commands in COMPASS is not working


vonsogt:
## Version information

 - Laravel: v5.8.29
 - Voyager: v1.2.6
 - PHP: 7.3.7
 - Database: [type] [version] (e.g. MySQL 8.0)

## Description

The path in Environment on Windows not working at website which configuration file i should change?

## Steps To Reproduce
1. Go to 'Tools/Compass'
2. Click on 'Commands'
3. Scroll down to 'Any commands'
4. Run Command
5. See error

## Screenshots
![image](https://user-images.githubusercontent.com/35516476/61573562-30839400-aadb-11e9-9811-9348ab1414ce.png)

## Additional context
'xxx' is not recognized as an internal or external command,
operable program or batch file.

fletch3555:
PHP is not on your system path. Just add it

vonsogt:
> PHP is not on your system path. Just add it

I added it but, not working from web and working for CMD/Git bash
![image](https://user-images.githubusercontent.com/35516476/61574532-4dc05e80-aaeb-11e9-920c-395bac7441b9.png)

emptynick:
Normally you have to restart your computer.

fletch3555:
Did you restart the webserver (looks like xampp to me)?

vonsogt:
It's still same,
I do restart my computer and XAMPP too

![image](https://user-images.githubusercontent.com/35516476/61600823-e6c2b700-ac5c-11e9-94c8-ebb83fc2e2b3.png)

Still not working

vonsogt:
And yeah another project like "laravel-admin" cannot run any commands from Front End like this

> Starting backup...
Dumping database laravel_admin_ppat...
Backup failed because The dump process failed with exitcode 1 : General error : '"mysqldump"' is not recognized as an internal or external command,
operable program or batch file.

Really weird? any missing configuration on my computer
I'm using:
- Laravel installer : 2.1.0

to make new project then require any project from GitHub

fletch3555:
> General error : '"mysqldump"' is not recognized as an internal or external command

That's 100% saying you don't have mysql installed OR the command is not available on your system path.  Nothing to do with Laravel or whatever package you were using, and especially nothing to do with Voyager.  Since you're using xampp, my guess is it's just not on your path

vonsogt:
> > General error : '"mysqldump"' is not recognized as an internal or external command
> 
[...]

PR #4842, opened 2020-03-28
Fix Compass commands not working

hmorejong:
I think this can solve the question - issue regarding why all compass commands are not working #4275. 
Symfony made an upgrade to the Process class.


emptynick:
Thanks for your contribution.
We ended up using `Artisan` instead of `Process`.

Superseded by https://github.com/the-control-group/voyager/pull/4856

PR #4851, opened 2020-04-01
Scope relationship options

emptynick:
This allows to use a local scope on the related model in a relationship.

**Usage:**
Create a scope, for example
```
public function scopeActive($query)
{
    return $query->where('active', 1);
}
```

and add to the relationship options:

```
{
    "scope": "active"
}
```

Supersedes #4760

### DIFF
--- a/docs/bread/relationships.md
+++ b/docs/bread/relationships.md
@@ -1,5 +1,7 @@
 # Relationships
 
+## Adding a relationship
+
 Using the BREAD builder you can easily create Relationships between tables. At the bottom of the page you will see a new button that says 'Create Relationship'
 
 ![](../.gitbook/assets/bread_relationship.png)
@@ -21,6 +23,8 @@ You can also specify which columns you would like to see in the dropdown or the 
 
 Now, you can easily create `belongsTo`, `belongsToMany`, `hasOne`, and `hasMany` relationships directly in Voyager.
 
+## Advanced options
+
 If you need to set advanced options for `belongsToMany` relationship you can set, after saving relationship, these parameters in details field:
 
 ```php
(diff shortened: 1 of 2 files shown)


fletch3555 (approved):

pcorrick:
I'm getting the following error in laravel.log, when I click any dropdown BREAD field related to a relationship (Add or Edit). The field says "the results could not be loaded".

[2020-04-03 15:17:01] production.ERROR: Undefined property: stdClass::$scope {"userId":1,"exception":"[object] (ErrorException(code: 0): Undefined property: stdClass::$scope at [APPNAME]/vendor/tcg/voyager/src/Http/Controllers/VoyagerBaseController.php:860)

emptynick:
Is $options an object in your case? Or just an empty string?

pcorrick:
I'm pretty new to php/laravel, I think $options is a json formatted string, which in my case doesn't have a scope property (yet).

I commented out lines 860-862  in VoyagerBaseControler.php and I can see the dropdown contents again.

Then I uncommented those lines and put in a scope property (empty string) into the relationship properties, and I can see the dropdown contents for that relationship (but not any others).

Relationship details is now:
{
    "display": {
        "width": "4"
    },
    "scope": ""
}

MrCrayon:
```php
!empty($options)
```
should cover all cases without throwing an error, 0 is not used anyway.

emptynick:
I think `property_exists($options, 'scope')` should work better.

MrCrayon:
If we are sure `$options` is always an object

MrCrayon:
https://github.com/emptynick/voyager/blob/8e049d196669e62002ae74391fb6b79624d44a67/src/Models/DataRow.php#L71

emptynick:
Yes, there's an accessor in the datarow model decoding it.

mjomble:
I also encountered this error just now.

A fix that worked for me was editing line 860 in VoyagerBaseController and changing

`if ($options->scope && ...`

to

`if (isset($options->scope) && ...`

emptynick:
Fixed

PR #4856, opened 2020-04-03
Fix compass commands not working

emptynick:
Uses Artisan instead of Process.
Also removed some legacy workarounds for `redirect` and `response` and added some basic tests.


MrCrayon (approved):
Works just fine :+1:

PR #4858, opened 2020-04-04
Relationship result sorting

### LINKED ISSUE #3617: How to sort relationship results
## Description
There is no documentation on how to control the order of data in the select control. 

For example:
I have a me_device_types table with the columns **id, me_device_type** and **order**.

In my BREAD for me_devices, I have the relationship setup as shown here:
image

I want the result sorted by **order**

How can I do this?

emptynick:
This allows relationship results to be sorted by defining the following in the options:

```
"sort": {
    "field": "my_field",
    "direction": "asc"
}
```

Fixes #3617


src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -891,6 +891,21 @@ public function relation(Request $request)
                     ];
                 }
 
+                // Sort results
```
MrCrayon, line 898: It works but I'd prefer to be implemented checking first the mandatory field like this:
```php
// Sort results
if (!empty($options->sort->field)) {
    if (!empty($options->sort->direction) && strtolower($options->sort->direction) == 'desc') {
        $relationshipOptions = $relationshipOptions->sortByDesc($options->sort->field);
    } else {
        $relationshipOptions = $relationshipOptions->sortBy($options->sort->field);
    }
}
```

MrCrayon (changes requested):
Ordering works fine

docs/bread/relationships.md
```diff
@@ -35,6 +35,22 @@ If you need to set advanced options for `belongsToMany` relationship you can set
...
 
+## Sorting relationships
+
+You can sort the results of a relationship by setting the `sort` object in the relationship options:
+
+```php
+{
+    "sort": {
+        "field": "my_field",
+        "direction": "asc"
```
MrCrayon, line 46: Since `asc` is the default I would put `desc` here

docs/bread/relationships.md
```diff
@@ -35,6 +35,22 @@ If you need to set advanced options for `belongsToMany` relationship you can set
...
+    }
+}
+```
+
+This will sort the results ascending by the field `my_field`. Use `desc` as the direction to sort it in the opposite order.
```
MrCrayon: If setting `desc` in the example this can be changed as:
```
This will sort the results descending by the field `my_field`. Default value for direction is `asc`.
```

src/Http/Controllers/VoyagerBaseController.php
MrCrayon: Why the adversion on using `empty`? :smile: 
I think it fits perfectly in this case.

WIth your conditions you are still not excluding when `$options->sort->field` is an empty string, it's not going to give an error but still it also not going to do anything inside that block of code.

MrCrayon:
You can add Closes #3617

emptynick:
Done

src/Http/Controllers/VoyagerBaseController.php
fletch3555: yeah, I'm inclined to agree with @MrCrayon.  `empty()` seems like the obvious choice here.

emptynick: Ok ok, another try.

MrCrayon: Ah! You are still missing the beauty of `empty` :cry: , you really don't need to check all the steps just use it as in my previous snippet
```php
if (!empty($options->sort->field)) {
```

emptynick: Err strange. That behaved different in previous PHP versions...well

MrCrayon (approved):
:+1: 
Not sure if we need to specify that HasMany relationships won't be ordered.

PR #4869, opened 2020-04-06
Fix redirect to browse not working with scopes

MrCrayon:
After updating, if there is a scope applied to BREAD, user is not redirected to the list page even if it has browse permission set.


emptynick (changes requested):
I think we also need to do this here:
https://github.com/the-control-group/voyager/blob/3b35f201f0ebe768d18e4077b0a6e46dc048c1df/src/Http/Controllers/VoyagerBaseController.php#L419

MrCrayon:
> I think we also need to do this here:

That's not needed because there is no scope in store method.
Not sure if scope needs to be added but it's outside this PR scope :grin:

emptynick:
Yeah sorry, did not consider this.
Just saw that there is the same part in the code

emptynick (approved):

PR #4875, opened 2020-04-06
Fix additional attributes when using local scope

arxeiss:
Fix error: `Property [additional_attributes] does not exist on the Eloquent builder instance.`

This happens, when a local scope is used. Then `$model` does not contain Model but Builder.


src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -868,7 +870,7 @@ public function relation(Request $request)
                 // If search query, use LIKE to filter results depending on field label
                 if ($search) {
                     // If we are using additional_attribute as label
-                    if (in_array($options->label, $model->additional_attributes ?? [])) {
```
MrCrayon, line 871: I don't understand, you just moved this.
How is it changing anything?

emptynick: The scope is applied in between his changes.
$model changes it's type then

arxeiss: When you apply local scope by this `$model = $model->{$options->scope}();` then `$model` does not contains model instance but Builder instance which then does not support `additional_attributes`.

Try to create relation in BREAD with local scope. When you try to select it, it works but when you try to search/filter, it doesn't work. See my MR description, there is original message which is thrown

MrCrayon (commented):
I see thanks

emptynick:
I'm more curious if the accessor is available on a `Builder` instance then?

moisesh18:
This worked for me. Why it is not in merged?

moisesh18 (approved):

emptynick (approved):

PR #4876, opened 2020-04-07
Fix saving relations

arxeiss:
This is not merge ready probably. I just had problems with BelongsTo relation. They were not saved and I got always error, that `category_id` doesnt have default value etc.

### DIFF
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -23,7 +23,7 @@
             @else
 
                 <select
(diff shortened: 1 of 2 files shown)


MrCrayon:
Those change are for sure wrong, if it's not saving you probably need to check edit on foreign key in BREAD settings.

You can read here https://github.com/the-control-group/voyager/pull/4729/files

arxeiss:
Oh man! Totally missed this new thing in v1.4! Now it works as expected. Thank you! Closing this MR

PR #4879, opened 2020-04-09
fix bread radio_btn formfield list undisplay bug

Feiir:
parse error occurs when the data of radio is list ,fix it


MrCrayon:
Can you post an example on how you are defining radio options in BREAD details for the error to occur?

Feiir:
BREAD details eg :{
    "default": 0,
    "options": [
        "open",
        "close"
    ]
} 
this triggers the problem

emptynick:
It shouldn't be an array but an object, as written in the docs.

Feiir:
ok i got it

PR #4881, opened 2020-04-09
Fix voyager models not returning save result

### LINKED ISSUE #4880: Error still show up after successfully restored a post
## Version information

 - Laravel: v7.0
 - Voyager: v1.4
 - PHP: 7.3
 - Database: MySQL 5.5.5-10.4.12-MariaDB

## Description

I create a new Post model extends the Voyager Post, and then I add SoftDeletes to the new Post model. The restore function works fine but it still shows an error message: "Sorry it appears there was a problem restoring this Post". 

I do the same thing to User model, and the restore function works perfectly.

I try to dump the $res variable in VoyagerBaseController.php (line 511) and the result is true for the User, null for the Post.

My Post model looks like this:

_namespace App;

use Illuminate\Database\Eloquent\SoftDeletes;

class Post extends \TCG\Voyager\Models\Post
{
    use SoftDeletes;
}_

## Steps To Reproduce
1. Add deleted_at field to the posts table
2. Add Post model extends Voyager Post, change the model name of Post bread to "App\Post"
3. Delete then restore the post
4. See error: "Sorry it appears there was a problem restoring this Post"

## Expected behavior
It should show "Successfully restored post"

## Screenshots
image

MrCrayon:
Result is needed while checking on restore, it's also how parent method is defined.
https://github.com/laravel/framework/blob/55070d4e57abd710783d29329d6f6846b570f452/src/Illuminate/Database/Eloquent/Model.php#L666

Fixes #4880

### DIFF
--- a/src/Models/Page.php
+++ b/src/Models/Page.php
@@ -34,7 +34,7 @@ class Page extends Model
             $this->author_id = Auth::user()->getKey();
         }
 
-        parent::save();
+        return parent::save();
     }
 
     /**
--- a/src/Models/Post.php
+++ b/src/Models/Post.php
@@ -27,7 +27,7 @@ class Post extends Model
             $this->author_id = Auth::user()->getKey();
         }
 
-        parent::save();
+        return parent::save();
     }
 
     public function authorId()


emptynick (approved):
👍

Issue #3172, opened 2018-05-12
Multilanguage not working. How to add or override language ?


mrtieutien90:
- Laravel Version: 5.6.*
- Voyager Version: 1.1.1
- PHP Version:7.1.3
- Database Driver & Version:

### Description:
Multilanguage not working  . How to add new or override language ?
I try config files.
config/voyager.php

```
'multilingual' => [
        /*
         * Set whether or not the multilingual is supported by the BREAD input.
         */
        'enabled' => true,

        /*
         * Select default language
         */
        'default' => 'vi',

        /*
         * Select languages that are supported.
         */
        'locales' => [
            'vi',
            //'pt',
        ],
    ],
```

and config/app.php 

```
return [
    ...
    'locale' => 'vi',
    ....
];
```

---- 
I created folder lang/vi . see image.
http://www.upsieutoc.com/image/45hwiN

### Steps To Reproduce:
Install Laravel with Voyager
Change multilingual in config/voyager.php
Change locales in config/app.php

emptynick:
This is described in the [Laravel docs](https://laravel.com/docs/5.6/localization#overriding-package-language-files)

mrtieutien90:
Thanks for answer @emptynick ,
But Not working. I can not override it . Maybe  Voyager not accept override language or it error.

fletch3555:
First of all, if you get an error, it's immensely helpful to provide the actual error message and stacktrace.

For your issue, it's likely because you're only adding the translations to your app.  For Voyager to be able to find them, you need to add them to the Voyager namespace (`resources/lang/vendor/voyager/vi/`)

mrtieutien90:
Thanks a lot.
I just want to use Vietnamese.
pls see images : http://www.upsieutoc.com/image/48w8XA 
I have read the voyager's document follow the instructions but it only accepts english language.

Thanks.

hungxinh:
@mrtieutien90,
Nếu bạn chỉ dùng ngôn ngữ Vietnamese (VI), bạn copy vào đường dẫn sau là xong.
Root: /vendor/tcg/voyager/publishable/lang/vi

If you only use the Vietnamese language (VI), you just copy folder VI into the following path is finished:
Root: /vendor/tcg/voyager/publishable/lang/vi

It work fine!!!

![screenshot 2018-05-16 09 01 49](https://user-images.githubusercontent.com/23396671/40092674-b8afb3f4-58e8-11e8-8bc7-038122585425.png)

[...]

PR #4889, opened 2020-04-14
Update multilanguage.md

josepcrespo:
In order to avoid headaches for others trying to implement Voyager's multilingual feature. I thought it would be positive to improve this part of the documentation. Some people had this problem in 2018 and today it is still happening in the current version (1.4).

I found this thread that helped me solve the problem in my own project:

https://github.com/the-control-group/voyager/issues/3172

I hope that you can accept the changes so that you can help others in the future. Awesome project, thanks!

### DIFF
--- a/docs/core-concepts/multilanguage.md
+++ b/docs/core-concepts/multilanguage.md
@@ -28,10 +28,10 @@ class Post extends Model
 }
 ```
 
-After following the previous indications, in order to see the language-selection on your BREAD pages, you need to run the Laravel Artisan command for removing the configuration cache file:
+After following the previous indications, if you don't see the language-selection on your BREAD pages, you need to run the Laravel Artisan command for removing the configuration cache file:
 
 ```bash
-php artisan config:cache
+php artisan config:clear
 ```
 
 Now you will see a language-selection in your BREAD pages.
--- a/docs/core-concepts/multilanguage.md
+++ b/docs/core-concepts/multilanguage.md
@@ -28,7 +28,13 @@ class Post extends Model
 }
 ```
 
-Now you will see a language-selection in your Pages BREAD.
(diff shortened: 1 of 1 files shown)


fletch3555:
These instructions are actually incorrect.  `artisan config:cache` will actually set the cache, not clear it

MrCrayon:
To what @fletch3555 said I'd add that config is not cached by default and should be cached only during deployment process.
If you cached your config you should know that you need to clear cache when you apply a change to config just like you need to clear route cache if you add a route.
If we don't consider this basic knowledge we should add that cache should be cleared if cached everywhere else in documentation that referers to config changes.

josepcrespo:
> These instructions are actually incorrect. `artisan config:cache` will actually set the cache, not clear it

Sorry, my bad. I copy/pasted the Artisan description for "config:clear" but wrote the incorrect Artisan command. The correct one should be "config:clear" instead of "config:cache".

josepcrespo:
> To what @fletch3555 said I'd add that config is not cached by default and should be cached only during deployment process.
> If you cached your config you should know that you need to clear cache when you apply a change to config just like you need to clear route cache if you add a route.
> If we don't consider this basic knowledge we should add that cache should be cleared if cached everywhere else in documentation that referers to config changes.

Ok, maybe is a basic knowledge for you, but if you are reading the documentation:

1. It's your first time dealing with that problem or,
2. It's been a long time since you did it previously and, probably you don't remember that "basic knowledge".

Documentation is for… well documenting precesses… it seems clear that I'm not the only one confused about this and, instead of reading a simple Artisan command on the official documentation I had to waste time searching on Google and digging on the GitHub repository issues from two years ago…

I've reformulated the sentence, maybe you can reconsider the pull request now.

josepcrespo:
Instead of accepting a simple pull request you prefer to just throw a personal opinion and waste your precious time writing three long sentences arguing it's unnecessary for you. Very productive 👍

Issue #3027, opened 2018-04-21
Slugify for Voyager - Russian map 


Margo57:
- Laravel Version: 5.5
- Voyager Version: 1.0
- PHP Version: 7.2
- Database Driver & Version:

### Description:

In Russian, "ь": "" and "ъ": "", but they are replaced by "-". It is not right!
Печенье = Pechene, not Pechen-e...

taiwanleaftea:
I have the same issue in the latest version of Voyager. 

## Workaround

_map_russian function in resources/assets/js/slugify.js for Russian language contains this for 'ъ', 'ь', 'Ъ', 'Ь':

` _map_russian: function() {
                return {<skipped>
                   'ъ':'', 'ь':'', 
                    'Ъ':'', 'Ь':''
<skipped>
                };
            },`

So **_slug += (this.chars[str.charAt(i)]) ? this.chars[str.charAt(i)] : str.charAt(i)** doesn't replace those letters, and the letters are replaced later for '-' in this regexp **replace(/[^a-z0-9]/g, _sep)**

Adding this regexp **replace(/[ьЬъЪ]/g, "")** solve the problem.

PR #4899, opened 2020-04-22
Slugify for Russian charmap

taiwanleaftea:
Slugify for Voyager - Russian map #3027

### DIFF
--- a/resources/assets/js/slugify.js
+++ b/resources/assets/js/slugify.js
@@ -100,7 +100,8 @@
                 }
 
                 str = _slug
-                .replace(/[^a-z0-9]/g, _sep)
+                .replace(/[ьЬъЪ]/g, '')
+                .replace(new RegExp('[^a-z0-9]', 'g'), _sep)
                 .replace(new RegExp('\\'+_sep+'\\'+_sep+'+', 'g'), _sep)
                 .replace(new RegExp('^\\'+_sep+'+|\\'+_sep+'+$', 'g'), '');
 


emptynick:
I think it's better to "fix" the way those characters are replaced.
This way we keep language-specific things inside the array.

taiwanleaftea:
I agree, but the problem is, that 'ь' and 'ъ' should be just removed from the transliterated string: 
i.e. платье -> plate or олень -> olen
So the character map should have special token to remove letter from the sting.

MrCrayon:
@taiwanleaftea Can't we just set `'ь':''` and other cases in __map_russian_?
https://github.com/the-control-group/voyager/blob/af482b70eda00b94b8e18bb2d7f60e6d6a3b0333/resources/assets/js/slugify.js#L205

Edit:
looks like it's already there :thinking:

taiwanleaftea:
It's already set, but the "replace charmap cycle" doesn't replace this letters and leave them "as is". Later **regexp replace(/[^a-z0-9]/g, _sep)** replaces the letters for separator.

MrCrayon:
Then I think we should fix that, most probably the problem is here:
https://github.com/the-control-group/voyager/blob/af482b70eda00b94b8e18bb2d7f60e6d6a3b0333/resources/assets/js/slugify.js#L97
That should pass if `this.chars[str.charAt(i)]` is empty so we need to change how it's checked.

MrCrayon:
This should work:
```javascript
_slug += (this.chars.hasOwnProperty(str.charAt(i))) 
```

taiwanleaftea:
Yes, it works. Will you make the changes or should I do PR?

PR #4900, opened 2020-04-23
Fix slugify not replacing to empty string

taiwanleaftea:
fix for Slugify for Voyager - Russian map #3027


MrCrayon:
Thanks for the submission, since you are first time contributor a couple of advices:

- Use a descriptive title like: _Fix slugify not replacing to empty string_, so who is going to compile the release notes can use that.
- Add labels, in this case bugfix
- When PR fixes a bug put in description: Fixes #3027 (or Closes for features) that will automatically close related issue when merged.

You can edit both title and description, also if you need to change your PR you can just push new commits to your branch.

PR #4909, opened 2020-04-28
Slugify filename 

cotiga:
Slugify filename uploaded with BREAD when detail is set with "preserveFileUploadName": true
If "preserveFileUploadName": true is set without slugify, uploaded files can causes W3C errors (spaces, non autorised caracters...)
And is good for SEO :p


emptynick:
Don't mind StyleCI. It will merge all fixes once we merge this in!

cotiga:
> Don't mind StyleCI. It will merge all fixes once we merge this in!

Thank you, I'll know for the next time

PR #4922, opened 2020-05-06
Fix validation for array elements

MrCrayon:
Fixes validation when array does not contain files, ex. belongsToMany relationship.

```
Symfony\Component\Debug\Exception\FatalThrowableError
Call to a member function getClientOriginalName() on string
```


fletch3555 (approved):

EduCavalheiro:
Tested, it works well.

emptynick (approved):

Issue #4945, opened 2020-05-20
Array validation error with Relationship Fields


eddielee394:
## Version information

 - Laravel: v7.10
 - Voyager: v4.1, v4.1.x-dev
 - PHP: 7.3
 - Database: [mysql] [5.7.19] (e.g. MySQL 8.0)

## Description

When applying validation rules to a belongsToMany or other array based relationships, the validation rules are applied to the individual elements in the array & not to the parent field.  This causes some unexpected behavior with validation rules like `max`, `min`, `between`.  

## Steps To Reproduce
1. Create belongsToMany relation
2. Add `max`validation rule in "Relationship Details" bread field
3. Add relations in add/edit view

## Expected behavior
In the case of relations, validation should be applied to the relationship field and in the case of arrays, the array should be validated in addition to the elements within the array. 

## Additional context
Line causing the issue can be found here:  https://github.com/the-control-group/voyager/blob/cbf9ceed421e6153323f659d502b5694b16c0b3b/src/Http/Controllers/Controller.php#L222
Issue was reproducible & confirmed with @emptynick .  PR to be made with a suggested workaround.

PR #4946, opened 2020-05-20
- fix issue with relationship validations;

eddielee394:
See https://github.com/the-control-group/voyager/issues/4945

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -200,9 +200,10 @@ abstract class Controller extends BaseController
         foreach ($fieldsWithValidationRules as $field) {
             $fieldRules = $field->details->validation->rule;
             $fieldName = $field->field;
+            $isRelationshipField = $field->type === 'relationship';
 
             // Show the field's display name on the error message
-            if (!empty($field->display_name)) {
+            if (!empty($field->display_name) && !$isRelationshipField) {
                 if (!empty($data[$fieldName]) && is_array($data[$fieldName])) {
                     foreach ($data[$fieldName] as $index => $element) {
                         if ($element instanceof UploadedFile) {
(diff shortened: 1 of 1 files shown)


MrCrayon:
I think it would be better if we can manage to keep both validations for array and elements.
In relationships validation for elements is probably less common but it can still be used and for other arrays we are not solving the problem at all. 

I was thinking of adding a prefix to rule name if we want to be applied to array, maybe: 
```
{
    "validation": "[]required|[]max:2|unique:tasks"
}
```

Thoughts?

eddielee394:
>I think it would be better if we can manage to keep both validations for array and elements.

@MrCrayon That was my initial thought as well.  Either prefix like you suggested, or allow for an additional key in the validation options with the field name for the parent and the element validation then would get handled with the existing code.  Similar to the way the add/edit options are handled currently. I feel like this would be a bit more inline with the laravel conventions for validations.  Something along the lines of:  
```
{
    "validation": {
        "rule": "min:3",
		"relation_field_name":{
			"rule": "max:3"
		},
        "edit": {
            "rule": "nullable"
        },
        "add": {
            "rule": "required"
        }
    }
}
```


Also worth noting, I did do some quick debugging after implementing the workaround and somehow the elements for the btm relation field were validating properly with the `in` and `exists` rules.  Not sure if there's validation logic elsewhere that picked it up or something else is going on under the hood as I didn't dig too deeply into that controller method.

MrCrayon:
I like the idea of nested rules, I discharged it for now because it's going to bring back application of rule on array by default and items only if specified and that would be a breaking change.
But I agree it looks cleaner, that would allow to do:
```json
{
    "validation": {
        "rule": "array|min:3",
        "*.order": {
            "rule": "integer"
        },
    }
}
```
For the rule on belongsToMany I was thinking about the unique rule on pivot table, but that's actually broken anyway in this case.

I think for now your fix should do with the drawback that if someone is relying on a custom rule set for belongsToMany items that will break.

This should probably be skipped too:
https://github.com/the-control-group/voyager/blob/cbf9ceed421e6153323f659d502b5694b16c0b3b/src/Http/Controllers/Controller.php#L206

eddielee394:
Yeah, i like the items as the nested key vs the field as nested key much better.  Seems to be way more intuitive.  
```
{
    "validation": {
        "rule": "array|min:3",
        "*.order": {
            "rule": "integer"
        },
    }
}
```
Maybe add to the roadmap for the next major release? 

For now I updated the commit with the suggested change & assigned the relationship field type to a variable so the conditions would be a bit more expressive.

PR #4964, opened 2020-06-02
Refactoring the login page to allow template usage

akiyamaSM:
Allow to use the login template in other pages such as Reset password if needed

```
@extends('voyager::auth.master', ['title' => 'Reset Password'])

@section('content')

@endsection
```

### DIFF
--- a/resources/views/auth/master.blade.php
+++ b/resources/views/auth/master.blade.php
@@ -1,0 +1,75 @@
+<!DOCTYPE html>
+<html lang="{{ config('app.locale') }}" dir="{{ __('voyager::generic.is_rtl') == 'true' ? 'rtl' : 'ltr' }}">
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="robots" content="none" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimal-ui">
+    <meta name="description" content="admin login">
+    <title>Admin - {{ $title }}</title>
+    <link rel="stylesheet" href="{{ voyager_asset('css/app.css') }}">
+    @if (__('voyager::generic.is_rtl') == 'true')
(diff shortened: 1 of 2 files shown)


akiyamaSM:
Hello @emptynick Hope it doesn't bother you taking look at this PR

emptynick:
This would be a breaking change.
Why not simply add `@section('...')` and `@show` around the login-container in `login.blade.php` to make it overridable?

akiyamaSM:
Why is a breaking change?
And in my case i have pages where i need to reset the password and I have to change in the title too.

emptynick:
Because other people might have overriden this view as well.
The title could also be just a section/yield

akiyamaSM:
I really think it should be out of the box. But since its a breaking change. Hope it can make to an other major version.

emptynick:
But like I said before, it can be done much more easy with ~3 lines of code.

MrCrayon:
As @emptynick said title could be:
```blade
@yield('title', 'Admin - '.Voyager::setting("admin.title"))
```
That way we can keep compatibility and it's easily overridable.

akiyamaSM:
I really don't see why it shouldn't be part of its core. I believe that reseting password's flow should be a part of the package.

MrCrayon:
I didn't say I'm against this PR, personally I think it makes sense.

I'm only saying that I like more the use of `yield` for title.
But I have to correct myself, that's not a breaking change because right now users either keep the template or replace it completely.

akiyamaSM:
@mrCrayon thanks, Ill make the change about the yield to fit the requirements

akiyamaSM:
Hello @MrCrayon Changes has been made :D

akiyamaSM:
Any ups ?
@MrCrayon 
@emptynick

MrCrayon (approved):
I'm good with this, now you have to convince @emptynick :grin:

akiyamaSM:
@MrCrayon hahaha thanks.

akiyamaSM:
Hey @emptynick Hope we can get an answer, I have an other PR that I want to submit for a fix.

emptynick:
Always create a sub-branch from 1.4 so you can create multiple PRs at the same time.

PR #4970, opened 2020-06-04
Fix picking media not updating thumbnails

### LINKED ISSUE #4969: Selecting Image using Image Picker does not update thumbnail until clicking another image
## Version information
 - Laravel: v7.11.0
 - Voyager: v1.4.1
 - PHP: 7.3.13
 - Database: [MySQL 5.7.22 (e.g. MySQL 8.0)

## Description
Note that this is a UI bug, when double clicking, or clicking an image & adding selected the thumbnail appears not to update, when selecting save, functionality works as expected. The issue being that when selecting images in the image picker, the thumbnail above does not update until selecting another image, which is not expected behaviour.

## Steps To Reproduce
1. Log into admin portal
2. Assign an image picker field to a BREAD model
3. Upload multiple images.
4. Select one of the images to save
5. Click back & forth between multiple images and try to add them to thumbnail

## Expected behavior
When double clicking, or selecting an image & selecting "add all selected" using an image picker, the thumbnail above will update.

## Screenshots
Please see GIF...

Image Picker UI Bug

emptynick:
When a file was double-click to pick it, the UI didn't refresh.
This fixes it.
Also fixed a problem where images weren't displayed when the extension was uppercase (ex. `myfile.JPG`)

Fixes https://github.com/the-control-group/voyager/issues/4969

### DIFF
--- a/resources/views/media/manager.blade.php
+++ b/resources/views/media/manager.blade.php
@@ -540,7 +540,7 @@
             fileIs: function(file, type) {
                 if (typeof file === 'string') {
                     if (type == 'image') {
-                        return this.endsWithAny(['jpg', 'jpeg', 'png', 'bmp'], file);
+                        return this.endsWithAny(['jpg', 'jpeg', 'png', 'bmp'], file.toLowerCase());
                     }
                     //Todo: add other types
                 } else {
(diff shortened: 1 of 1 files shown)


MrCrayon (approved):
Works :+1: 
This is the reason I thought it was buggy but I never looked into it properly.

PR #4975, opened 2020-06-11
Fixes issue with NULL primary keys

lesichkovm:
When a primary key column has a value set to NULL, Voyager throws an unrelated exception message i.e. Missing required parameters for [Route: voyager.bread.show] [URI: admin/bread/{id}] 

This fix will throw a helpful message pointing the user into the right direction

### DIFF
--- a/src/Actions/EditAction.php
+++ b/src/Actions/EditAction.php
@@ -28,6 +28,12 @@ class EditAction extends AbstractAction
 
     public function getDefaultRoute()
     {
-        return route('voyager.'.$this->dataType->slug.'.edit', $this->data->{$this->data->getKeyName()});
+        $id = $this->data->{$this->data->getKeyName()};
+
+        if (is_null($id)) {
+            throw new \RuntimeException('Primary key in ' . $this->dataType->model_name . ' CANNOT BE null');
+        }
+        
+        return route('voyager.'.$this->dataType->slug.'.edit', $id);
     }
 }
--- a/src/Actions/ViewAction.php
+++ b/src/Actions/ViewAction.php
@@ -28,6 +28,12 @@ class ViewAction extends AbstractAction
 
     public function getDefaultRoute()
     {
-        return route('voyager.'.$this->dataType->slug.'.show', $this->data->{$this->data->getKeyName()});
+        $id = $this->data->{$this->data->getKeyName()};
+
+        if (is_null($id)) {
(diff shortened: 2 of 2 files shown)


MrCrayon:
It's not Voyager but Laravel throwing the exception. 
A primary key set to NULL means your data is invalid and also you didn't set your database field properly.
IMHO we shouldn't change anything.

lesichkovm:
> IMHO we shouldn't change anything.
Sorry to disagree but its a very bad practice to use any variable without checking its valid first. This pull request does validate the column specified as primary key in the model, does not have a NULL value

MrCrayon:
Value is already checked by Laravel, that's why you see an Exception.
You are only moving the exception and changing message to fit your case but it's much more likely that the primary key is not named as `id` than having a primary key set as NULL, that in most if not all databases supported can't even happen.

lesichkovm:
Laravel does not check the primary key. It is Viager that requires the PK to not be NULL. Which is why it is Viager that throws the exception - **Missing required parameters for [Route: voyager.bread.show] [URI: admin/bread/{id}]**

As a result Viager has to do the check before it tries to use it.

MrCrayon:
https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/src/Illuminate/Routing/Exceptions/UrlGenerationException.php#L17

lesichkovm:
This is exactly what proves my point. When you see **Missing required parameters for [Route: voyager.bread.show] [URI: admin/bread/{id}]** how does this actually help the user? I had to dig inside the Voayger code to find out what is going on. This is not what you would like a regular Voyager user to do, do you?

MrCrayon:
> This is exactly what proves my point.

You were saying it's not Laravel that checked primary key, that shows otherwise.

> When you see **Missing required parameters for [Route: voyager.bread.show] [URI: admin/bread/{id}]** how does this actually help the user?

Because if you search Voyager "Missing required parameters for", forth link is Voyager documentation explaining that you don't have a primary key.

emptynick:
Those are all [Laravel/Eloquent conventions](https://laravel.com/docs/7.x/eloquent#eloquent-model-conventions).
We can not cover all cases otherwise our code would be 99% checking for bad habits.

lesichkovm:
> We can not cover all cases otherwise our code would be 99% checking for bad habits.
Beg to disagree with this statement. Any variable should be propely checked before attempting to use it.

PR #4976, opened 2020-06-11
Update missing-required-parameter.md

lesichkovm:


MrCrayon:
In which database a primary key can be set to NULL?

lesichkovm:
In any database - MySQL, SQLite

In my case it was an old MySQL table. The ID was varchar for UUID. There were rows with NULL IDs, and the errors thrown by Voyager were not helpful.

MrCrayon:
AFAIK in MySQL a primary key can't be NULL, if you had NULL values it means field was not set as primary key.
That should still fall in the "you don't have a primary key" case.

PR #4985, opened 2020-06-18
Fix SQL Issue with multiple orderBy('order')

tjventurini:
Fixing SQL Issue: A column has been specified more than once in the order by list. Columns in the order by list must be unique.

Just had this issue on a MsSQL server. Whenever the menu items should be shown, I was facing the error from above. And I figured out, that there is indeed the `orderBy('order')` applied multiple times: Once in `Menu::display()` and the other time in the relationship itself, on `MenuItem::children`. 

Since I think it makes somehow sense to sort the MenuItems per default, I decided to remove the `orderBy` in the `Menu::display` method. But this MR is just to share the awareness of the issue, so feel free to do something else to fix the issue 😉

Have a great day! 😁

### DIFF
--- a/src/Models/Menu.php
+++ b/src/Models/Menu.php
@@ -55,7 +55,9 @@ class Menu extends Model
         // GET THE MENU - sort collection in blade
         $menu = \Cache::remember('voyager_menu_'.$menuName, \Carbon\Carbon::now()->addDays(30), function () use ($menuName) {
             return static::where('name', '=', $menuName)
-            ->with(['parent_items.children'])
+            ->with(['parent_items.children' => function ($q) {
+                $q->orderBy('order');
+            }])
             ->first();
         });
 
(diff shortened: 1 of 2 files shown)


MrCrayon:
Is ordering working as expected with this PR?

I see that order was introduced first in #1811 and then it was introduced in `MenuItem::children` with #4186.

P.S.
It looks like it should work because `display` is also calling `children` but I didn't test it.

tjventurini:
Looking good for me :smiley: 

Bildschirmfoto von 2020-06-19 09-11-32

The only difference is, that `orderBy('order')` is only applied once now.

tjventurini:
Is this going to be merged or will it be fixed somewhere else?

emptynick:
I think the order should be applied in `display()` not `children()`.

tjventurini:
Either way, It's probably gonna take some time until it's merged so feel free to update it meanwhile.

emptynick:
Where is that second ordering tho?
Is it this one?
https://github.com/the-control-group/voyager/blob/8d60fe15d5b51393edf07117d347de16e0f0f4cd/src/Models/Menu.php#L74


Edit: Nevermind. In the model.

tjventurini:
awesome, thx 👍

emptynick (approved):

PR #4988, opened 2020-06-24
Allow thumbnails in BREAD Image to fit using the position option

fsmonter:
When using the "crop" option in a BREAD Image FormField it actually uses the Intervention fit method but only using the width and height options.

This PR will allow the user to define the `position` of the [fit](http://image.intervention.io/api/fit) with this options:

```JSON
{
    "thumbnails": [
        {
            "name": "cropped",
            "crop": {
                "width": "350",
                "height": "250",
                "position": "top-left" 
            }
        }
    ]
}
```

### DIFF
--- a/docs/bread/formfields/images.md
+++ b/docs/bread/formfields/images.md
@@ -21,7 +21,8 @@
             "name": "cropped",
             "crop": {
                 "width": "300",
-                "height": "250"
+                "height": "250",
+                "position": "center" // Optional. Refer to http://image.intervention.io/api/fit
             }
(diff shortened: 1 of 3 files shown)


emptynick:
We'll definitely have to document this.

emptynick (changes requested):
Please also make those changes for MultipleImages

fsmonter:
Maybe we need to update the option `crop` to `fit` so it's the same as the MediaPicker options. But I think that would be a breaking change.

PR #5001, opened 2020-07-03
Menu model display method override bugfix

stuzer05:
Hello.

In the Voyager [docs](https://voyager-docs.devdojo.com/customization/overriding-files) it is a great example on how to override Voyager models. Unfortunately, Menu::processItems() method was made private, so the only way to override Menu->display() method is to copy-paste Menu::processItems() into the new class.

This commit makes Menu::processItems() method protected.


emptynick (approved):
Thanks!

PR #5009, opened 2020-07-13
Greek translation

fractalbit:
Hello, i translated voyager to the Greek language (i am a native speaker). It is a complete translation, i only left some terminology intact that didn't make much sense to translate. I was also careful not to translate special directives like :string or \_START\_.

I absolutely **love** voyager and i hope you will accept my translation. Thank you!


emptynick:
Thanks for your contribution!
Are you finished or do you want to push some more things?

fractalbit:
Thank you for the quick reply!

I just spotted a small correction I need to make, but, since it is my first
ever PR I am not sure what to do.
Should I just "git push origin greek-translation" and the PR will be
updated?

Also, if i spot more mistakes in the future should i make a new PR after
this has been accepted?

Thank you for your time!

Στις Δευ, 13 Ιουλ 2020 στις 7:01 μ.μ., ο/η Christoph Schweppe <
notifications@github.com> έγραψε:

emptynick:
You can just push commits like you did before and it will be updated here automatically.
Once it's merged you have to create a new PR.

Let me know when you are done

fractalbit:
I think now it is ready. Thank you.

emptynick (approved):

publishable/lang/el/generic.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+return [
+    'is_rtl'                 => 'όχι',
```
MrCrayon, line 4: This field is used to check if language is rtl, so even if it's still going to work as far as is not literal `'true'` it's probably better if it's set as `'false'`.

MrCrayon (commented):
If you can submit another PR with this fix would be great. 
Thanks!

publishable/lang/el/generic.php
emptynick: Fixed [here](https://github.com/the-control-group/voyager/commit/aa44a8e40c72d4282e31f2df9637291c8dc069e1)

PR #5026, opened 2020-07-29
Update docs to include extra step for overriding models

juliolopez78:
Hello,

Yesterday I ran into an issue while trying to override the Voyager Menu model in order to add soft deletes to it. It was a very simple thing I overlooked and thought I'd help out by adding what was missing to the docs as per @emptynick's suggestion

Hope you find this helpful!

Julio


MrCrayon (changes requested):
The overriding method is generic for Voyager core Models, most of them like DataRow do not have a BREAD associated.
Maybe it should be something like: 
If the model you are overriding has an associated BREAD go to BREAD settings...

MrCrayon:
BTW thanks for PR

juliolopez78:
Thanks for the feedback @MrCrayon! Yes, that totally makes sense. Will edit and push it up :wink:

juliolopez78:
@MrCrayon, changes are made. Please let me know if you have any other feedback.

MrCrayon (approved):
Looks good to me :+1: 

Thanks!

PR #5030, opened 2020-08-01
Fix: display (Read) of multiple_images field for empty array

igordelorenzi:
When all images were removed from the array, no empty state message was displayed.

Before:
Before

After:
After

### DIFF
--- a/resources/views/bread/read.blade.php
+++ b/resources/views/bread/read.blade.php
@@ -57,16 +57,12 @@
                                      src="{{ filter_var($dataTypeContent->{$row->field}, FILTER_VALIDATE_URL) ? $dataTypeContent->{$row->field} : Voyager::image($dataTypeContent->{$row->field}) }}">
                             @elseif($row->type == 'multiple_images')
                                 @php $image_files = json_decode($dataTypeContent->{$row->field}); @endphp
-                                @if (is_array($image_files))
-                                    @if (count($image_files) === 0)
-                                        {{ __('voyager::generic.none') }}
-                                    @else
-                                        @foreach ($image_files as $file)
-                                            <img class="img-responsive"
-                                                 style="display: inline-block; margin-top: 4px;"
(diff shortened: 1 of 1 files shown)


emptynick:
This can be simplified with a [forelse](https://laravel.com/docs/7.x/blade#loops) loop

igordelorenzi:
> This can be simplified with a [forelse](https://laravel.com/docs/7.x/blade#loops) loop

Nice tip, @emptynick, I appreciate it. I updated the code and tested using json_encode called with the following: "NULL", "null", null, "[]", "[\"teste\"]". Worked with all.

resources/views/bread/read.blade.php
```diff
@@ -56,15 +56,14 @@
                                 <img class="img-responsive"
                                      src="{{ filter_var($dataTypeContent->{$row->field}, FILTER_VALIDATE_URL) ? $dataTypeContent->{$row->field} : Voyager::image($dataTypeContent->{$row->field}) }}">
                             @elseif($row->type == 'multiple_images')
-                                @if(json_decode($dataTypeContent->{$row->field}))
-                                    @foreach(json_decode($dataTypeContent->{$row->field}) as $file)
-                                        <img class="img-responsive"
-                                             src="{{ filter_var($file, FILTER_VALIDATE_URL) ? $file : Voyager::image($file) }}">
-                                    @endforeach
-                                @else
+                                @php $image_files = json_decode($dataTypeContent->{$row->field}) ?? []; @endphp
```
MrCrayon: Not sure if @emptynick agrees but I would remove this temporary variable and just have:
```php
@forelse (json_decode($dataTypeContent->{$row->field}) ?? [] as $file)
```
in the next line

emptynick: Doesn't really matter but I agree

igordelorenzi: Wow! The elegance of this code is escalating fast, haha. Just updated.

PR #5049, opened 2020-08-20
Fix array handling

domainregistrar:


resources/views/bread/browse.blade.php
```diff
@@ -246,7 +246,8 @@
                                                     @endif
                                                 @else
                                                     @include('voyager::multilingual.input-hidden-bread-browse')
-                                                    <span>{{ $data->{$row->field} }}</span>
+                                                    {{--<span>{{ $data->{$row->field} }}</span>--}}
```
fletch3555, line 249: Please delete commented code.  There's no point in leaving it when everything is tracked by git

resources/views/bread/read.blade.php
```diff
@@ -127,7 +127,8 @@
                                 @endif
                             @else
                                 @include('voyager::multilingual.input-hidden-bread-read')
-                                <p>{{ $dataTypeContent->{$row->field} }}</p>
+                                {{-- <p>{{ $dataTypeContent->{$row->field} }}</p> --}}
```
fletch3555, line 129: Same here

fletch3555:
When opening pull requests, we ask that you provide justification for the change.  A link to an open issue or description of the problem would be enough.

terax6669:
Looks like the author of the PR doesn't intend to participate further. I ask that one of the maintainers clean this up and merge.
This is what happens currently when you use eloquent's cast to array:

```php
	protected $casts = [
		'attributes' => 'array',
	];
```

```
ErrorException
htmlspecialchars() expects parameter 1 to be string, array given
```

Obviously blade can't output an array via `{{ }}`.

PR #5056, opened 2020-08-29
Allow action buttons to be filtered by row

leith:
The filter by data type is sometimes just not enough if you want conditions based on the row data

### DIFF
--- a/resources/views/bread/partials/actions.blade.php
+++ b/resources/views/bread/partials/actions.blade.php
@@ -5,9 +5,11 @@
         $action = new $class($dataType, $data);
     @endphp
     @can ($action->getPolicy(), $data)
-        <a href="{{ $action->getRoute($dataType->name) }}" title="{{ $action->getTitle() }}" {!! $action->convertAttributesToHtml() !!}>
-            <i class="{{ $action->getIcon() }}"></i> <span class="hidden-xs hidden-sm">{{ $action->getTitle() }}</span>
-        </a>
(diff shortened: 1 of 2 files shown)


fletch3555:
This is already possible with the authorization check (`@can()`) just above what you've changed.

leith:
You are completely right, I just hadn't delved enough into creating policies. I'll close this and remove the branch.

It may be worth documenting this or providing an example - the only custom action in the project at the moment is the menu 'builder' button, but since that's is built into a custom view it doesn't go through the same BREAD pipeline where the default action buttons are rendered.

MrCrayon:
Policies are not specific to Voyager it's standard Laravel, but you can see a custom one used for Posts in Voyager codebase.
https://github.com/the-control-group/voyager/blob/1.4/src/Policies/PostPolicy.php

leith:
@MrCrayon absolutely, policies are a far broader-use thing in Laravel, I was more meaning the use of them when restricting custom Action buttons.

The example `PostPolicy` is used with the "standard" read/edit/delete rather than anything custom. Additional documentation is probably more useful than the example in this case - it's not especially clear that the `getPolicy()` method is for defining the *permission* (for the `@can` call) to look up on the Policy file, and the policy is determined by association on the DataType (so need to look at the `PostsSeeder` from the current example). The `shouldActionDisplayOnDataType()` documents how to filter new action buttons so they only appear for certain data types, which is good.

PR #5061, opened 2020-08-31
Allow scopes to be dynamic if parameters are defined by the controller

leith:
At the moment the BREAD controllers are only capable of using fixed "scope" filters that take no parameters.

This adds support for dynamic scopes through a straightforward function override convention. e.g.

```php
class Post extends Model {
    public function scopeForCategory($query, $category) {
        return $query->where('category', $category);
    }
}
```
Then in the BREAD setup, the 'forCategory' scope becomes selectable, but will fail on render due to too few parameters. Simple controller override solves this:
```php
class MyPostController extends VoyagerBaseController {
    public function scopeForCategoryParams(Request $request) {
        // grab the category ID dynamically - example is from the request, but could be from anywhere
        return [$request->route()->parameter('category')];
    }
}
```
This pull request is in particular for when you need to grab the parameter dynamically, although it would be also be possible to have a "fixed" parameter passed to a dynamic scope by adding an Optional Details JSON with a `scopeParams` array to the `dataType` configuration itself.

### DIFF
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -71,7 +71,12 @@ class VoyagerBaseController extends Controller
             $model = app($dataType->model_name);
 
             if ($dataType->scope && $dataType->scope != '' && method_exists($model, 'scope'.ucfirst($dataType->scope))) {
-                $query = $model->{$dataType->scope}();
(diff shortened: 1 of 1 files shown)


MrCrayon:
Thanks for PR, interesting idea. :+1: 
We are all quite busy so don't think we are ignoring PRs but it gets time to review when you did not write the code.
Adding tests when you can would help.

PR #5068, opened 2020-09-07
Initial Laravel 8 support

emptynick:
This PR adds Laravel 8 support.
Tests **will** fail until Laravel releases a stable version 8.
Tests **pass** when `minimum-stability` is set to `dev` in `composer.json`

Once 8.0.1 is released, re-run actions to verify all tests pass, merge this PR, release a new version and we are good!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
 
       # Install PHP dependencies
       - name: Install legacy factories for Laravel
-        if: matrix.laravel == "8.*"
+        if: ${{ matrix.laravel == "8.*" }}
         run: composer require laravel/legacy-factories
 
       # Display installed laravel version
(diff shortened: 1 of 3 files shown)


MrCrayon:
Once 8 is out we should also move `laravel/legacy-factories` from the action to `require-dev`.

emptynick:
Has been there, but it requires Laravel 8 and fails for L6 and 7.
See https://github.com/the-control-group/voyager/pull/5068/commits/7a676c00efa84c6f4f21659a8d3ba53a449a0bd0

MrCrayon:
That's unfortunate, maybe we can add:
```
"suggest": {
    "laravel/legacy-factories": "Required to run Voyager tests with Laravel 8.x"
}
```
Just to make contributors life easier

emptynick:
Sounds reasonable 👍

PR #5070, opened 2020-09-09
Fixes for Laravel 8

MrCrayon:
Laravel 8 moved default location of `User` from `App` to `App\Models`.

This PR enables checking both locations before giving up.

Fixed Paginator since Laravel 8 uses Tailwind as default styling.

### DIFF
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -77,13 +77,15 @@ class InstallCommand extends Command
         $this->call('migrate', ['--force' => $this->option('force')]);
 
         $this->info('Attempting to set Voyager User model as parent to App\User');
-        if (file_exists(app_path('User.php'))) {
-            $str = file_get_contents(app_path('User.php'));
+        if (file_exists(app_path('User.php')) || file_exists(app_path('Models/User.php'))) {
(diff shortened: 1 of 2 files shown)


src/Commands/InstallCommand.php
```diff
@@ -77,13 +77,15 @@ public function handle(Filesystem $filesystem)
...
+            $str = file_get_contents($userPath);
 
             if ($str !== false) {
                 $str = str_replace('extends Authenticatable', "extends \TCG\Voyager\Models\User", $str);
 
-                file_put_contents(app_path('User.php'), $str);
+                file_put_contents($userPath, $str);
             }
         } else {
             $this->warn('Unable to locate "app/User.php".  Did you move this file?');
```
emptynick: Let's as a `or app/Models/User` here.

emptynick (dismissed):

src/Commands/InstallCommand.php
MrCrayon: What do you mean? `$userPath` already contains the correct path

emptynick: I mean the warning shown if no model is found

MrCrayon: oh, got it! I'll fix it later

emptynick (approved):

PR #5072, opened 2020-09-09
Check user permissions before uploading file

MrCrayon:
VoyagerController@upload is used by tinyMCE to upload images.
This PR introduces checks to verify that user has permission to edit or add the BREAD and that there is a visibile tinyMCE editor in that BREAD action.


src/Http/Controllers/VoyagerController.php
```diff
@@ -101,4 +107,10 @@ public function assets(Request $request)
 
         return response('', 404);
     }
+
+    protected function userCannotUploadImageIn($dataType, $action)
+    {
+        return auth()->user()->cannot($action, app($dataType->model_name))
```
fletch3555, line 113: Since we're in a controller class, any reason not to use `$this->authorize(...)` instead?

https://laravel.com/docs/8.x/authorization#via-controller-helpers

MrCrayon: If I'm not mistaken authorize does not return a Boolean so we won't be able to check both cases.
I'll verify tomorrow

MrCrayon: Yes, authorize throws an AuthorizationException so we can't use it in this case.

emptynick (approved):

PR #5093, opened 2020-09-23
Fix Seeder for Laravel 8

MrCrayon:
### Problem
Laravel 8 moved default location for seeders from `database/seeds` to `database/seeders` also seeder classes now need to be namespaced.
These are the changes in new Laravel project `composer.json`:
```diff
    "autoload": {
        "psr-4": {
            "App\\": "app/",
+            "Database\\Factories\\": "database/factories/",
+            "Database\\Seeders\\": "database/seeders/"
+        }
-        },
-        "classmap": [
-            "database/seeds",
-            "database/factories"
-        ]
    },
```

### Solution
Check Laravel version to publish seeders in the correct folder and add namespace if needed

### Implementation
To implement the solution we need to change Install command to:
1. Check Laravel version `getSeedsFolderName`
2. Publish files in the correct folder
3. Change seeders to add namespace `addNamespaceIfNeeded`
4. Seed

#### Flow on Install Command
Previous to this PR on Install command seeders were first seeded and than published, that order has been changed.

#### Autoloading seeders on test enviroment
To be able to seed seeders while testing we need to be able to autoload seeders inside `orchestra/testbench-code/laravel/database/(seeds|seeders)/`.
Autoload does not normally work with testbench because there is no `tests/TestCase.php` file included with test app environment provided but it's referred to in composer.json:
```json
    "autoload": {
        "classmap": [
            "database",
            "tests/TestCase.php"
        ],
    },
```
The solution has been to create an empty file within `TestCase` `setup` method

#### Seeders need to be available during install
Calling
```php
$this->composer->dumpAutoloads();
```
 will create autoloader but will not load files, since we need seeders within the same process this has been added right after:
```php
require_once base_path('vendor/autoload.php');
```

#### Removed references to Seedable Trait
Seedable Trait was used to solve the problem of seeders not available because not autoloaded, that Trait is not _used_ anymore since autoloading allows us to use standard:
```php
$this->call([
    DataTypesTableSeeder::class,
]);
```
to call other seeders.

#### Changed how composer is called
Laravel provides a class to call composer so `Symfony\Component\Process\Process` has been removed in favor of `Illuminate\Support\Composer`

### DIFF
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -7,15 +7,10 @@ use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Composer;
 use Symfony\Component\Console\Input\InputOption;
 use TCG\Voyager\Providers\VoyagerDummyServiceProvider;
-use TCG\Voyager\Traits\Seedable;
 use TCG\Voyager\VoyagerServiceProvider;
 
 class InstallCommand extends Command
 {
-    use Seedable;
-
-    protected $seedersPath;
-
     /**
      * The console command name.
      *
@@ -51,8 +46,6 @@ class InstallCommand extends Command
         $this->composer = $composer;
 
         $this->seedFolder = version_compare(app()->version(), '8.0') >= 0 ? 'seeders' : 'seeds';
-
-        $this->seedersPath = database_path($this->seedFolder).'/';
     }
 
     protected function getOptions()
@@ -151,14 +144,14 @@ class InstallCommand extends Command
         $this->composer->dumpAutoloads();
 
         $this->info('Seeding data into the database');
-        $this->seed('VoyagerDatabaseSeeder');
+        $this->call('db:seed', ['--class' => 'VoyagerDatabaseSeeder']);
 
         if ($this->option('with-dummy')) {
             $this->info('Migrating dummy tables');
             $this->call('migrate');
 
             $this->info('Seeding dummy data');
-            $this->seed('VoyagerDummyDatabaseSeeder');
+            $this->call('db:seed', ['--class' => 'VoyagerDummyDatabaseSeeder']);
         }
 
         $this->info('Setting up the hooks');
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -129,18 +129,17 @@ class InstallCommand extends Command
             $this->call('vendor:publish', ['--provider' => VoyagerDummyServiceProvider::class, '--tag' => $tags]);
 
             $this->addNamespaceIfNeeded(
-                collect($filesystem->files("{$publishablePath}/database/seeds/"))
-                    ->concat($filesystem->files("{$publishablePath}/database/dummy_seeds/")),
+                collect($filesystem->files("{$publishablePath}/database/dummy_seeds/")),
                 $filesystem
             );
         } else {
             $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class, '--tag' => ['config', 'voyager_avatar']]);
-
-            $this->addNamespaceIfNeeded(
-                collect($filesystem->files("{$publishablePath}/database/seeds/")),
-                $filesystem
-            );
         }
+
+        $this->addNamespaceIfNeeded(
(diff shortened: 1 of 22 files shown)


src/Commands/InstallCommand.php
```diff
@@ -109,21 +121,40 @@ public function handle(Filesystem $filesystem)
...
+            $this->addNamespaceIfNeeded(
+                collect($filesystem->files("{$publishablePath}/database/seeds/"))
+                    ->concat($filesystem->files("{$publishablePath}/database/dummy_seeds/")),
+                $filesystem
+            );
+        } else {
+            $this->call('vendor:publish', ['--provider' => VoyagerServiceProvider::class, '--tag' => ['config', 'voyager_avatar']]);
+
+            $this->addNamespaceIfNeeded(
+                collect($filesystem->files("{$publishablePath}/database/seeds/")),
```
fletch3555: This is common to both paths, maybe pull it out of the conditional?

src/Commands/InstallCommand.php
```diff
@@ -134,4 +165,23 @@ public function handle(Filesystem $filesystem)
...
+    {
+        if ($this->seedFolder != 'seeders') {
+            return;
+        }
+
+        $seeds->each(function ($file) use ($filesystem) {
+            $path = database_path('seeders').'/'.$file->getFilename();
+
+            $stub = str_replace(
+                "<?php\n\nuse",
```
fletch3555, line 178: This breaks down quickly if someone ever adds a `declare` statement (i.e. for strict_types) or changes the amount of whitespace, etc... Maybe change it to (pseudocode): "if !file_contains(namespace) { add_namespace }"

src/Voyager.php
```diff
@@ -353,4 +353,9 @@ public function getLocales()
 
         return $allLocales;
     }
+
+    public function getSeedsFolderName()
```
fletch3555: Not too sure I like having this exposed through the Voyager facade.  Doesn't feel like the right place (though I don't really have a better solution at the moment).  Facade just feels like it should be more of an external API, whereas stuff like this is 100% internal (and honestly not even Voyager-specific)

fletch3555 (dismissed):

src/Commands/InstallCommand.php
MrCrayon: Yes I could create a variable to keep the list.

src/Commands/InstallCommand.php
MrCrayon: But we are using this on our own seeders only.

src/Commands/InstallCommand.php
fletch3555: Eh, just call the method twice.  Once in the dummy condition, and once below for everything else

src/Commands/InstallCommand.php
fletch3555: Yup, but it's something easily missed in PR reviews.  Anyone could add a new seeder as part of a feature change/addition, and we currently have no controls around code style for them (as far as I know)

src/Voyager.php
MrCrayon: Yes, I agree. That was the easier place but probably not the best suited.
I could create a new class with an handle method or put it in an helper :confused: 
What do you think?

fletch3555: Helper functions are also exposed globally, so they pollute the app.  Not really a fan personally.  Let's go with a "helper class".  A class with static methods that can be used within Voyager, but don't necessarily need to be documented anywhere

src/Commands/InstallCommand.php
MrCrayon: I'm not against making it better but I can't think of a quick way because you'll still need to understand where to insert the namespace.
In the other hand I'm not really sure we need to change it, if for any reason adding the namespace is not going to work when adding some seeder whoever is coding the seeder should already notice that.

src/Commands/InstallCommand.php
MrCrayon: Done

src/Voyager.php
MrCrayon: Done, not sure about class name and position but I could not put it in Helpers and not in Actions.

emptynick (approved):

Issue #4758, opened 2020-02-12
empty "media picker" with "base_path" option, File not found at path


levantez:
Sorry for my english...

## Version information

 - Laravel: v6.2
 - Voyager: v1.3.0
 - PHP: 7.3
 - Database: MySQL 5.6

## Description

I have a table "news", and i have field "gallery" (text, default null), in BREAD this field have type "Media Picker" with option "base_path":"/news/{pk}-gallery/", if i add news and leave this field empty, i take error like this: 
"League\Flysystem\FileNotFoundException
File not found at path: news/f5d692d1-9d3a-461a-9d60-578d0407a4e7 "
This error happens in ''/vendor/tcg/voyager/src/Http/Controllers/Controller.php" on line 154 `Storage::disk(config('voyager.storage.disk'))->move($old_path, $new_path);`, controller try to rename temp directory to directory from "base_patch", but temp directory was not created , because the field is empty.

If add some pictures to gallery - its work ok.
If remove "base_path":"/news/{pk}-gallery/" - its work ok without pictures.

I fix this problem by checking exists temp directory, change line 152
`if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path))`
to 
`if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path) && Storage::disk(config('voyager.storage.disk'))->exists($old_path))`

fletch3555:
I would entertain this as a Pull Request if you wouldn't mind opening one.

levantez:
> 
> 
> I would entertain this as a Pull Request if you wouldn't mind opening one.

Yes. Do it Yourself, please

msrivela:
Hey @fletch3555, just sent a [pull request](https://github.com/the-control-group/voyager/pull/5099) to fix this issue.

PR #5099, opened 2020-09-28
Fix for empty "media picker" with "base_path" option, File not found …

msrivela:
Just ran across an error described on issue #4758 and applied the proposed fix.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -168,7 +168,7 @@ abstract class Controller extends BaseController
                 $data->{$row->field} = str_replace($uuid, $data->getKey(), $data->{$row->field});
             });
             $data->save();
-            if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path)) {
+            if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path) && Storage::disk(config('voyager.storage.disk'))->exists($old_path)) {
                 $request->session()->forget([$slug.'_path', $slug.'_uuid']);
                 Storage::disk(config('voyager.storage.disk'))->move($old_path, $new_path);
                 Storage::disk(config('voyager.storage.disk'))->deleteDirectory($folder_path);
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/Controller.php
```diff
@@ -168,7 +168,7 @@ public function insertUpdateData($request, $slug, $rows, $data)
                 $data->{$row->field} = str_replace($uuid, $data->getKey(), $data->{$row->field});
             });
             $data->save();
-            if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path)) {
+            if ($old_path != $new_path && !Storage::disk(config('voyager.storage.disk'))->exists($new_path) && Storage::disk(config('voyager.storage.disk'))->exists($old_path)) {
```
fletch3555:
```suggestion
            $disk = Storage::disk(config('voyager.storage.disk'));
            if ($old_path != $new_path && !$disk->exists($new_path) && $disk->exists($old_path)) {
```

Just a readability change.  ~~Can you also update the other usages of `Storage::disk(...)`?~~ (Edit: nvm, didn't realize I could batch suggestions)

src/Http/Controllers/Controller.php
```diff
@@ -168,7 +168,7 @@ public function insertUpdateData($request, $slug, $rows, $data)
...
                 $request->session()->forget([$slug.'_path', $slug.'_uuid']);
                 Storage::disk(config('voyager.storage.disk'))->move($old_path, $new_path);
```
fletch3555:
```suggestion
                $disk->move($old_path, $new_path);
```

src/Http/Controllers/Controller.php
```diff
@@ -168,7 +168,7 @@ public function insertUpdateData($request, $slug, $rows, $data)
...
                 Storage::disk(config('voyager.storage.disk'))->deleteDirectory($folder_path);
```
fletch3555:
```suggestion
                $disk->deleteDirectory($folder_path);
```

PR #5120, opened 2020-10-22
Feature enable ordering and searching by belongsTo relationship column

### LINKED ISSUE #2661: Sort by relationship column

### LINKED ISSUE #3595: BelongsTo ServerSide Search
## Version information

 - Laravel: v5.4
 - Voyager: v1.1.3
 - PHP: 7.1
 - Database: MySQL 5.6.4

## Description
Can't search BelongTo items on Browse view in Server-Side Search

## Steps To Reproduce
1. Go to BREAD
2. Create a new BelongsTo relationship.
3. Select Server-Side pagination
4. Go to the model Browse View
5. Try to search content based on the Related table.

## Expected behavior
What is needed here, is to customize the seach-key dropdown filter, so it is possible to search based on the related model data.

## Screenshots
ss

MrCrayon:
Tested both soft delete and scope and they are still working.

Closes #2661 and closes #3595

### DIFF
--- a/resources/views/bread/browse.blade.php
+++ b/resources/views/bread/browse.blade.php
@@ -85,7 +85,7 @@
                                         @endif
                                         @foreach($dataType->browseRows as $row)
                                         <th>
-                                            @if ($isServerSide && $row->type !== 'relationship')
+                                            @if ($isServerSide && ($row->type !== 'relationship' || $row->details->type == 'belongsTo'))
                                                 <a href="{{ $row->sortByUrl($orderBy, $sortOrder) }}">
                                             @endif
                                             {{ $row->getTranslatedAttribute('display_name') }}
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -95,10 +95,23 @@ class VoyagerBaseController extends Controller
                 $query->where($search->key, $search_filter, $search_value);
             }
 
-            if ($orderBy && in_array($orderBy, $dataType->fields())) {
+            $row = $dataType->rows->where('field', $orderBy)->firstWhere('type', 'relationship');
+            if ($orderBy && (in_array($orderBy, $dataType->fields()) || !empty($row))) {
                 $querySortOrder = (!empty($sortOrder)) ? $sortOrder : 'desc';
+                if (!empty($row)) {
+                    $query->select([
+                        $dataType->name.'.*',
+                        $row->details->table.'.'.$row->details->label.' as relationshipOrderField',
+                    ])->leftJoin(
+                        $row->details->table,
+                        $dataType->name.'.'.$row->details->column,
+                        $row->details->table.'.'.$row->details->key,
+                    );
+                    $orderByRelationship = 'relationshipOrderField';
+                }
+
                 $dataTypeContent = call_user_func([
-                    $query->orderBy($orderBy, $querySortOrder),
+                    $query->orderBy($orderByRelationship ?? $orderBy, $querySortOrder),
                     $getter,
                 ]);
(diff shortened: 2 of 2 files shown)


Stehos:
@MrCrayon  perfectly working with different names of tables. The problem is at the moment when we have join to the same table. E.g. you have categories where each category has its parent_id from the same table. You will receive the following error:
`
SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'categories' (SQL: select count(*) as aggregate from `categories` left join `categories` on `categories`.`parent_id` = `categories`.`id`) `

MrCrayon:
@Stehos thanks for checking, using an alias for the joined table should solve the problem. I updated PR.

Stehos (dismissed):
@MrCrayon  Thank you. Tested, it's working properly.

Stehos:
@MrCrayon Note: right now I also checked the search functionality and it's not working with those changes. The following error will be received:
```
Illuminate\Database\QueryException
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'parent_id' in where clause is ambiguous (SQL: select count(*) as aggregate from `categories` left join `categories` as `joined` on `categories`.`parent_id` = `joined`.`id` where `parent_id` LIKE %test%) 
```

Anyway, this search functionality was not working properly in any of the relational columns, neither before changes. It was working only for static columns.

MrCrayon:
@Stehos yes it was not working already but I should not break it even more, you must have guessed I don't use the standard server side pagination very often :grin: 

Prefixing search field with table name should probably fix the problem :thinking:

MrCrayon:
@Stehos prefixing with table worked but since I was there I tried to also fix search, it worked with my tests.
Please if you have time give it a try.

Stehos:
@MrCrayon you are great. Thank you for your investigation. I just made some tests and the search is working properly with different names of tables. Or better saying, it's also working for the join from the same table - you receive correct rows but with the same value in both columns. E.g. you have Title and Parent columns and when you are going to search via Parent field, you get the same values in both of those two columns.

obrázok


Thank you.

MrCrayon:
@Stehos thanks again for checking, I see why nobody did this before it's quite tricky.
I'll have a look at it tomorrow.

MrCrayon:
I didn't have time to check today but I think to make search work we will need to manage pagination manually, maybe I'll just fix ordering here and keep the searching WIP in another branch.

MrCrayon:
Superseded by #5185

@Stehos Happy New Year! :champagne: 

I'm closing this PR in favor of #5185 it would be really useful if you could have a look at that one, both ordering and search should work :crossed_fingers:

Stehos:
@MrCrayon Happy New Year and a lot of health also to you.

I have checked your PR on my local environment and sorting was without problem. Searching in via title was also fine but when I tried to search by parent I didn't get anything - just a blank white page.

PR #5129, opened 2020-10-30
Change Request::url to Request::path

b4oshany:
The URL for the site was showing up in the dashboard. Request::path removes the URL

image

### DIFF
--- a/resources/views/dashboard/navbar.blade.php
+++ b/resources/views/dashboard/navbar.blade.php
@@ -7,7 +7,7 @@
             @section('breadcrumbs')
             <ol class="breadcrumb hidden-xs">
                 @php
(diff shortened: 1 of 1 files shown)


b4oshany:
I think I spot the problem from my side... This change isn't needed.

b4oshany:
Just for reference, this is what is happening:
- `route('voyager.dashboard')` return a `https` URL.
- `Request::url()` returns a `http` URL

I actually thought I forced HTTPs for all paths, but I guess not.

I'm using a proxy server, which resulted in error.

b4oshany:
Problem resolved.... I had to set the trusted proxy servers... See https://laravel.com/docs/8.x/requests#configuring-trusted-proxies

However, I do think it should switch over to `Rquest::path` along with the path of the route

PR #5153, opened 2020-12-03
Breadcumbs segments by path

dlopezp:
When the app is behind a reverse proxy that handle SSL connection, APP_URL is secure but the current Url in the app is non https.
This trigger in a bad formed breadcumbs


MrCrayon:
Configuring Trusted Proxies doesn't fix it?
https://laravel.com/docs/8.x/requests#configuring-trusted-proxies

dlopezp:
True, sorry for not check it.

PR #5186, opened 2021-01-06
Add support for custom pagination view

ohmygodvt95:
### Add support for custom pagination

By default laravel 8.x uses tailwind for pagination. So the pagination will look like below:

Screen Shot 2021-01-06 at 20 12 10
Of course it is possible to use the instructions below to change the view of pagination but it will take away the flexibility of this option. So I added an option to allow users to change the view of pagination.

Refs:
- 8.x uses tailwind: https://laravel.com/docs/8.x/pagination#displaying-pagination-results
- Using Bootstrap: https://laravel.com/docs/8.x/pagination#using-bootstrap

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -257,4 +257,26 @@ return [
            ],
        ]*/
     ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Pagination specific settings
+    |--------------------------------------------------------------------------
+    |
+    | Here you change pagination view setting
+    |
+    | Currently Laravel support:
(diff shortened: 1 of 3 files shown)


MrCrayon:
Thanks for PR but a fix for pagination with Laravel 8 has been already merged with #5070.
I don't think having an option for tailwindcss pagination is going to be useful since all templates already use bootstrap.
Just in case anyway I think configuration can be overriden in your App Service provider.

ohmygodvt95:
@MrCrayon thank you for u comment. I will close PR

PR #5209, opened 2021-02-02
Fix Tools/Database cannot be created when set voyager path to "/" (#5…

ahmadsyamim:
…208)


MrCrayon:
Thanks for PR but as explained in related issue the fact that is posting on /create means you have some other problem probably javascript related.

PR #5228, opened 2021-02-11
Add Support for default value in BREAD relationships

### LINKED ISSUE #5227: Add Support for default value in BREAD relationships
## Description of problem
https://stackoverflow.com/questions/55863868/how-to-force-a-default-value-on-a-referenced-field-in-voyager
I would like to set a default value for a relationship field.

## Proposed solution
Add support for "default" option same as regular BREAD fields.

## Alternatives considered
-

## Additional context
#5228

terax6669:
Closes #5227

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build
 .lock
 composer.lock
 .phpunit.result.cache
+nbproject
--- a/resources/views/formfields/relationship.blade.php
+++ b/resources/views/formfields/relationship.blade.php
@@ -27,12 +27,18 @@
                     data-get-items-route="{{route('voyager.' . $dataType->slug.'.relation')}}"
                     data-get-items-field="{{$row->field}}"
                     @if(!is_null($dataTypeContent->getKey())) data-id="{{$dataTypeContent->getKey()}}" @endif
-                    data-method="{{ !is_null($dataTypeContent->getKey()) ? 'edit' : 'add' }}"
+                    data-method="{{ $dataTypeContent->exists ? 'edit' : 'add' }}"
                     @if($row->required == 1) required @endif
                 >
                     @php
-                        $model = app($options->model);
-                        $query = $model::where($options->key, old($options->column, $dataTypeContent->{$options->column}))->get();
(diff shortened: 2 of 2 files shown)


MrCrayon:
Thanks for you PR!
Can you please remove everything unrelated like changing variable names and formatting?

terax6669:
Why? I see nothing but improvements there. I've allowed maintainers to make edits, so you can make any adjustments you see fit.

MrCrayon:
Because it's overhead for whoever needs to check your PR to have 68 lines changed instead of 2 or 3 needed, it also makes harder to use git blame in the future.

terax6669:
So for the maintainers' convenience? You can just glance at most of these lines and you can go as deep in the git history as you need. That code should have never made it's way into the repository in the first place.

Indention not adhering to rules... come on. Those are the basics.
Why make a variable and immediately overwrite it? It's confusing. Wanna argue about wasted time? I've wasted time trying to understand what `$selected_value` was because it's value and meaning was changing.
The whole project uses camelCase for the variables, that one odd `$selected_value` shouldn't be there.

You could make the argument that changing
```php
->get()->map(function ($item, $key) use ($options) {
                                return $item->{$options->key};
                            })->all()
```
to `->pluck($options->key)->all()` or `!is_null($dataTypeContent->getKey())` to `$dataTypeContent->exists` wasn't necessary. Still, this way is faster (two function calls and a negation vs simply reading a boolean), shorter, easier to read and understand. Also the old way was guessing that the model doesn't exist by reading the value of the primary key. What if the key was composite giving you `[null, null]`? Or maybe someone has a UUID trait that assigns a random id in the trait's boot method? The exact purpose of the Eloquent\Model's `exists` property is to determine if the model exists in the database or not i.e. if we're adding or editing.

fletch3555:
@terax6669 yes for the maintainers convenience.  There's a reason it's called a pull REQUEST.  We're happy to work with the community to get changes merged in, but you have to understand that we're often quite busy and don't have time to fix code style or whatever for each PR we receive.  Please be cognizant of the tone and how your comments read, and we won't have a problem.

I happen to agree that @MrCrayon 's request is fair.  Please make the appropriate changes and we'll be happy to review this again.

PR #5269, opened 2021-03-22
keep api with clean, keep voyager querys only to web route

eodicarvallho:
Hi guys, this is just a small fix that i hope can help.
In api calls, we lose at least 200 to 700ms with voyager calls, so i made a check to verify if the request is from api or not, if so, then it hides the voyager provider, else, keep everything as it was.
check this print to illustrate:
Voyager

if this is not helpfull , i hope we can find a better way.

### DIFF
--- a/src/VoyagerServiceProvider.php
+++ b/src/VoyagerServiceProvider.php
@@ -32,339 +32,344 @@ use TCG\Voyager\Providers\VoyagerEventServiceProvider;
 use TCG\Voyager\Seed;
 use TCG\Voyager\Translator\Collection as TranslatorCollection;
 
-class VoyagerServiceProvider extends ServiceProvider
-{
-    /**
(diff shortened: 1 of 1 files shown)


shahruslan:
The idea is good, but the implementation is bad
I did the following on my project:
1. Disable auto-discovery in `composer.json`
```json
"extra": {
        "laravel": {
            "dont-discover": [
                "tcg/voyager"
            ]
        }
    },
```
2. In `AppServiceProvider`, I configured the conditional registration of the voyager provider:
```php
if ($this->app->runningInConsole() || request()->is('admin', 'admin/*')) {
    $this->app->register(VoyagerServiceProvider::class);
}
```

The same can be done inside `VoyagerServiceProvider`, only the problem is that the address (/ admin /) is not known in advance. It is set in `web.php`.

PR #5280, opened 2021-04-01
Make installing on Windows work

emptynick:
As Windows uses `\r\n` as new-line, renaming the seeders does not work and thus, installing fails when using Laravel 8 on Windows.


MrCrayon:
Maybe we should change this one too, it's probably not a problem but it should be better.

https://github.com/the-control-group/voyager/blob/0d509a021fbc95654c699c74b6c9d98e1b1c2bdb/src/Commands/InstallCommand.php#L120

emptynick:
This was never a problem for me on windows, but done.

PR #5281, opened 2021-04-01
Only search browsable fields

emptynick:
When using server-side pagination, all fields are shown in the search dropdown and are searchable, even when they are unticked in BREAD.
This PR prevents this behaviour as a non-browsable field should not be searchable.


howdu:
Relation method also allows open searching.

https://github.com/the-control-group/voyager/blob/00f2c540e568cfb69729c16a0797a8babebf0a5d/src/Http/Controllers/VoyagerBaseController.php#L876

Issue #5294, opened 2021-04-18
Key name is too long


rickmacgillis:
## Version information

 - Laravel: v8.12.0
 - Voyager: v1.4.2
 - PHP: 7.4
 - Database:MySQL (MariaDB) 10.5.9

## Description

After requiring Voyager, I ran the installation command below, and it gave me the following error during the migration run.
```
SQLSTATE[42000]: Syntax error or access violation: 1059 Identifier name 'kjvh_translations_table_name_column_name_foreign_key_locale_unique' is too long (SQL: alter table `kjvh_translations` add unique `kjvh_translations_table_name_column_name_foreign_key_locale_unique`(`table_name`, `column_name`, `foreign_key`, `locale`))
```

## Steps To Reproduce
Steps to reproduce the behavior: follow steps 1-3 at https://github.com/the-control-group/voyager#readme

## Expected behavior
A finished installation

## Screenshots
See: https://www.loom.com/share/6c462f6a80e14166a1f021c15f957cee

## EDIT
Solution? https://www.loom.com/share/1b05834acb31444daf1a8c2237f2885f

## EDIT 2
On second look, it seems to be making a composite key making it unique across ALL columns, not making sure that each column has unique values. So, my approach in the video doesn't work, plus the names need to be column names, so shortening them won't work.

The simplest solution would be to change the column names throughout the codebase so that we can fit the key name into the DB. This could have consequences for existing users. I'll leave it up to the dev team on how to handle this. For now, I'll try things on my end to make it go through.

akiyamaSM:
Hello, @rickmacgillis I guess a better solution would be to publish the migration and change in it localy. then migrate.

MuhammadSadeeq:
I solved it by setting engine to 'InnoDB' in database.php file in config folder.
![carbon (1)](https://user-images.githubusercontent.com/54463147/115530913-a539c280-a2ad-11eb-9429-6089653e916a.png)

rickmacgillis:
@MuhammadSadeeq - That's fine, but keep in mind that Innodb and MyISAM both have their benefits and drawbacks. You should choose the best engine for the job at hand instead of choosing one that fits the key name. I haven't looked into it Innodb would be the best option here or not.

For anyone who would like to read a short post on some of the differences of Innodb and MyISAM, see Spencer's post at https://stackoverflow.com/a/12614698

MuhammadSadeeq:
[...]

Issue #3751, opened 2018-11-15
Identifier name 'laravel_translations_table_name_column_name_foreign_key_locale_unique' is too long


lbottoni:
## Version information

 - Laravel: v5.7.*
 - Voyager: v^1.1
 - PHP: 7
 - Database: mysql  Ver 15.1 Distrib 10.1.30-MariaDB, for Win32 (AMD64)

## Description
php artisan voyager:install --with-dummy

Publishing the Voyager assets, database, and config files
Copied Directory [\vendor\tcg\voyager\publishable\assets] To [\public\vendor\tcg\voyager\assets]
Copied Directory [\vendor\tcg\voyager\publishable\database\seeds] To [\database\seeds]
Publishing complete.
Copied File [\vendor\intervention\image\src\config\config.php] To [\config\image.php]
Publishing complete.
Migrating the database tables into your application
Migrating: 2016_01_01_000000_add_voyager_user_fields
Migrated:  2016_01_01_000000_add_voyager_user_fields
Migrating: 2016_01_01_000000_create_data_types_table
Migrated:  2016_01_01_000000_create_data_types_table
Migrating: 2016_05_19_173453_create_menu_table
Migrated:  2016_05_19_173453_create_menu_table
Migrating: 2016_10_21_190000_create_roles_table
Migrated:  2016_10_21_190000_create_roles_table
Migrating: 2016_10_21_190000_create_settings_table
Migrated:  2016_10_21_190000_create_settings_table
Migrating: 2016_11_30_135954_create_permission_table
Migrated:  2016_11_30_135954_create_permission_table
Migrating: 2016_11_30_141208_create_permission_role_table
Migrated:  2016_11_30_141208_create_permission_role_table
Migrating: 2016_12_26_201236_data_types__add__server_side
Migrated:  2016_12_26_201236_data_types__add__server_side
Migrating: 2017_01_13_000000_add_route_to_menu_items_table
Migrated:  2017_01_13_000000_add_route_to_menu_items_table
Migrating: 2017_01_14_005015_create_translations_table

   Illuminate\Database\QueryException  : SQLSTATE[42000]: Syntax error or access violation: 1059 Identifier name 'laravel_translations_table_name_column_name_foreign_key_locale_unique' is too long (SQL: alter table `laravel_translations` add unique `laravel_translations_table_name_column_name_foreign_key_locale_unique`(`table_name`, `column_name`, `foreign_key`, `locale`))

[...]

PR #5295, opened 2021-04-18
Prevent "index too long" error

rickmacgillis:
This PR fixes the error from issue #5294.

### DIFF
--- a/migrations/2017_01_14_005015_create_translations_table.php
+++ b/migrations/2017_01_14_005015_create_translations_table.php
@@ -22,8 +22,11 @@ class CreateTranslationsTable extends Migration
             $table->string('locale');
 
             $table->text('value');
-
-            $table->unique(['table_name', 'column_name', 'foreign_key', 'locale']);
+            
+            $table->unique(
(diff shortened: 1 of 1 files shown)


fletch3555:
@rickmacgillis, As you've undoubtedly discovered at this point, your prefix is the cause of the error you're getting.  Obviously this problem is not unique to you, but I do have a couple concerns with your solution.  I agree that hard-coding a name for that index is likely the correct solution, but we also have to weigh that against the complaints we'll get from users about how it's not using their configured prefix.

Perhaps there's a way to use the configured prefix and a simplified index name to keep it under the 64 character limit.  Obviously, a 63 character prefix would still break, but at least that would be consistent across the entire laravel app, not just this one migration.

@emptynick / @MrCrayon, any thoughts?

pegelf:
I think that would be the right solution:
https://github.com/the-control-group/voyager/issues/3751#issuecomment-555580542

Setting the unique-option per column is the same that was intended by bulk-setting this and it doesn't have a problem with the character limit.

PR #5296, opened 2021-04-20
Update menu controller to extend base controller for bread

howdu:
Menu is bread and should extend the base controller. This only become an issue if you extend menu controller you get missing index function. 

`class CustomMenuController extends \TCG\Voyager\Http\Controllers\VoyagerMenuController`

### DIFF
--- a/src/Http/Controllers/VoyagerMenuController.php
+++ b/src/Http/Controllers/VoyagerMenuController.php
@@ -6,7 +6,7 @@ use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
 use TCG\Voyager\Facades\Voyager;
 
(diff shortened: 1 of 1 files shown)


fletch3555:
I don't understand what problem you're trying to solve here... the parent class for the provided MenuController has no bearing on your ability to extend the MenuController for your own purposes...

howdu:
Yup it throws error for me when I try to extend the menu controller: Method TCG\Voyager\Http\Controllers\VoyagerMenuController::index does not exist.

Even without extending:
UPDATE `data_types` SET `controller` = 'TCG\\Voyager\\Http\\Controllers\\VoyagerMenuController' WHERE `slug` = 'menus';

PR #5302, opened 2021-04-29
Update voyager.php

mepsd:
in laravel 8 default namespace is App\Model\ so it is better uncomment in default and show the correct route while creating bread to table

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -42,7 +42,7 @@ return [
     */
 
     'models' => [
-        //'namespace' => 'App\\',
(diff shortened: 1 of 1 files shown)


fletch3555:
It's commented out so it won't affect anything.  If you want to change it, I don't have any concerns, but please leave it commented in the PR

mepsd:
Okay Thanks, i just saw this relative path error while creating bread so i think to update that.

emptynick (approved):

emptynick (dismissed):
Please also change voyager_dummy.php

emptynick (approved):

PR #5312, opened 2021-05-08
Exclude relationships from server-side search

### LINKED ISSUE #5412: Admin serverside search resulting in Unknown column 'posts.post_belongsto_category_relationship'
### Laravel version

8.52.0

### PHP version

7.4.9

### Voyager version

 1.x-dev aaee6ab

### Database

MySQL 8.0.22

### Description

While attempting to filter blogs based on category name using the supplied UI I get an error stating:

_Column not found: 1054 Unknown column 'posts.post_belongsto_category_relationship' in 'where clause' (SQL: select count(*) as aggregate from `posts` where `posts`.`post_belongsto_category_relationship` LIKE %cat%)_

### Steps to reproduce

1. Using the Dummy Data change Posts to server-side and Categories to Browse
2. Go to Posts on Admin dashboard
3.  Use filter drop down to select Categories, then contains, then use any search string (I used cat in my example)
4. Error is reproduced stating an unknown column. 

### Expected behavior

Posts should filter based on the search string within the input across the belongsto relationship. 

### Screenshots

_No response_

### Additional context

_No response_

emptynick:
This excludes all relationships from server-side search.
BelongsTo relationships still work as expected as the related field exists as a non-relationship.

As a POC: Change the user BREAD to server-side and search by `Role`.

Closes #5412


jegardiner:
I see this is a more succinct version of the change I suggest on issue #5412 .
Looking forward to this change passing review and being officially merged.

asefsoft:
please add this necessary feature

vitormicillo:
tested and working. 
Just a not: to work properly, it is necessary to rename the referral id field to the same visible field name, for example:

- company_id (the key field for datarow)
- Company Name (visible field name)

For the datarow reference is necessary, to rename company_id to Company Name as well.

vitormicillo (approved):
Note: to work properly, it is necessary to rename the referral id field to the same visible field name, for example:
- company_id (the key field)
- Company Name (visible field name)
for the datarow reference is necessary, to rename company_id to Company Name as well.

PR #5333, opened 2021-05-21
Fix error syntax error, unexpected ')'

marcelosantos89:
Some of the computers in our company were getting errors because of the comma on the end of line 112 giving syntax error, unexpected ')'.
This fixed it for our PCs, maybe it's just windows PHP that gets this error.


emptynick:
It's actually just PHP lower than PHP 7.3 which we don't support anymore (and it's end-of-life).
Nonetheless that comma is completely unnecessary.

Thanks!

emptynick (approved):

PR #5336, opened 2021-05-24
Update Controller.php

vitormicillo:
This will correct the error generated when saving a new record with media picker type fields.
The error will be thrown when an item is added but no file is being uploaded


emptynick:
A little more information:
When an item is added, a UUID for a folder is generated where files will be uploaded to.
After storing this folder will be renamed to the primary-key.
When no image is uploaded, the folder won't be generated and thus will fail.

emptynick (approved):

PR #5350, opened 2021-06-04
Update voyager.php

vitormicillo:
The user is able to define the number of widgets to display per line
Related to file tcg/src/Voyager.php

### DIFF
--- a/publishable/config/voyager_dummy.php
+++ b/publishable/config/voyager_dummy.php
@@ -148,6 +148,15 @@ return [
             'TCG\\Voyager\\Widgets\\PostDimmer',
             'TCG\\Voyager\\Widgets\\PageDimmer',
         ],
+        
+        // The admin is able to define the number of widgets to display per line
+        // Related to file tcg/src/Voyager.php
+        'widgets_per_line' => 4,        
+
+        // Defines the number of records to display when server-side paging is disabled.
+        'data_tables' => [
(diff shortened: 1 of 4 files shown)


publishable/config/voyager_dummy.php
```diff
@@ -148,6 +148,10 @@
             'TCG\\Voyager\\Widgets\\PostDimmer',
             'TCG\\Voyager\\Widgets\\PageDimmer',
         ],
+        
+        // The admin is able to define the number of widgets to display per line
+        // Related to file tcg/src/Voyager.php
+        'widgets_per_line' => 4,
```
MrCrayon, line 154: The default value should stay 3

resources/views/dimmers.blade.php
```diff
@@ -9,7 +9,7 @@
     $classes = [
         'col-xs-12',
         'col-sm-'.($count >= 2 ? '6' : '12'),
-        'col-md-'.($count >= 3 ? '4' : ($count >= 2 ? '6' : '12')),
+        'col-md-'.($count >= 3 ? '3' : ($count >= 2 ? '6' : '12')),
```
MrCrayon, line 12: This should be calculated base on how many widgets you want to show per row and the default should stay the same as before.

MrCrayon (changes requested):
Tests are failing but reverting default value should fix that.

PR #5368, opened 2021-06-12
Fix link to website

### LINKED ISSUE #5365: Broken link in documentation
The 'Home' link in the docs direct you to 'laravelvoyager.com' which is a wrong url

MrCrayon:
Not sure if we want to link that site or docs


emptynick:
Just edited to not fix the issue, as there's another link in the documentations header which we have to fix through gitbook itself

Issue #5353, opened 2021-06-05
[feature] multiple scopes for a datatype


Cyrille37:
## Version information

 - Laravel: v8.x
 - Voyager: v1.x|2.x
 - PHP: 7.3
 - Database: MariaDb

## Description of problem

The scope DataType (*`$dataType->scope`*) option is a single value, it would be nice if it is an Array (or whatever struct) that we can assign to a BREAD none, one or several local scopes. Like a scopePublished and a scopeTeam and so on.

## Proposed solution

None, sorry :-(
But thanks & cheers :-)

emptynick:
While I see what you mean I am not sure right now if you can chain scopes.
For now, I think the easiest is to make one scope returning everything from multiple scopes.

Edit: seems like you can chain scopes.
We are open for PRs.
Should be pretty simple.

Cyrille37:
> Edit: seems like you can chain scopes.

Yes, it's for me an important concept in Laravel, as you can create a simili query language like `$posts->published()->inTeamAbc()...`

I will think to a PR and come back :-) But I remember to have read something about Voyager v1.x will not have any new feature for the benefit of v2 that is in work in progress. Do you confirm ?

Thanks & Cheers.

emptynick:
That's correct. But this is a small change which we'd be happy to merge.

Cyrille37:
Related PR [Allow scopes to be dynamic if parameters are defined by the controller #5061](#5061).

Cyrille37:
I'm writing the PR, work in progress. Just a step to test if I'm on the right way ...

First step about the multi-selectbox for scope at BREAD edition : 

![image](https://user-images.githubusercontent.com/368965/122026409-79791b80-cdca-11eb-92dd-40854ed13778.png)

The VoyagerBreadController::getModelScopes() will be : 
```php
    public function getModelScopes($model_name)
    {
        $reflection = new ReflectionClass($model_name);

        return collect($reflection->getMethods())->filter(function ($method)
        {
            // method name start with 'scope'
            if( Str::startsWith($method->name, 'scope') )
            {
                // and have only one Required parameters (the $query)
                if( $method->getNumberOfRequiredParameters() == 1 )
                    return true ;
            }
            return false ;
        })->whereNotIn('name', ['scopeWithTranslations', 'scopeWithTranslation', 'scopeWhereTranslation'])->transform(function ($method) {
            return lcfirst(Str::replaceFirst('scope', '', $method->name));
        });
    }
```
[...]

PR #5373, opened 2021-06-15
1.4 bread multiple scopes

Cyrille37:
Here is the PR for [[feature] multiple scopes for a datatype #5353](5353).
It permits to define and apply several query scopes to BREAD model.

### DIFF
--- a/resources/views/tools/bread/edit-add.blade.php
+++ b/resources/views/tools/bread/edit-add.blade.php
@@ -260,12 +260,8 @@
                                             <option></option>
                                             @foreach($scopes as $scope)
                                             <option value="{{ $scope }}"
-                                                @if(is_array($dataType->scope))
-                                                    @if(in_array($scope,$dataType->scope)) selected @endif
-                                                @else {{-- for backward compat --}}
-                                                    @if($dataType->scope == $scope) selected @endif
-                                                @endif
-                                            >{{ $scope }}</option>
+                                                @if(in_array($scope, Arr::wrap($dataType->scope))) selected @endif
+                                                >{{ $scope }}</option>
                                             @endforeach
                                         </select>
                                     </div>
--- a/resources/views/tools/bread/edit-add.blade.php
+++ b/resources/views/tools/bread/edit-add.blade.php
@@ -252,8 +252,12 @@
                                 @if (isset($scopes) && isset($dataType))
                                     <div class="col-md-3 form-group">
                                         <label for="scopes">{{ __('voyager::bread.scope') }}</label>
-                                        <select id="scopes" name="scope[]" class="select2 form-control" multiple>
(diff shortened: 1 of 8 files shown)


resources/views/tools/bread/edit-add.blade.php
```diff
@@ -251,12 +251,20 @@ class="toggleswitch"
...
+                                        <label for="scopes">{{ __('voyager::bread.scope') }}</label>
+                                        <!-- an mandatory hidden field to give data to request when the select is empty -->
+                                        <input type="hidden" name="scope[]" value="" />
+                                        <select id="scopes" name="scope[]" class="select2 form-control" multiple
+                                            data-placeholder="Optional scope(s)"
+                                            data-allow-clear="true">
+                                            <option></option>
                                             @foreach($scopes as $scope)
                                             <option value="{{ $scope }}"
+                                                @if(is_array($dataType->scope))
```
MrCrayon: I think I would prefer just
```php
@if(in_array($scope, Arr::wrap($dataType->scope))) selected @endif
```

src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -996,4 +988,17 @@ protected function relationIsUsingAccessorAsLabel($details)
...
+        if (isset($dataType->scope) && !empty($dataType->scope)) {
+            if( ! is_array($dataType->scope))
+                $dataType->scope = [$dataType->scope];
+            foreach ($dataType->scope as $scope)
+            {
+                if (method_exists($model, 'scope'.ucfirst($scope)))
+                    $query->{$scope}();    
+            }
+        }
+    }
```
MrCrayon, line 1004: I find it could be more readable this way:

```php
protected function applyScopes($dataType, $model, $query)
{
    if (empty($dataType->scope)) {
        return;
    }

    foreach (Arr:wrap($dataType->scope) as $scope) {
        if (method_exists($model, 'scope'.ucfirst($scope))) {
            $query->{$scope}();
        }
    }
}
```

No need to use `isset` if you are using `empy` and less levels of nesting.

Cyrille37:
Hi @MrCrayon 
I've applied proposed changes.
Cheers.

Cyrille37:
Automatic tests failed with messages like:
```
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/25a956c84d5dd820d28caab9f86b8d183aeeff3d'. Error: The operation was canceled.
Warning: Back off 18.674 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/25a956c84d5dd820d28caab9f86b8d183aeeff3d'. Error: The operation was canceled.
Warning: Back off 21 seconds before retry.
```

PR #5399, opened 2021-07-08
Change method where to override Voyager models

### LINKED ISSUE #5389: Documentation "Overriding files" > "Voyager-Models" has lack or wrong information
## Version information

 - Laravel: v8.49.0
 - Voyager: v1.x-dev aaee6ab
 - PHP: 7.4.14
 - Database: mySQL 5.7.32

## Description
When i overide the MenuItem (but i guess the problem is the same with other files) i put in my `AppServiceProvier.php`

```
{
    public function register()
    {
        Voyager::useModel('MenuItem', \App\Models\MenuItem::class);
```

and in my `\App\Models\MenuItem.php`

```
    public static function boot()
    {
        parent::boot();

        static::retrieved(function ($model) {
            dd('retrieved');
        });
    }
```
sadly, the `static::retrieved` is never triggered.

## Steps To Reproduce
1. Then i go to the Admin panel > Tools > Menu Builder > admin .... builder

## Expected behavior
output the dump "retrieved"

## Additional context

but if i put `Voyager::useModel('MenuItem', \App\Models\MenuItem::class);` in the `boot()` method  instead of the `register()`, then the `static::retrieved` event is triggered...

So i suggest to modify the documentation to tell to put `Voyager::useModel` in the `boot` method instead.

the ultimate goal is to be able to override MenuItem and be sure that the original `$model->menu->removeMenuFromCache();` are still executed. Otherwise, we get only the menu from the cache and not from the databse.

i didn't wanted to do a PR because i have no idea if movig that `Voyager::useModel` from the register() to boot() would had a negative impact on the voyager package.

MrCrayon:
If `Voyager::useModel` is used in `register` method `boot` events do not work in the Model.

See related issue for more details.

Closes #5389

### DIFF
--- a/docs/customization/overriding-files.md
+++ b/docs/customization/overriding-files.md
@@ -90,7 +90,7 @@ use TCG\Voyager\Facades\Voyager;
 
 class AppServiceProvider extends ServiceProvider
 {
-    public function register()
+    public function boot()
     {
         Voyager::useModel('DataRow', \App\DataRow::class);
     }


fletch3555:
I'd kinda like to know why it doesn't work in register... seems like it should

MrCrayon:
For the record and future clarity the reason why it does not work in `register` is because in `useModel` we boot the Model to check if the instance is of the right type.
In Laravel models are booted only once per request and booting a model in `register` breaks events.

I did not go deep to understand why events breaks if the model is booted in `register` but this is reproducible with any model not related to Voyager.

Laravel documentation contains the warning:

> You should never attempt to register any event listeners, routes, or any other piece of functionality within the register method. Otherwise, you may accidentally use a service that is provided by a service provider which has not loaded yet.

So that's probably related.

fletch3555:
Ahhh okay, good call.  Thanks for the clarification

fletch3555 (approved):

emptynick:
Should we add a note that you can not do this in the register method?

fletch3555:
Nah, I think it's alright.  If people are doing that, then they're likely not reading the docs and therefore won't end up seeing the note anyway

Issue #5427, opened 2021-08-17
Database / BREAD : Deleting a Column let rubbish children in the database Translations table


Ronan-Lenor:
### Laravel version

8.53

### PHP version

7.4.14

### Voyager version

1.x-dev 1102296

### Database

MySQL 5.7.32

### Description

in the database, if i edit a table and delete a column that was translated. then its translation remain in the `translations` table with a dead key in the `foreign_key` column.

### Steps to reproduce

database > table : add a field
bread : update (will add the field in `data_rows` table)
Model Class : add the field name in the property $translatable = [... 'field'];
bread : (change translated field text) update (will add an entry in the `translations` table)
database > table : remove field
bread : update (will remove the field in `data_rows` table)
/!\ `translations` table keep the field entry

### Expected behavior

delete the dead entry

PR #5428, opened 2021-08-17
Create 2021_08_17_120402_alter_translations_table.php

Ronan-Lenor:
To resolve the problem mentionned here :
https://github.com/the-control-group/voyager/issues/5427


emptynick:
Not sure if I'm wrong, but the `translations` table references a lot of stuff, not only `data_rows`.
So when I have an item for any model with the same ID of a `data_row` and I delete that entry, a random `data_row` will be deleted.

Ronan-Lenor:
this is a one way cascading. `data_row` deletation trigger the deletation inside `translations`
but if a translation is deleted (by example through a menu_item deletation) then it won't cascade in `data_rows`
i did a test, everything look fine.

Edit: but it probably need further investigation.

Ronan-Lenor:
this part should be removed but i can't reedit the file
`DB::table('translations')->whereNotIn('foreign_key', DB::table('data_rows')->select('id'))->delete();`
this should be replace with something like this
```
        // to avoid the Illuminate\Database\QueryException :
        // Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails
        Schema::disableForeignKeyConstraints();

        Schema::table('translations', function (Blueprint $table)
        {
            $table->foreign('foreign_key')->references('id')->on('data_rows')->onDelete('cascade');
        });

        Schema::enableForeignKeyConstraints();
```

Ronan-Lenor:
i think i have found a better solution, some i will close this one

Ronan-Lenor:
see here now :
https://github.com/Ronan-Lenor/voyager/commit/e6d6c80cb9b660973aa1ac8f667de407d1127620#commitcomment-55720069

Issue #4980, opened 2020-06-15
Thumbnails are not deleted when deleting a media image


mko-tapptic:
## Version information
 - Laravel: v6.18.18
 - Voyager: v1.4.1
 - PHP: 7.3.11
 - Database: n/a

## Description
Deleting an image from media does not delete related thumbnails

## Steps To Reproduce
1. Define a bread media field with thumbnail options
2. Upload image
3. Delete image
4. Go to media, see that the original image is deleted, but thumbnail is not

## Expected behavior
Deleting image deletes the thumbs as well.

MrCrayon:
What's the type you are selecting for that field: image, multiple_images or media_picker ?

mko-tapptic:
media_picker, duh

alexsergivan:
I have the same issue, using media_picker

michalniks:
I have the same issue with media_picker
Laravel: v8.10.0
Voyager: v1.4.1
PHP: 7.3.1

AmrSubZero:
i also have the same issue with `Media Picker` with `Laravel 8.61.0` `Voyager 1.5` `PHP 7.4.13`

```
"thumbnails": [{
    "type": "fit",
    "name": "fit-500",
    "width": 500,
    "position": "center",
    "upsize": true
}]
```

egidiusmengelberg:
I'm having the same issue. I think I'll fix it for now by adding a delete all button. Not really an elegant solution but it's better than nothing.

rdupont55:
In addition, if you upload a file, that file has a thumbnail generated, and then you rename the root file, the thumbnail does not get renamed, yet remains in the directory with no way through the GUI to manage it.

I don't see why thumbnails don't show up in the GUI to be managed.  It seems reasonable that you may want to delete a specific thumbnail, rename the thumbnail, or move the thumbnail like you would with a base image.

PR #5444, opened 2021-09-17
Fix media picker delete thumbnails

howdu:
'thumbnails' was already in the request. Simple hooked this up with the file delete. 

Fixes issue:
https://github.com/the-control-group/voyager/issues/4980


emptynick:
While I think this is a good idea I think we should make this configurable somehow.
Only because one deleted the "parent" image doesn't mean you necessarily want to delete the thumbnails as well.

PR #5454, opened 2021-09-29
Custom field views for specific action

jimmysp:
Hi, first of all, this is my first PR, hope I got it right. 
I would like to improve the custom field views, having options for each separate action. I also made an edit in the documentation section, the custom views can be set like this:
{
    "view_browse": "my_browse_view",
    "view_read": "my_read_view",
    "view_add": "my_add_view",
    "view_edit": "my_edit_view",
    "view_order": "my_order_view"
}

### DIFF
--- a/docs/bread/formfields/introduction.md
+++ b/docs/bread/formfields/introduction.md
@@ -1,8 +1,8 @@
 # Formfields
 
-Formfields are the hearth of Voyagers BREAD-system.  
-Each formfield represents a field in your database-table and one input (or output) in BREAD.  
-To tweak your formfields you can insert JSON options which are described in the following pages.  
+Formfields are the hearth of Voyagers BREAD-system.
+Each formfield represents a field in your database-table and one input (or output) in BREAD.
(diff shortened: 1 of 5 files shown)


docs/bread/formfields/introduction.md
```diff
@@ -1,8 +1,8 @@
 # Formfields
 
-Formfields are the hearth of Voyagers BREAD-system.
```
emptynick: This tells the markdown parser to add a line-break here which we want.
Please revert those changes (same for the other lines below)

emptynick (dismissed):

jimmysp:
Ok, I restored the whitespaces with a new commit.

emptynick:
I think for consistency it would be great to pass `$view` as well.
This way you could also have the same view for browse and read for example, which are very similar from code perspective.

jimmysp:
Indeed, it could be useful to use the same view for similar actions... I pushed a new commit for this.

emptynick (approved):

PR #5463, opened 2021-10-13
Beautify pagination on server side rendering of bread browse

protectedfor:


emptynick:
Voyager does not use Bootstrap 4 and this view does not exist in Voyager or Laravel.

krio-rogue:
You can change default pagination template https://laravel.com/docs/8.x/pagination#using-bootstrap

PR #5465, opened 2021-10-14
Don't load all relationship results

emptynick:
In a belongs to many relationship all related items were loaded here:
https://github.com/the-control-group/voyager/blob/eb83ecb0f10192eed83dd8d2a3975dc3c8b32252/resources/views/formfields/relationship.blade.php#L174

Although we use ajax to load related items.
Another problem was that enabling tagging disabled ajax which does not make sense at all.

Possibilities tested:
- Adding
- Editing
- Editing with failing validation (selected relationships stay)
- Tagging
- No tagging (yes, this needed to be considered)


vitormicillo (approved):
Works perfect

vitormicillo (approved):

fletch3555 (approved):

FromSi:
belongsToMany in BREAD

Error when entering the edit page in BREAD

photo_2021-11-03 18 25 42

33Piter:
> belongsToMany in BREAD
> 
> Error when entering the edit page in BREAD
> 
> photo_2021-11-03 18 25 42

Same problem here.

emptynick:
That's not really a lot information to reproduce.
Can you guys figure out which query causes this?

Edit: Seems like this happens when you have an ID field in the pivot table - is there a reason to have this?

Edit 2: Fixed in https://github.com/the-control-group/voyager/commit/8c3c1dde896a047b1334d68f6d6a1d32fd07df1b

33Piter:
> That's not really a lot information to reproduce. Can you guys figure out which query causes this?
> 
> Edit: Seems like this happens when you have an ID field in the pivot table - is there a reason to have this?
> 
> Edit 2: Fixed in [8c3c1dd](https://github.com/the-control-group/voyager/commit/8c3c1dde896a047b1334d68f6d6a1d32fd07df1b)

Hello, @emptynick, thanks for the quick fix. 

Came here to give more info, but you already found it.

That's exactly the scenario that causes the problem: I'm working on a project that have a pivot table with id column.

PR #5488, opened 2021-11-08
Add Burmese Translations

ShaungBhone:
Add Burmese Translations in `seeders.php`, `settings.php`, and `theme.php`.


emptynick:
Simply looking at wikipedia it says burmese lang code is `my`.
Is there any difference between `mm` and `my`?

ShaungBhone:
Both are same. But I follow the wiki. We are using mm in local. We are Myanmar. Western call Burma. We are complex.

emptynick (approved):

PR #5498, opened 2021-12-02
Add "use_intervention" option in config, for an uploaded images in media manager.

karam-mustafa:
Voyager uses intervention to make the images for each upload request in the media manager, but there are no options to stop this process.
I added a new key to a media config that allows us to determine if we want to compress the images or not.

### DIFF
--- a/publishable/config/voyager.php
+++ b/publishable/config/voyager.php
@@ -239,6 +239,7 @@ return [
         'allow_delete'        => true,
         'allow_create_folder' => true,
         'allow_rename'        => true,
+        'use_intervention'    => true,
         /*'watermark'           => [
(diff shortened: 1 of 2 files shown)


Ahmed0saeed (commented):
Good option :)

publishable/config/voyager.php
```diff
@@ -239,6 +239,7 @@
         'allow_delete'        => true,
         'allow_create_folder' => true,
         'allow_rename'        => true,
+        'use_intervention'    => true,
```
emptynick, line 242: This also needs to be added to the `voyager_dummy.php` config file

emptynick (changes requested):

PR #5505, opened 2021-12-19
Fixes image URLs incorrectly encoded with backslashes

### LINKED ISSUE #5230: Voyager encodes / to %5C
## Version information

 - Laravel: v5.8.38
 - Voyager: v1.2
 - PHP: 7.3

## Description

I connected S3 DigitalOcean Spaces with my Laravel App using Voyager as an admin panel, the upload works fine although it encodes the forward slash into backslash as "%5C" and i get a 404 error, although when i change manually the link from "%5C" to "/" it worked, Voyager Media in admin panel displays the images just fine

## Steps To Reproduce
1. Connect S3 DO Spaces with Laravel 
2. Upload an image with Voyager
3. 404 Error because it converts forward slash to encoded backslash "%5C"

mikepsinn:
Fixes #5230

### DIFF
--- a/src/Voyager.php
+++ b/src/Voyager.php
@@ -257,7 +257,10 @@ class Voyager
     public function image($file, $default = '')
     {
         if (!empty($file)) {
-            return str_replace('\\', '/', Storage::disk(config('voyager.storage.disk'))->url($file));
+            $url = Storage::disk(config('voyager.storage.disk'))->url($file);
+            $url = str_replace('\\', '/', $url);
+            $url = str_replace('%5C', '/', $url);
+            return $url;
         }
 
         return $default;


src/Voyager.php
```diff
@@ -257,7 +257,10 @@ public function setting($key, $default = null)
     public function image($file, $default = '')
     {
         if (!empty($file)) {
-            return str_replace('\\', '/', Storage::disk(config('voyager.storage.disk'))->url($file));
+            $url = Storage::disk(config('voyager.storage.disk'))->url($file);
+            $url = str_replace('\\', '/', $url);
```
emptynick, line 261: This can be simplified by using an array as the needle-parameter like `$url = str_replace(['\\', '%5C'], '/', $url);`

emptynick (changes requested):

PR #5532, opened 2022-02-15
Laravel 9 support

### LINKED ISSUE #5523: Laravel 9 support
### Laravel version

9.0

### PHP version

8.1.0

### Voyager version

1.5

### Description of problem

It is not possible to install on Laravel 9.

image

### Proposed solution

Laravel 9 support

### Alternatives considered

_No response_

### Additional context

_No response_

emptynick:
This PR adds Laravel 9 support.
Because Flysystem changed their API a few changes had to be done related to the media manager.
Everything was tested with both Laravel 8 and 9.
Some quirks in the media manager might appear later unfortunately.

Also had to remove some tests from `SearchTest` which might be reintroduced later.

I also bumped some NPM dependencies to their latest version where possible.

Fixes #5523


composer.json
```diff
@@ -16,19 +16,19 @@
     ],
     "require": {
         "php": "^7.3|^7.4|^8.0",
-        "illuminate/support": "~6.0|~7.0|~8.0",
```
fletch3555, line 19: Do we want to maintain support for laravel 6/7?  Does this require a semver major bump, or only minor? (On my phone and it won't show me changes in a few of the files)

emptynick: This does not require a major or minor version bump. All versions above should work.
We could deprecate 6 and 7 and bump one minor version if we fancy.

fletch3555: Given the scope of the changes, I'd say a minor bump is probably prudent either way, so deprecating seems reasonable.  They're no longer supported by paragel anyway

ibnuhalimm:
Please release this merge, we love voyager and want to upgrade to laravel 9

PR #5541, opened 2022-02-21
Get model relationships after and with translation

almooradi-dev:
We had a problem, that we weren't getting the model relations after translating the main model.
For example:
`$posts = Post::with(['author', 'translations'])->get()->translate('locale', 'fallbackLocale');`
The problem is that the `author` parameter (relation) is not appearing in the final translated object, just the main attributes.

And we have fixed this problem.


YoussefMaged:
Please check this also #5554

src/Translator.php
```diff
@@ -43,6 +52,24 @@ public function translate($locale = null, $fallback = true)
             $this->translateAttribute($attribute, $locale, $fallback);
         }
 
+        // Translate Relations
+        foreach ($this->model->getRelations() as $key => $relation) {
+            if ((new Voyager)->translatable($relation)) {
```
AlaaL, line 57: First of all you are a life save I came here looking for this as I was trying to do the same thing, the only problem for me with your solution that i didn't want the translations to be loaded separately after the translate did the work, so I added this peace of code to remove the translations relations 

if ($key == "translations") {
                unset($this[$key]);
                continue;
  }

a-ahmed116 (approved):

a-ahmed116:
Why are these changes not merged

amrahmed000 (approved):

AlaaL (approved):

AlaaL (approved):

AlaaL (approved):

AlaaL:
I approved the review why it not merged ?

src/Translator.php
YoussefMaged:
```suggestion
            if ((new Voyager)->translatable($relation)) {
                        if ($key == "translations") {
                                    unset($this[$key]);
                                    continue;
                        }
```

YoussefMaged: Thank you so much for your modification, that's really good too, agreed!

YoussefMaged (approved):

almooradi-dev:
First, I'm glad that this was a part to solve your problem.

Second, does the merging problem have something to do with me (as I have the first commit) or is it something related to the "Voyager" authors?

AlaaL:
> First, I'm glad that this was a part to solve your problem.
> 
> Second, does the merging problem have something to do with me (as I have the first commit) or is it something related to the "Voyager" authors?

I think the Voyager authors not doing any merges, there are a lot of pull-requests not merged, I fell like the project is dead

PR #5560, opened 2022-03-25
fixes: fixed json_decode(): Passing null to parameter of type string …

ahmadsyamim:
…is deprecated


emptynick:
Superseeded by #5575

PR #5562, opened 2022-03-28
fixes: prevent exception error on missing table

ahmadsyamim:


emptynick:
You _could_ rename the settings table and tell the model about it.
Then this code doesn't work anymore.

PR #5565, opened 2022-03-30
Fix #5564 "The HTTP status code \"xxx\" is not valid." error when trying to delete an image field image in Model edit page, and toastr error message not showing

### LINKED ISSUE #5564: "The HTTP status code \"xxx\" is not valid." error when trying to delete an image field image in Model edit page, and toastr error message not showing
### Laravel version

8.83.5

### PHP version

7.4.27

### Voyager version

1.5.2

### Database

10.6.5

### Description

When trying to delete an image field image from the edit page, and the ajax returns an error for some reason a wrong HTTP response code is used which causes the following error (as an example) to be thrown "The HTTP status code \"23000\" is not valid.". 

### Steps to reproduce

1. Create a table with a field to store an image URL that is not nullable (so an error is thrown when trying to delete the image).
2. Create the needed BREAD and use "Image" field type.
3. Create a record with an image.
4. Go to the edit page.
5. Try to delete the image.
6. You will get "The HTTP status code \"xxx\" is not valid." message in the console and ajax call will fail.

### Expected behavior

A 500 HTTP response is to be returned from the Ajax call after an error.

### Screenshots

image

### Additional context

_No response_

mbasfour:
Fixes #5564

### DIFF
--- a/resources/views/bread/edit-add.blade.php
+++ b/resources/views/bread/edit-add.blade.php
@@ -208,6 +208,9 @@
                     } else {
                         toastr.error("Error removing file.");
                     }
+                })
+                .fail(function() {
+                    toastr.error("Error removing file.");
                 });
 
                 $('#confirm_delete_modal').modal('hide');
--- a/src/Http/Controllers/VoyagerBaseController.php
+++ b/src/Http/Controllers/VoyagerBaseController.php
@@ -63,9 +63,9 @@ class VoyagerBaseController extends Controller
         if (strlen($dataType->model_name) != 0) {
             $model = app($dataType->model_name);
 
-            $query = $model::select($dataType->name.'.*');
+            $query = $model::select($dataType->name . '.*');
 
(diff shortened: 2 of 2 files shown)


resources/views/bread/edit-add.blade.php
```diff
@@ -208,6 +208,9 @@ function deleteHandler(tag, isMulti) {
                     } else {
                         toastr.error("Error removing file.");
                     }
+                })
+                .fail(function() {
+                    toastr.error("Error removing file.");
```
mbasfour, line 213: Handle jQuery request failure and show error message

src/Http/Controllers/VoyagerBaseController.php
```diff
@@ -686,7 +686,7 @@ public function remove_media(Request $request)
                     'status'  => $code,
                     'message' => $message,
                 ],
-            ], $code);
+            ], 500);
```
mbasfour, line 689: Set HTTP response code to 500 when there is an error instead of the exception code

PR #5569, opened 2022-04-09
fix bug

taochangle:
fix:You cannot modify your own role filed in the user edit page , but you can modify others'


emptynick:
That's exactly how we want it to be.
Being able to change your own roles would allow you to lock yourself out of Voyager or, even worse, grant yourself higher permissions

emptynick:
That's exactly how we want it to be.
Being able to change your own roles would allow you to lock yourself out of Voyager or, even worse, grant yourself higher permissions

emptynick:
Closing as this is wanted behaviour

PR #5573, opened 2022-04-12
Throw exception if undefined form field type

howdu:
Happens if you try to use `Voyager::formField` on a relationship field type.


emptynick (approved):

novruzrhmv:
It should be `throw new \Exception` otherwise `Class TCG\Voyager\Exception` error will be thrown.

Issue #5578, opened 2022-04-14
Can't delete an Image of Multiple Images field


ibnuhalimm:
### Laravel version

6.20.44

### PHP version

7.3.27

### Voyager version

1.5.2

### Database

MySQL 5.7.37

### Description

I'm getting 500 error message "Invalid argument supplied for foreach()" while trying to delete an image of "multiple image" field.

```
[2022-04-14 06:19:51] production.ERROR: Invalid argument supplied for foreach() {"userId":9,"exception":"[object] (ErrorException(code: 0): Invalid argument supplied for foreach() at /var/www/html/application/vendor/tcg/voyager/src/Http/Controllers/VoyagerBaseController.php:757)
[stacktrace]
#0 /var/www/html/application/vendor/tcg/voyager/src/Http/Controllers/VoyagerBaseController.php(757): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Invalid argumen...', '/var/www/html/a...', 757, Array)
#1 /var/www/html/application/vendor/tcg/voyager/src/Http/Controllers/VoyagerBaseController.php(746): TCG\\Voyager\\Http\\Controllers\\VoyagerBaseController->deleteBreadImages(Object(App\\Trip), Array, 'trips/April2022...')
...
```

### Steps to reproduce

1. Make a field as "multiple_images" at BREAD system
2. Create new data
3. Upload some images to these "multiple_images" field
4. Edit the data
5. Try to delete an image

### Expected behavior

Response success, and image file should be deleted

### Screenshots

![image](https://user-images.githubusercontent.com/14322078/163326937-ebb65bec-65cd-40a1-a619-5f622de5961d.png)

ibnuhalimm:
It says the error on `Controllers/VoyagerBaseController.php` on line 757.
```php
...
       foreach ($rows as $row) {
            if ($row->type == 'multiple_images') {
                $images_to_remove = json_decode($data->getOriginal($row->field), true) ?? [];
            } else {
                $images_to_remove = [$data->getOriginal($row->field)];
            }

            foreach ($images_to_remove as $image) {
                // Remove only $single_image if we are removing from bread edit
...
```

I've checked that the `$images_to_remove` variable contains array data, but the error shows like the `$images_to_remove` is not an array.

ibnuhalimm:
Maybe it's bug of php7.3

PR #5579, opened 2022-04-14
fix delete an image of multiple images field

ibnuhalimm:
This PR to fix issue #5578


emptynick:
This won't help you as you are using Voyager 1.4 but are targeting 1.5 here.

ibnuhalimm:
Thanks for the correction. I've tried to upgrade my voyager to v1.5, but the problem still out there.
My issue was updated.

PR #5628, opened 2022-07-06
Prefix body class

howdu:
Fixes display issue with table/bread called 'widgets'


emptynick:
I am not sure why this is there at all.
Its either used somewhere (in javascript for example) or it is completely useless anyway.
I think we could simply remove it entirely.

Issue #5656, opened 2022-08-12
Menu Icon Issue & Alignment


AliRaza-commits:
### Laravel version

8.75

### PHP version

7.4.8

### Voyager version

1.5

### Database

MySql 8.0

### Description

Hi,
I tried to use menu after adding it to **Menu Builder**.  The code i used in my blade was:
~~~
{!! menu('customer') !!}
~~~

However, my menu was without icon as well as improper positioning.

### Steps to reproduce

To fix this issue, I've updated file in resources Menu/default. 

### Expected behavior

Menu should've menu item icons with proper alignment.

### Screenshots

![BugScreenShot](https://user-images.githubusercontent.com/80054032/184417217-b3258497-b966-4085-af5e-bc909d9f7e56.jpg)

emptynick:
The `menu` helper is not made to be used inside of Voyager but inside your frontend.
Showing and hiding items in Voyager should be done based on permissions.

PR #5657, opened 2022-08-12
Fixed menu display so that it display icon when using new menu from m…

AliRaza-commits:
Hello @emptynick  I've fixed the Bug for the given issue highlighted by me: #5656


emptynick:
See my comment in https://github.com/the-control-group/voyager/issues/5656

PR #5662, opened 2022-08-19
Bread - Browse and Read for MediaPicker

toto975:
Depends on the mimetype of the media/file, displays the image or the Download link for media picker field.

In the browse and read view, "Couverture" is a Media picker with an image, "Pdf complet" is a Media picker with a pdf file, with the download link.

image

image


toto975:
Update 12/27/2022
If for some reason, the file has disappeared, we need to add some test to check if the file exists.
If the file doesn't exists anymore, display blank text / nothing

PR #5672, opened 2022-09-16
php_8.1_update_1

jimmythewebexpert:
new php version 8.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,37 +1,32 @@
 {
-    "name": "tcg/voyager",
+    "name": "jimmythewebexpert/voyager",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -15,23 +15,23 @@
         }
     ],
     "require": {
-        "php": "^7.3|^7.4|^8.0",
-        "illuminate/support": "~6.0|~7.0|~8.0|~9.0",
+        "php": "^7.3|^7.4|^8.0|^8.1|^9.0",
```
fletch3555: ^8.0 will already include 8.1, and 9.0 isn't out yet so we cannot accept this

emptynick:
PHP 8.1 is now supported in version 1.6.0

PR #5679, opened 2022-10-16
Add changed relationship attributes info

haydar:
I faced this problem in a project. This, I believe, should be in docs.


emptynick:
1.2 is more than 3 years old and we don't support it anymore.

haydar:
@emptynick Many projects continue to use legacy versions. My objective is to motivate developers to upgrade to the most recent version. Not documented this breaking change does not profit anyone, some developers could save hours when attempting to upgrade howewer you merge this pr. Please reconsider your desicion.

PR #5705, opened 2022-12-08
Softdelete toggle button align

toto975:
Before changing css
image
image

After changing css (padding left and right, margin-bottom)

image
image


dimer47:
Hi @toto975,

As a result, we no longer see the switch button that is on the left in the initial screenshot. 
Shouldn't we truncate the text with text-overflow : ellipsis ?

🇫🇷  French Voyager too

PR #5732, opened 2023-02-03
Laravel 10.x Compatibility

laravel-shift:
This is an automated pull request from [Shift](https://laravelshift.com) to update your package code and dependencies to be compatible with Laravel 10.x.

**Before merging**, you need to:

- Checkout the `l10-compatibility` branch
- Review **all** comments for additional changes
- Thoroughly test your package

If you do find an issue, please report it by commenting on this PR to help improve future automation.

PR #5750, opened 2023-03-29
Update composer.json

elmehdi66:
laravel 10


jf-m:
Updating illuminate/support is not enough unfortunately. 
See #5732

PR #5857, opened 2024-05-06
Fix "Unable to init from given binary data." when upload media

nuocgansoi:
Storage::get($path) returns file contents as a string which may not be cast to valid binary data for Image::make() to be able to read.


nuocgansoi:
This is only necessary when running on Wsl 1.

PR #5867, opened 2024-06-20
Update fonts.blade.php

canerksk:
Added bootstrap icons


resources/views/compass/includes/fonts.blade.php
```diff
@@ -1654,6 +1654,28 @@
     </li>
 </ul>
 
+@php
+    $response = \Http::get('https://raw.githubusercontent.com/twbs/icons/main/font/bootstrap-icons.json');
```
Emerica, line 1658: This will break and or could be abused.

PR #5868, opened 2024-06-20
Update master.blade.php

canerksk:
Added bootstrap-icons cdn and font-awesome icons cdn links


Emerica:
You can do this already via Voyagers config file, there's a section called additional_js
image

PR #5875, opened 2024-07-04
Adds a Three Columns List Formfield

JdSeus:
Adds a formfield that serves as a user-friendly list editor with three columns and saves the data as JSON.

ThreeColumnsList
Bread
Database


soinalastudio:
Why limit to 3 columns? It's nice but I think it's better to allow dynamical column counts. Have you seen https://github.com/MonstreX/voyager-extension?tab=readme-ov-file#field-ve-json-fields ?

PR #5893, opened 2024-10-08
Enable sort for relationship on full datas

toto975:
Enable sort for relationship on full data, not only page per page dataset.

For example, datas are 

eat
drink
apple
orange
diner
pen

Default Voyager sorting and displaying are, with 3 datas per page :

**first page**
apple
drink
eat

**second page**
diner
orange
pen

With my PR, sorting and displaying are

**first page**
apple
diner
drink

**second page**
eat
orange
pen


bobbyiliev (approved):
Thank you!

